PR #128, opened 2026-05-01
fix: scope Custom Claude Installation override per workspace

omartelo:
## Summary

The "Custom Claude Installation" field in **Settings → Project → Claude Agent** was persisted as a single global key (top-level `customClaudeCodePath` in `ai-settings.json`), so any value set in the Project tab leaked into the User scope and into every other project. It was the only AI setting in the panel that bypassed the per-workspace override mechanism already used by `defaultProvider`, `apiKey`, `models`, `defaultModel`, and `enabled`.

This PR wires `customClaudeCodePath` into the existing override infrastructure, mirroring the `defaultProvider` pattern.

Fixes #125.

## Approach

- **Schema** (`packages/electron/src/main/utils/aiSettingsMerge.ts`, `packages/electron/src/main/utils/store.ts`, `packages/electron/src/renderer/store/atoms/appSettings.ts`): add `customClaudeCodePath` to `GlobalAISettings`, `EffectiveAISettings`, `EffectiveAISettings.overrides`, and `AIProviderOverrides`. Extend `mergeAISettings` with the same conditional override block already used for `defaultProvider`. Update `normalizeAIProviderOverrides` (in `store.ts` and the parallel implementation in `AIService.ts`) so the new field is considered when collapsing an empty override object.
- **Loader** (`packages/runtime/src/ai/server/providers/claudeCode/dependencyInjection.ts`, `packages/runtime/src/ai/server/providers/claudeCode/sdkOptionsBuilder.ts`): change the loader signature from `() => string` to `(workspacePath?: string) => string`. `sdkOptionsBuilder` now forwards the active `workspacePath` (already in scope at the `cwd` line). The loader registered in `packages/electron/src/main/index.ts` resolves the project override first and falls back to the global value.
- **UI** (`packages/electron/src/renderer/components/GlobalSettings/panels/ClaudeCodePanel.tsx`, `packages/electron/src/renderer/components/Settings/SettingsView.tsx`): add `scope` and `workspacePath` props to `ClaudeCodePanel` (matching the convention used by `MCPServersPanel`). In project scope the panel reads/writes the workspace override through the existing `ai:getProjectSettings` / `ai:saveProjectSettings` IPC handlers; clearing the field clears the override so the project re-inherits the global value. The placeholder and helper copy surface the inherited global path so the inheritance state is obvious. `SettingsView` passes the matching scope and workspace path when rendering the panel.

## Backward compatibility

[...]

Issue #125, opened 2026-05-01
[Bug]: Custom Claude Installation: project-scoped override is persisted globally and leaks across scopes


omartelo:
## Summary

The "Custom Claude Installation" setting under **Settings → Project → Claude Agent** does not actually persist as a project-scoped override. Any value set there is written to the global `customClaudeCodePath` key in `ai-settings.json` and bleeds into the User scope and into every other project. The "OVERRIDE" toggle and the User/Project tabs suggest scoped storage, but only a single global key exists.

This is the only AI-setting field exposed in the Project panel that lacks the per-workspace override mechanism — `defaultProvider`, `apiKey`, `models`, `defaultModel`, and `enabled` all merge global + project correctly via the existing infrastructure.

## Steps to Reproduce

1. Open a project (Project A).
2. Go to **Settings → Project → Claude Agent**.
3. Toggle **OVERRIDE** on for "Custom Claude Installation" and set a path (e.g. `/home/user/.local/bin/claude-corp.sh`).
4. Switch to **Settings → User → Claude Agent**, or open Project B's settings.

**Expected:** the User scope (and Project B) show their own values, independent of Project A's override.

**Actual:** all scopes show the same path that was set in Project A. The value is global.

## Evidence

`~/.config/@nimbalyst/electron/ai-settings.json` after setting via Project scope:

```json
{
  "defaultProvider": "claude-code",
  "providerSettings": { "...": "..." },
  "customClaudeCodePath": "/home/user/.local/bin/claude-corp.sh",
  "trackerAutomation": { "...": "..." }
}
```

`customClaudeCodePath` lives at the top level — there is no `aiProviderOverrides.customClaudeCodePath` in any workspace state.

## Root Cause

The infrastructure for project-scoped overrides already exists in `packages/electron/src/main/utils/aiSettingsMerge.ts` and `WorkspaceState.aiProviderOverrides`. `customClaudeCodePath` was simply never wired into it:

- Not declared in `GlobalAISettings`, `EffectiveAISettings`, `EffectiveAISettings.overrides`, or `AIProviderOverrides`.
- `AIService.ts` (lines ~2499, 2521, 2632–2633) reads/writes the top-level store key with no scope parameter.
- `index.ts:1678` registers a loader that reads the global key directly: `setCustomClaudeCodePathLoader(() => store.get('customClaudeCodePath', ''))`.
[...]

PR #131, opened 2026-05-01
fix: inherit project customClaudeCodePath override inside worktrees

### LINKED ISSUE #130: [Bug]: Custom Claude Installation override is ignored inside git worktrees
## Summary

Follow-up to #125 (fixed by #128).

The per-workspace `customClaudeCodePath` override now works for sessions opened in the project root, but is **silently ignored** for sessions opened inside a git worktree of that project. Worktree sessions fall back to the bundled SDK binary, bypassing the wrapper the user configured at the project scope.

## Root cause (from runtime logs)

The override is keyed by the normalized workspace path. A worktree has its own path (`<projectRoot>_worktrees/<name>`), which does not match the path under which the override is stored, so `getAIProviderOverrides(worktreePath)` returns `null`.

The loader registered in `packages/electron/src/main/index.ts` does not perform the worktree -> parent-project fallback that other code paths in `AIService` already do (e.g. `permissionsPath` via `inferWorktreePathFromFilePath`).

## Steps to reproduce

1. Open any project in Nimbalyst.
2. Settings -> AI -> Claude Code -> scope **Project** -> set **Custom Claude Installation** to a wrapper script (any executable that prints/uses different credentials, e.g. `/home/<user>/.local/bin/<wrapper>.sh`).
3. Confirm the UI shows ``Project-specific path active``.
4. Start a new AI session in the **project root** -> expected: wrapper is used.
5. Click **New Worktree** to create a worktree from the project.
6. Start a new AI session **inside the worktree** -> actual: bundled SDK binary is used; the override is ignored.

## Evidence

Paths anonymized below (`<user>`, `<project>`, `<worktree>`, `<wrapper>`, `<nimbalyst-checkout>`).

### Session in project root (works)

```
[HH:MM:SS]  (MAIN) : [ClaudeCodeProvider] customClaudeCodePath loader: workspace=/home/<user>/projects/<project> globalPath=(empty) projectPath=/home/<user>/.local/bin/<wrapper>.sh overridesObject={"providers":{"claude-code":{"enabled":true}},"customClaudeCodePath":"/home/<user>/.local/bin/<wrapper>.sh"} effective=/home/<user>/.local/bin/<wrapper>.sh
[HH:MM:SS]  (MAIN) : [CLAUDE-CODE] Binary path: custom=/home/<user>/.local/bin/<wrapper>.sh resolved=/home/<user>/<nimbalyst-checkout>/packages/electron/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64/claude effective=/home/<user>/.local/bin/<wrapper>.sh
```

### Session in a worktree of the same project (broken)

```
[HH:MM:SS]  (MAIN) : [ClaudeCodeProvider] customClaudeCodePath loader: workspace=/home/<user>/projects/<project>_worktrees/<worktree> globalPath=(empty) projectPath=(unset) overridesObject=null effective=(empty)
[HH:MM:SS]  (MAIN) : [CLAUDE-CODE] Binary path: custom=(none) resolved=/home/<user>/<nimbalyst-checkout>/packages/electron/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64/claude effective=/home/<user>/<nimbalyst-checkout>/packages/electron/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64/claude
```

Same project, same override, same Nimbalyst process - only the workspace path passed to the loader differs. The worktree path (`...<project>_worktrees/<worktree>`) does not match the stored key (`...<project>`), so `overridesObject=null`.

(The diagnostic log line above is local instrumentation added to the loader to confirm the lookup; the underlying behavior is reproducible without it.)

## Expected behavior

Sessions running inside a worktree should inherit the parent project's `customClaudeCodePath` override unless the worktree itself defines a more specific override.

## Related

- Issue #125 (closed)
- PR #128 (merged) - fixed the global vs project leak but did not cover worktrees.

omartelo:
## Summary

The per-workspace `customClaudeCodePath` override is keyed by the normalized workspace path. Sessions running in a git worktree pass the worktree path (`<root>_worktrees/<name>`) into the loader, which never matches the entry stored under the parent project path. The loader silently fell back to the (empty) global value and the bundled SDK binary, bypassing the wrapper the user configured at the project scope.

This change makes the loader resolve the worktree to its parent project on a miss, reusing the same DB-aware resolution that `PermissionService` already uses for trust. Worktree-specific overrides still win because the direct lookup runs first.

## Lookup order

1. `getAIProviderOverrides(workspacePath)` — direct hit; a worktree-keyed override wins if one is ever written.
2. If `isWorktreePath(workspacePath)`, resolve the parent via `resolveWorkspacePathForPermissions(workspacePath)` (DB lookup against `worktrees.path`), then `getAIProviderOverrides(parentPath)`.
3. Fall back to the global value.

Resilience: the parent resolution call is wrapped in `try/catch`; on any failure (DB not initialized, query error, race during shutdown) the loader falls back to the synchronous `resolveProjectPath` regex so a session is never blocked from starting.

## Other small cleanups

- Replaced the `await import('./utils/store')` inside `app.whenReady` with a top-level static import, per the rule in `CLAUDE.md` against dynamic imports in the main process. The dynamic form also risked picking up a separate module instance with a stale `electron-store` cache.
- Updated `customClaudeCodePathLoader` and the static forwarder on `ClaudeCodeProvider` to accept `string | Promise<string>`, since the loader now needs an async path. The single call site in `sdkOptionsBuilder.ts` is already inside an async function and has been updated to `await` the call.

## Files changed

- `packages/electron/src/main/index.ts` — loader registration; static imports; two-step lookup with try/catch fallback.
- `packages/runtime/src/ai/server/providers/ClaudeCodeProvider.ts` — static forwarder signature.
- `packages/runtime/src/ai/server/providers/claudeCode/dependencyInjection.ts` — loader type.
- `packages/runtime/src/ai/server/providers/claudeCode/sdkOptionsBuilder.ts` — `await` the loader call.

No new utilities introduced; reuses `resolveWorkspacePathForPermissions`, `isWorktreePath`, `resolveProjectPath`, and `getAIProviderOverrides`.

## Test plan

- [x] TypeScript build of `packages/runtime` (`npm run build`).
- [x] TypeScript typecheck of `packages/electron` (`npx tsc --noEmit -p .`) — pre-existing unrelated `ThemesPanel.tsx` errors only; no new errors.

### Verification logs (anonymized)

Project root:

```
[HH:MM:SS] (MAIN) : [CLAUDE-CODE] Binary path: custom=/home/<user>/.local/bin/<wrapper>.sh resolved=/home/<user>/<nimbalyst-checkout>/packages/electron/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64/claude effective=/home/<user>/.local/bin/<wrapper>.sh
```

Worktree:

```
[HH:MM:SS] (MAIN) : [ClaudeCodeProvider] customClaudeCodePath: worktree '/home/<user>/projects/<project>_worktrees/<worktree>' inherited from parent '/home/<user>/projects/<project>'
[HH:MM:SS] (MAIN) : [CLAUDE-CODE] Binary path: custom=/home/<user>/.local/bin/<wrapper>.sh resolved=/home/<user>/<nimbalyst-checkout>/packages/electron/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64/claude effective=/home/<user>/.local/bin/<wrapper>.sh
```

## Related

Closes #130. Follow-up to #128 / #125.

### DIFF
--- a/packages/electron/src/main/index.ts
+++ b/packages/electron/src/main/index.ts
@@ -1702,7 +1702,7 @@ app.whenReady().then(async () => {
           parentPath = await resolveWorkspacePathForPermissions(workspacePath);
         } catch (error) {
           logger.main.warn(
-            `[ClaudeCodeProvider] Failed to resolve worktree parent via DB, falling back to regex: ${(error as Error).message}`
+            `[ClaudeCodeProvider] Failed to resolve worktree parent via DB, falling back to regex: ${error instanceof Error ? error.message : String(error)}`
           );
           parentPath = resolveProjectPath(workspacePath);
         }
--- a/packages/electron/src/main/index.ts
+++ b/packages/electron/src/main/index.ts
@@ -63,7 +63,9 @@ import {
     clearPendingThemeFallback,
     updateWorkspaceState,
     runMigrations,
-    getAppSetting
+    getAppSetting,
+    store,
+    getAIProviderOverrides
 } from './utils/store';
 import { registerMCPConfigHandlers } from './ipc/MCPConfigHandlers';
 import { getOpenCodeConfigService, registerOpenCodeConfigHandlers } from './ipc/OpenCodeConfigHandlers';
@@ -74,7 +76,7 @@ import { MCPConfigService } from './services/MCPConfigService';
 import { registerDatabaseBrowserHandlers } from './ipc/DatabaseBrowserHandlers';
 import { registerTerminalHandlers, shutdownTerminalHandlers } from './ipc/TerminalHandlers';
 import { AIService } from './services/ai/AIService';
-import { detectFileWorkspace, suggestWorkspaceForFile, getAdditionalDirectoriesForWorkspace } from './utils/workspaceDetection';
+import { detectFileWorkspace, suggestWorkspaceForFile, getAdditionalDirectoriesForWorkspace, isWorktreePath, resolveProjectPath } from './utils/workspaceDetection';
 import { cliManager, initEnhancedPath, getEnhancedPath, getShellEnvironment } from './services/CLIManager';
 import { registerWorkspaceWindow, registerExtensionTools, shutdownHttpServer, startMcpHttpServer, updateDocumentState } from './mcp/httpServer';
 import { startSessionContextServer, cleanupSessionContextServer, shutdownSessionContextServer } from './mcp/sessionContextServer';
@@ -124,7 +126,7 @@ import { initTrackerSchemaService, updateTrackerSchemaWorkspace } from './servic
 import { registerTeamHandlers, autoMatchTeamForWorkspace } from './services/TeamService';
 import { registerOrgKeyHandlers } from './services/OrgKeyService';
 import { registerDocumentSyncHandlers } from './ipc/DocumentSyncHandlers';
-import { getPermissionService } from './services/PermissionService';
+import { getPermissionService, resolveWorkspacePathForPermissions } from './services/PermissionService';
 import { ClaudeSettingsManager } from './services/ClaudeSettingsManager';
 import { TrayManager } from './tray/TrayManager';
 import { pathToFileURL } from 'url';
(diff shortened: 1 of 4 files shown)


omartelo:
Not taking the suggestions on `sdkOptionsBuilder.ts:116`. The loader handles its only throw-capable path internally and falls back to the regex resolver; an extra guard at the caller would mask real bugs rather than catch realistic failures.

omartelo:
Superseded by e44fa2de on `main`, which solves #130 with a simpler sync, regex-based resolution and adds a unit test. Closing.

PR #136, opened 2026-05-02
Add LaTeX math rendering to agent transcript markdown

### LINKED ISSUE #129: [Feature]: math markdown in agent response - update to MarkdownRenderer.tsx

I'm asking math-related questions agents and getting back ```$...$ (inline) and $$...$$ (block)``` LaTeX that is not rendered as math. Request you add:
``` 
import remarkGfm from 'remark-gfm';
import remarkMath from 'remark-math';
import rehypeKatex from 'rehype-katex';
import 'katex/dist/katex.min.css';
```
to MarkdownRenderer.tsx and then update the <ReactMarkdown> call:

```
<ReactMarkdown
  remarkPlugins={[remarkGfm, remarkMath]}
  rehypePlugins={[rehypeKatex]}
  components={{
    ...
  }}
>
```
This will require installing dependencies for remark-math rehype-katex and katex. It would be great to also update the Lexical editor, as well - but that looks like more work.

lisah2u:
Add remark-math, rehype-katex, and katex to render inline ($...$) and block ($$...$$) math notation in agent responses via MarkdownRenderer.

Summary

Agent responses containing LaTeX math notation were displayed as raw text because MarkdownRenderer only used remark-gfm. This adds three packages to the react-markdown pipeline in MarkdownRenderer.tsx:

•  remark-math — parses $ and $$ delimiters into math AST nodes
•  rehype-katex — renders those nodes into formatted math
•  katex — provides the CSS for styled output (bundled into runtime.css)

No changes to the Lexical document editor or other rendering paths.

Related issue

Closes #129

Testing

•  Verified runtime package builds successfully with new dependencies
•  Confirmed KaTeX CSS is bundled into dist/runtime.css
•  TypeScript typecheck passes
•  All existing unit tests pass
•  Manually verified in dev mode that inline ($x^2$) and block ($$\sum_{i=0}^{n} x_i$$) math renders correctly in agent transcript


ghinkle:
Synced with `main` and resolved the package-lock / runtime/package.json / THIRD_PARTY_LICENSES conflicts.

Two intentional changes on top of the original branch:

1. **Dropped the second commit** (`feat: add KaTeX math rendering to the Lexical document editor`). Looked like an accidental include — that change is the same content as #138, and keeping it here would have made #136 broader than the original "narrow transcript-only" scope.
2. **Added defense-in-depth options to `rehype-katex`** in `MarkdownRenderer.tsx` per the original KaTeX issue review: `trust: false`, `strict: 'ignore'`, `output: 'html'` (drops MathML, which has had its own XSS history), `maxSize: 25`, `maxExpand: 100`, empty `macros: {}`, `throwOnError: false`. Agent transcript markdown is untrusted input, and these match what we said we'd lock down before re-enabling KaTeX.

Typecheck is clean and `MarkdownRenderer` / `RichTranscriptView` / `TranscriptWidgets` tests pass (77/77).

lisah2u:
Thanks! Much appreciated. Thought I'd addressed the second commit for the additional changes... but apparently not quite. You guys move fast!

ademczuk:
Hi @lisah2u, clean diff and the security-defaults choice (`trust: false, strict: 'ignore'`) is the right one for AI-rendered content. Three things worth checking before merge:

**1. Currency-vs-math collision is the biggest user-facing risk.**

AI responses commonly contain monetary values like "this costs $5 to process $10000 worth of data". `remark-math` v6's default behaviour for single-dollar text math is the failure mode here: with `singleDollarTextMath: true` (default in some configs), `$5 to process $10000` could match as inline math because both dollar signs see non-whitespace adjacent. Result: the text between renders garbled or stripped, depending on how rehype-katex parses it.

Two ways to guard:

```ts
remarkPlugins={[remarkGfm, [remarkMath, { singleDollarTextMath: false }]]}
```

This requires `$$...$$` for block math but disables single-`$` inline. Inline math becomes `\(...\)` syntax (LaTeX standard) instead. Slightly less convenient for users who write `$x^2$` inline, but eliminates the currency collision entirely.

Or keep the default but add a regression test with sample text: `"It costs $5 to process $1000 of data and $50 in fees"`. If the rendered output preserves all four dollar amounts as text, you're safe; if any of them get math-treated, the option needs flipping.

I'd lean toward `singleDollarTextMath: false` as the safer default for AI-generated content. Math notation is rare in chat, currency is common. Worth a maintainer call.

**2. Side-effect CSS import is global.**

```ts
import 'katex/dist/katex.min.css';
```

This loads `katex.min.css` into the global renderer style sheet. KaTeX's classnames (`.katex`, `.katex-display`, `.katex-html`, etc.) are scoped to KaTeX-rendered output, so genuine conflicts with existing Nimbalyst CSS are unlikely. But: KaTeX also brings its own `@font-face` declarations for KaTeX-Main / KaTeX-Math etc., adding ~10 web fonts to the renderer. Worth confirming the bundler is tree-shaking unused font weights, otherwise this adds ~300KB of font assets to every renderer load even for users who never see math content.

Minor, but if the runtime bundle grows noticeably after this PR, the unconditional CSS import is the likely cause.

**3. Sister PR #138 already covers the editor side.**

Worth a single note in the PR body that `MarkdownRenderer.tsx` is chat-transcript-only. Math in `.md` document files opened in the Lexical editor still renders raw until #138 lands. Reduces confusion if a user files "math doesn't render in my docs" after this merges.

(1) is the one I'd want answered before merge. (2) and (3) are follow-up and nice-to-have.

Good direction otherwise. Closes a real gap for AI math output.

lisah2u:
Agree on ` singleDollarTextMath: false`. More robust and actually more readable. I've also observed that other platforms don't allow for inline latex - so this is a common solution. On (2), I'll check into best practice for KaTeX fonts. Good catch!

ademczuk:
Thanks lisah2u, that's a good data point on other platforms. Happy to retest the regex against the currency cases once the fix lands.

PR #138, opened 2026-05-02
feat: add KaTeX math rendering to the Lexical document editor resolves #137

### LINKED ISSUE #137: [Feature]: Math rendering (KaTeX) in the document editor

Problem: Markdown documents containing latex-like formulas both inline `$...$` and in block math `$$...$$` are not rendered as math. The Lexical editor treats $ delimiters as plain text.

Solution: Add LaTex math rendering to the Lexical-based document editor using KaTeX (same library support as the recent issue #129). This can be done with a new plugin (e.g., MathPlugin) with registration and wiring.

lisah2u:
Add MathPlugin with block (`$$...$$`) and inline (`$...$`) math support:
- MathNode and InlineMathNode DecoratorNodes with KaTeX rendering
- MathBlockComponent (display mode, edit toggle) and MathInlineComponent (double-click to edit)
- MultilineElement and TextMatch markdown transformers for round-trip fidelity
- MathDiffHandler for atomic diff handling
- Component picker commands: Math Block and Inline Math
- normalizeMarkdown updated to preserve $$ block content

## Summary

Adds LaTeX math rendering to the Lexical-based document editor using KaTeX. Both block math (`$$...$$`) and inline math (`$...$`) are now parsed from markdown, rendered as KaTeX output, and round-trip correctly on export.

Math from issue #129 and related pull-request was only rendered in the agent chat transcript (via react-markdown + remark-math + rehype-katex). The Lexical editor still treated $ delimiters as plain text.

Closes #137 

What changes here:

New plugin: MathPlugin (packages/runtime/src/editor/plugins/MathPlugin/)
•  MathNode — Block math DecoratorNode for `$$...$$` blocks. Renders centered display-mode KaTeX with an Edit/Done toggle (same UX pattern as Mermaid diagrams).
•  InlineMathNode — Inline math DecoratorNode for `$...$`. Renders inline KaTeX; double-click to edit.
•  MathComponent — Shared React rendering using katex.renderToString() with graceful error fallback. Lazy-loaded to avoid bundling KaTeX when no math is present.
•  MathTransformers — MultilineElementTransformer for block math, TextMatchTransformer for inline math. Block transformer is registered before inline to prevent $$ from being consumed as two $ matches.
•  MathDiffHandler — Treats math nodes as atomic units in the diff system (same pattern as MermaidDiffHandler).

Registration & wiring
•  EditorNodes.ts — MathNode and InlineMathNode added to the node list.
•  registerBuiltinPlugins.ts — Plugin registered with transformers, nodes, and "Math Block" / "Inline Math" component picker commands.
•  DiffPlugin — MathDiffHandler registered in the handler registry.
•  MarkdownTransformers.ts — normalizeMarkdown updated to track $$ blocks (prevents line merging inside math content).

Test updates
•  Updated one pre-existing diff test (coverage-expansion.test.ts) whose inline math assertions changed now that $...$ is parsed into InlineMathNode instead of plain text.
•  All 1931 tests pass. Typecheck clean across all packages.

Known issue: KaTeX fonts blocked in dev mode

In npm run dev, Vite's server.fs.allow list doesn't include packages/runtime/node_modules/, so KaTeX font requests are blocked:
`The request id "/Users/.../packages/runtime/node_modules/katex/dist/fonts/KaTeX_Math-Italic.woff2"
is outside of Vite serving allow list.`

Math still renders (KaTeX falls back to system fonts), but the proper KaTeX fonts don't load in dev. This doesn't affect production builds since fonts are bundled into chunks. Fix options:
1. Add packages/runtime/node_modules to server.fs.allow in the electron Vite config
2. Move the katex/dist/katex.min.css import to a location that resolves through the electron package's node_modules

Not included:

•  Live-typing LaTeX shortcuts (auto-converting as you type)
•  KaTeX macro definitions / preamble
•  Math-specific toolbar buttons

### DIFF
--- a/packages/runtime/src/editor/markdown/MarkdownTransformers.ts
+++ b/packages/runtime/src/editor/markdown/MarkdownTransformers.ts
@@ -213,6 +213,7 @@ const CODE_SINGLE_LINE_REGEX =
   /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
 const TABLE_ROW_REG_EXP = /^(?:\|)(.+)(?:\|)\s?$/;
 const TABLE_ROW_DIVIDER_REG_EXP = /^(\| ?:?-*:? ?)+\|\s?$/;
+const MATH_BLOCK_DELIMITER_REGEX = /^\$\$/;
 
 const createBlockNode = (
   createNode: (match: Array<string>) => ElementNode,
@@ -509,6 +510,7 @@ export function normalizeMarkdown(
 ): string {
   const lines = input.split('\n');
   let inCodeBlock = false;
+  let inMathBlock = false;
   const sanitizedLines: string[] = [];
 
   for (let i = 0; i < lines.length; i++) {
@@ -530,6 +532,19 @@ export function normalizeMarkdown(
 
     // If we are inside a code block, keep the line unchanged
     if (inCodeBlock) {
+      sanitizedLines.push(line);
+      continue;
+    }
+
+    // Detect the start or end of a math block ($$)
+    if (MATH_BLOCK_DELIMITER_REGEX.test(line)) {
+      inMathBlock = !inMathBlock;
+      sanitizedLines.push(line);
+      continue;
+    }
+
+    // If we are inside a math block, keep the line unchanged
+    if (inMathBlock) {
       sanitizedLines.push(line);
       continue;
     }
--- a/packages/runtime/src/editor/nodes/EditorNodes.ts
+++ b/packages/runtime/src/editor/nodes/EditorNodes.ts
@@ -34,6 +34,8 @@ import {BoardColumnHeaderNode} from '../plugins/KanbanBoardPlugin/BoardColumnHea
 import {BoardColumnContentNode} from '../plugins/KanbanBoardPlugin/BoardColumnContentNode';
 import {BoardCardNode} from '../plugins/KanbanBoardPlugin/BoardCardNode';
 import {MermaidNode} from '../plugins/MermaidPlugin/MermaidNode';
+import {MathNode} from '../plugins/MathPlugin/MathNode';
+import {InlineMathNode} from '../plugins/MathPlugin/InlineMathNode';
 
 const EditorNodes: Array<Klass<LexicalNode>> = [
   HeadingNode,
@@ -67,6 +69,8 @@ const EditorNodes: Array<Klass<LexicalNode>> = [
   BoardColumnContentNode,
   BoardCardNode,
   MermaidNode,
+  MathNode,
+  InlineMathNode,
 
     // ThemelessCodeNode,
     // {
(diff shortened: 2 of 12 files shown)


ademczuk:
Hi @lisah2u, the Lexical-side architecture is sound - `DecoratorNode` for both block and inline math is the right shape (atomic, serializable, diff-handler-friendly). The `MathDiffHandler` integration plus `normalizeMarkdown` preserving `$$` blocks shows you thought about round-trip fidelity, which is the part most KaTeX-in-editor implementations get wrong. Three findings worth raising before merge:

**1. The inline-math regex has the same currency-vs-math collision as PR #136, possibly worse.**

```ts
importRegExp: /(?<!\$)\$(?!\$)([^\$\n]+?)\$(?!\$)/,
```

Trace through `"It costs $5 to process $1000"`:
- First `$` at the `$5` position: not preceded by `$`, not followed by `$`. Matches the open delimiter.
- `[^\$\n]+?` lazy-captures `5 to process `.
- Second `$` at the `$1000` position: not followed by `$`. Matches the close delimiter.
- Result: `$5 to process $` becomes an `InlineMathNode` with equation `"5 to process "`, KaTeX renders that as garbled math, then `1000` is plain text.

This is actually MORE permissive than PR #136's `remark-math@6` default, which has heuristics about whitespace-adjacency. Two ways to guard:

- Add a "no whitespace adjacent to delimiter" rule: `/(?<!\$|\w)\$(?!\$|\s)([^\$\n]+?)(?<!\s)\$(?!\$|\w)/`. So `$x^2$` matches (no whitespace touching delimiters) but `$5 to process $1000` doesn't (space inside).
- Or switch inline math to LaTeX-standard `\(...\)` syntax and reserve `$...$` for nothing. Less convenient for users, but eliminates the collision entirely.

The `regExp` for live typing is anchored with `$` (end-of-line), so live typing is safer. The `importRegExp` (used on doc open) is the unsafe one.

**2. `MultilineElementTransformer` for block math - verify the round-trip preserves leading/trailing whitespace inside the math block.**

If a user writes:
```
$$
   E = mc^2
$$
```

Does the export side preserve the leading three spaces inside the equation? KaTeX is whitespace-insensitive in rendering, so display-wise it doesn't matter, but the markdown round-trip should be byte-stable to keep diffs clean. Worth a unit test on the transformer's import to export to import cycle to confirm.

**3. KaTeX dedupe with PR #136.**

If both #138 and #136 land, both pull `katex@^0.16.45` into `packages/runtime/package.json`. npm workspace dedupe should resolve to a single copy at the workspace root - but worth confirming `package-lock.json` shows `node_modules/katex` at the root, not `packages/runtime/node_modules/katex` (which would mean two copies, ~2.4MB total). One-line check: `find . -name "package.json" -path "*/katex/*"` after install. If two copies, hoist to the root via the runtime package's peerDeps.

**Otherwise looks good.** The `MathDiffHandler` registration in `diffHandlerRegistry` is the right pattern for keeping diffs atomic across math nodes (no spurious char-by-char diffs inside equations). The component picker entries (`Math Block`, `Inline Math`) give users a discoverable insertion path. Round-trip via `MultilineElementTransformer` + `TextMatchTransformer` is correct - same shape Lexical uses for code blocks.

(1) is the one to answer before merge. (2) and (3) are cheap to verify and worth the few minutes.

ghinkle:
Thanks for the submission, but I had to rebuild this since I updated Lexical and changed the best practice for offering new nodes to lexical. 

Implemented in 9c48216ad5c3f0099cda2fdd0ffc108220d04fae

It's a bit different implementation, but feel free to provide feedback / tweaks.

PR #140, opened 2026-05-02
fix: use getTime() for TIMESTAMPTZ wakeups and queued prompts

### LINKED ISSUE #139: [Bug]: TIMESTAMPTZ values are being double-shifted in wakeup and queued prompt stores
## Describe the bug

I ran into a timezone handling bug in the Electron PGLite stores for session wakeups and queued prompts.

In both of these files:

- `packages/electron/src/main/services/PGLiteSessionWakeupsStore.ts`
- `packages/electron/src/main/services/PGLiteQueuedPromptsStore.ts`

`toMillis()` takes a `Date` from PGLite and rebuilds it with `Date.UTC(...)` using `getFullYear()`, `getHours()`, and the other local-time getters.

That has the effect of taking a real instant in time and reinterpreting its local calendar components as UTC. In practice, the value gets shifted by the machine’s timezone offset.

A concrete example: for someone in UTC-7, a `fire_at` value scheduled 2 minutes in the future can be computed as roughly 7 hours in the past, which makes the wakeup scheduler treat it as due immediately.

What made this stand out is that the repo already documents the opposite expectation elsewhere: the Electron package guidance and `worker.js` both say these columns use `TIMESTAMPTZ`, and that PGLite returns `Date` objects that already represent the correct instant.

## What I expected

I expected `TIMESTAMPTZ` values coming back from PGLite to be converted with `date.getTime()` for `Date` instances, with string parsing only as a fallback if needed.

## Steps to reproduce

1. Use the current main branch.
2. Schedule a wakeup a short time in the future on a machine that is not running in UTC.
3. Let the row round-trip through the database and back into `PGLiteSessionWakeupsStore`.
4. Observe that `toMillis()` reconstructs the `Date` with `Date.UTC(...)`.
5. The computed epoch time is shifted by the local timezone offset instead of preserving the original instant.

The same pattern is present in `PGLiteQueuedPromptsStore`.

## Logs / screenshots

I do not have a screenshot attached, but the relevant code paths are easy to point to:

- `PGLiteSessionWakeupsStore.ts` uses `Date.UTC(...)` inside `toMillis()`
- `PGLiteQueuedPromptsStore.ts` does the same
- `packages/electron/CLAUDE.md` and `packages/electron/src/main/database/worker.js` both say `TIMESTAMPTZ` values from PGLite should already be correct `Date` objects

## Suggested fix

I already have a minimal fix ready and can open it alongside this issue. If this is the wrong direction for the project, no hard feelings, but I’d rather bring a patch than just drop a complaint at the door.

The change is small:

- for `Date` values, use `date.getTime()`
- keep string parsing only as a defensive fallback
- do not reconstruct `Date` instances with `Date.UTC(...)`

That brings these two stores in line with:

- `PGLiteSessionStore.ts`
- `timestampUtils.ts`
- the `TIMESTAMPTZ` guidance already documented in the Electron package

## Nimbalyst version

Current `main` at `097eee1e62fb`

## Operating system

macOS

Dialectician:
## Summary

Fix timezone handling in the Electron PGLite wakeup and queued prompt stores by treating `TIMESTAMPTZ` `Date` values as real instants and converting them with `getTime()`.

## Related issue

Closes #139

## Testing

Validated the staged patch with `git diff --check`.

I did not run automated tests here because this is a narrow timestamp-conversion fix and I did not find targeted test coverage for these two store helpers.

## Notes for reviewers

The root cause is that both stores were reconstructing `Date` values with `Date.UTC(...)` using local-time getters such as `getHours()` and `getFullYear()`. That reinterprets local calendar components as UTC and shifts the instant by the machine timezone offset.

This PR brings those two stores back in line with the repo's existing `TIMESTAMPTZ` guidance and the `getTime()` approach already used elsewhere in the Electron package.

### DIFF
--- a/packages/electron/src/main/services/PGLiteQueuedPromptsStore.ts
+++ b/packages/electron/src/main/services/PGLiteQueuedPromptsStore.ts
@@ -73,26 +73,16 @@ type PGliteLike = {
 
 type EnsureReadyFn = () => Promise<void>;
 
-/**
- * Convert a Date from PGLite to epoch milliseconds.
- * PGLite returns Date objects that are parsed as local time but represent UTC.
- */
 function toMillis(date: Date | string | null | undefined): number {
   if (!date) return 0;
-  if (typeof date === 'string') {
-    return new Date(date).getTime();
+  if (date instanceof Date) {
+    return date.getTime();
   }
-  // PGLite returns Date objects - extract components and treat as UTC
-  const d = date as Date;
-  return Date.UTC(
-    d.getFullYear(),
-    d.getMonth(),
-    d.getDate(),
-    d.getHours(),
-    d.getMinutes(),
-    d.getSeconds(),
-    d.getMilliseconds()
-  );
+  const str = String(date).trim();
+  const hasTimezone =
+    str.endsWith('Z') || str.includes('+') || /-\d{2}:\d{2}$/.test(str);
+  const parsed = new Date(hasTimezone ? str : str.replace(' ', 'T') + 'Z').getTime();
+  return Number.isNaN(parsed) ? 0 : parsed;
 }
 
 function rowToQueuedPrompt(row: any): QueuedPrompt {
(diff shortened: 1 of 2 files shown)


ghinkle:
Thanks. good catch. i really need to centralize this timezone handling at some point.

PR #142, opened 2026-05-02
fix: only forward child session:completed to parent on terminal idle

### LINKED ISSUE #141: [Bug]: meta-agent forwards every child idle turn as a [Child Session Update] to the parent
## Describe the bug

When `MetaAgentService` forwards a child's lifecycle event to the parent as a `[Child Session Update]` queued prompt, it forwards every `session:completed` event. `SessionStateManager.endSession()` emits `session:completed` on every idle transition (the SDK is idle between turns, not just at terminal completion), so a child that processes N queued prompts produces N parent notifications instead of one.

There is also a `notificationSignatures` map intended as a dedup gate (`packages/electron/src/main/services/MetaAgentService.ts`), but the previous signature mixed in an always-incrementing counter and was never compared against the previous value, so it never collapsed anything.

The two defects compound. A parent that queues three back-to-back prompts on a child gets three `[Child Session Update]` notifications, each one of which is itself a queued prompt that costs the parent a model turn and tokens.

## What I expected

The parent should receive one `[Child Session Update]` per genuinely terminal idle (queue empty after the last queued prompt completes), and one per `session:error` / `session:waiting` / `session:interrupted`. Repeated identical-state events for the same child should collapse.

## Steps to reproduce

1. From a parent session, call `mcp__nimbalyst-meta-agent__spawn_session` to create a child with a short initial prompt (e.g. `reply with hello world`).
2. Immediately call `mcp__nimbalyst-meta-agent__send_prompt` twice more on the child with different short prompts.
3. Wait for the child to drain its queue.
4. Inspect `queued_prompts` for the parent session id.

Observed: three rows whose prompt starts with `[Child Session Update]` for that child, queued in tight succession after the child finishes.

Expected: one row.

## Logs / screenshots

DB-grounded evidence from a 14-day window on a packaged 0.58.17 instance: 10 distinct duplicate-prompt clusters in `queued_prompts`, 9 of them `[Child Session Update]`. Worst observed cluster was 5 identical notifications for the same child session over 632 seconds.

The relevant code paths:

- `packages/electron/src/main/services/MetaAgentService.ts` -- `start()` subscribes to `session:completed` / `session:error` / `session:waiting` / `session:interrupted` and forwards each via `handleChildSessionEvent`.
- `packages/electron/src/main/services/MetaAgentService.ts` -- `handleChildSessionEvent` queues a `[Child Session Update]` prompt for the parent on every event without consulting the child's queue depth.
- `packages/runtime/src/ai/server/SessionStateManager.ts` -- `endSession()` correctly emits on every idle transition; other consumers (e.g. `sessionProcessingAtom` in the renderer) need that, so the fix has to live downstream in `MetaAgentService`.

## Suggested fix

I have a fix ready and am opening it alongside this issue. The change is in `packages/electron/src/main/services/MetaAgentService.ts` plus a small new helper file:

- On `session:completed`, count `queued_prompts` rows for the child where `status = 'pending'`. If there are any, the just-finished turn is not the terminal one -- another `session:completed` will fire after the queue drains. Suppress this one. The currently-executing prompt is intentionally excluded from the count because it is the cause of the event we are processing; otherwise every notification (including the final terminal one) would be silenced.
- Repair the dedup gate: drop the always-incrementing counter from the signature, include `errorMessage`, and compare against the previously stored signature so identical-state duplicates collapse.
- Reset the dedup signature on `session:started` / `session:streaming` so distinct turns whose final text happens to match (e.g. both end with `done`) still each notify the parent.
- Other event types (`session:error`, `session:waiting`, `session:interrupted`) bypass the queue-empty gate.

Extracted the signature helper to its own file with vitest unit coverage so the dedup contract is regression-protected.

## Nimbalyst version

Reproduces against `main` at `3a2706cc`.

## Operating system

macOS

Dialectician:
## Summary

`MetaAgentService` was forwarding every child idle transition to the parent as a `[Child Session Update]` queued prompt. A child running through N queued prompts produced N parent notifications instead of one. A second defect in the dedup signature meant identical-state duplicates were never collapsed.

This PR adds a queue-empty gate on `session:completed` so between-turn idles are suppressed (the parent only hears about the genuinely terminal idle), and repairs the dedup signature so identical-state events collapse while distinct turns still each notify.

## Related issue

Closes #141

## Testing

- Added `packages/electron/src/main/services/__tests__/metaAgentNotificationSignature.test.ts` covering the dedup signature contract: discrimination by event type, status, pending prompt id, lastResponse, and errorMessage; identical-component collapse; null/undefined parity.
- Full vitest suite: `npx vitest run` -- 1990 passed, 26 skipped, 0 failures.
- Typecheck: `cd packages/electron && npx tsc --noEmit` -- clean.
- Live reproduction on dev mode with an isolated userData dir: a 3-prompt back-to-back parent -> child interaction now produces exactly one `[Child Session Update]` row in `queued_prompts` for the parent, on the genuinely terminal idle. Pre-fix on packaged 0.58.17 the same scenario produced 3.

## Notes for reviewers

The queue-empty gate counts only `queued_prompts.status = 'pending'`. The currently-executing prompt is intentionally excluded because at the moment `session:completed` fires the prompt that just finished is still in `executing` -- `MessageStreamingHandler` marks it `completed` only after `endSession()` returns. Counting `executing` would include the cause of the event we are processing and silence every notification, including the final terminal one. (This was the failure mode of an earlier internal iteration; the test on dev mode caught it before this PR.)

The signature is intentionally per-turn rather than per-session-lifetime. Without the reset on `session:started` / `session:streaming`, two consecutive turns whose final assistant text happens to match (e.g. both ended with `done`) would silence the second notification. Including `errorMessage` is the same idea: two distinct errors with the same `lastResponse` text must not be collapsed.

Other event types (`session:error`, `session:waiting`, `session:interrupted`) bypass the queue-empty gate -- they are always meaningful or action-required.

The signature helper is a standalone file with no Electron / runtime dependencies so it can be unit-tested without mocking the world.

The DB-grounded evidence from a 14-day window on a packaged 0.58.17 instance: 10 duplicate-prompt clusters in `queued_prompts`, 9 of them `[Child Session Update]`. Worst case was 5 identical notifications for one child session over 632 seconds.

### DIFF
--- a/packages/electron/src/main/services/MetaAgentService.ts
+++ b/packages/electron/src/main/services/MetaAgentService.ts
@@ -89,7 +89,6 @@ export class MetaAgentService {
   private sessionManager: SessionManager | null = null;
   private unsubscribeStateListener: (() => void) | null = null;
   private notificationSignatures = new Map<string, string>();
-  private notificationCounter = 0;
   private ipcHandlersRegistered = false;
 
   private constructor() {}
@@ -191,7 +190,6 @@ export class MetaAgentService {
     this.unsubscribeStateListener?.();
     this.unsubscribeStateListener = null;
     this.notificationSignatures.clear();
-    this.notificationCounter = 0;
     await shutdownMetaAgentServer();
     ClaudeCodeProvider.setMetaAgentServerPort(null);
     OpenAICodexProvider.setMetaAgentServerPort(null);
@@ -706,12 +704,45 @@ export class MetaAgentService {
         return;
       }
 
+      // NIM-6: session:completed fires on every turn idle, not only on terminal
+      // completion. If the child still has more prompts queued AFTER the one
+      // that just finished, this idle is a between-turn pause -- another
+      // session:completed will follow once the queue drains. Suppress it; the
+      // parent will be notified on the genuinely terminal idle (queue empty).
+      //
+      // The just-finished prompt is still in `executing` status at the moment
+      // session:completed fires (MessageStreamingHandler marks it `completed`
+      // only after endSession returns). So we count only `pending` rows --
+      // counting `executing` would include the current turn itself and
+      // suppress every notification, including the final terminal one.
+      //
+      // The other event types (error/waiting/interrupted) are always
+      // meaningful and pass through.
+      if (eventType === 'session:completed') {
+        const { rows: pendingRows } = await databaseWorker.query<{ count: string }>(
+          `SELECT COUNT(*)::text AS count FROM queued_prompts
+           WHERE session_id = $1 AND status = 'pending'`,
+          [sessionId]
+        );
+        const pendingCount = Number(pendingRows[0]?.count ?? '0');
+        if (pendingCount > 0) {
+          return;
+        }
+      }
+
       const metaStatusRow = await this.getSessionStatusRow(metaSession.id, metaSession.workspacePath);
       const metaStatus = (metaStatusRow?.status || 'idle') as SessionStatusValue;
 
       const result = await this.buildSessionResultData(sessionId, session.workspacePath);
-      this.notificationCounter += 1;
-      const signature = `${eventType}:${result.status}:${result.lastActivity || 0}:${result.pendingPrompt?.promptId || ''}:${this.notificationCounter}`;
+
+      // NIM-6: real dedup gate. Drop notifications whose semantic content is
+      // identical to the last one delivered for this child. The previous code
+      // mixed in an always-incrementing counter, which made every signature
+      // unique and the dedup useless.
+      const signature = `${eventType}:${result.status}:${result.pendingPrompt?.promptId || ''}:${result.lastResponse || ''}`;
+      if (this.notificationSignatures.get(sessionId) === signature) {
+        return;
+      }
       this.notificationSignatures.set(sessionId, signature);
 
       if (result.pendingPrompt?.promptType === 'exit_plan_mode_request') {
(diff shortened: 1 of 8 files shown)


Dialectician:
FYI from validating this PR end-to-end on a packaged 0.58.17 build: observed an adjacent ScheduleWakeup behavior worth flagging separately. When the app launches with persisted wakeups queued, they fire immediately at workspace-open and can cascade through the meta-agent path this PR touches (child wakes → completes → notifies parent). Today's repro was benign (1 wakeup, 12-char child response, parent absorbed cleanly), and this PR's dedup gate behaved correctly. But the UX of unsolicited LLM turns at app open feels like it deserves its own product decision (silent fire vs. user-gated vs. age-out). Out of scope for this PR; flagging so it's not lost. Peace , B.Sweet

Dialectician:
Follow-up is pushed in `89ee47cf`.

This keeps #142 narrowly scoped to the runaway notification bug:

- direct human input into a child now disables further parent notifications for that child
- any already-pending `[Child Session Update]` prompt for that parent/child pair is deleted

I also ran a live dev repro against both the original looping session and fresh control/race cases:

- normal orchestrated child without takeover still notifies the parent
- direct child takeover stops future parent updates
- takeover also clears an already-pending parent child-update row
- I could not reproduce a leak in the race case where takeover happens while the child is still running

I opened a separate issue for the broader product question around parent/child ownership and hand-back semantics so we can discuss direction without blocking this fix.

PR #143, opened 2026-05-03
fix: discover actual LM Studio model ID instead of hardcoded "local-model"

ArjunDivecha:
## Summary

- LM Studio no longer accepts `"local-model"` as a generic alias in chat completions — it requires the actual downloaded model name (e.g., `qwen3.6-27b-ud-mlx`)
- This caused the error: `Invalid model identifier "local-model"`

## Fix

**`LMStudioProvider.ts`** — three changes:

1. **`initialize()` now discovers the real model** from `/v1/models` and stores it in a new `resolvedModel` field  
2. **`sendMessage()` uses the discovered model** — `model: this.resolvedModel || this.config.model || 'local-model'`, so the auto-discovered model name takes priority  
3. **`getDefaultModels()` returns `[]`** — removed the non-functional `local-model` placeholder

## Test plan

- [ ] Start LM Studio, load a model, start local server
- [ ] In Nimbalyst, configure LM Studio provider  
- [ ] Start an LM Studio session — it should auto-detect the actual model name
- [ ] Verify console shows `[LMStudio] Resolved model: <actual-model-name>`

### DIFF
--- a/packages/runtime/src/ai/server/providers/LMStudioProvider.ts
+++ b/packages/runtime/src/ai/server/providers/LMStudioProvider.ts
@@ -22,14 +22,17 @@ interface LMStudioConfig extends ProviderConfig {
 export class LMStudioProvider extends BaseAIProvider {
   private baseUrl: string = 'http://127.0.0.1:8234';
   private abortController: AbortController | null = null;
-  
+  private resolvedModel: string | null = null;
+
+  // DEFAULT_MODEL is a UI placeholder — the actual model is resolved from
+  // LM Studio's /v1/models endpoint during initialize() and stored in resolvedModel.
   static readonly DEFAULT_MODEL = 'lmstudio:local-model';
 
   async initialize(config: LMStudioConfig): Promise<void> {
     this.config = config;
     this.baseUrl = config.baseUrl || 'http://127.0.0.1:8234';
-    
-    // Test connection to LMStudio
+
+    // Test connection and discover the actual loaded model
     try {
       const response = await fetch(`${this.baseUrl}/v1/models`, {
         signal: AbortSignal.timeout(5000) // 5 second timeout
@@ -37,11 +40,20 @@ export class LMStudioProvider extends BaseAIProvider {
       if (!response.ok) {
         throw new Error(`LMStudio server not responding at ${this.baseUrl}. Please ensure LMStudio is running and has a model loaded.`);
       }
+
+      // Discover the actual model ID from LM Studio instead of using "local-model"
+      const data = await response.json();
+      if (data.data && data.data.length > 0) {
+        this.resolvedModel = data.data[0].id;
+        console.log(`[LMStudio] Resolved model: ${this.resolvedModel}`);
+      } else {
+        throw new Error(`LMStudio is running but no model is loaded. Please load a model in LMStudio first.`);
+      }
     } catch (error: any) {
       if (error.cause?.code === 'ECONNREFUSED' || error.message?.includes('fetch failed')) {
         throw new Error(`Cannot connect to LMStudio at ${this.baseUrl}. Please ensure:\n1. LMStudio is running\n2. A model is loaded in LMStudio\n3. The local server is started (look for "Local Server" in LMStudio)`);
       }
-      throw new Error(`Failed to connect to LMStudio at ${this.baseUrl}: ${error.message || error}`);
+      throw error;
     }
   }
 
@@ -201,7 +213,7 @@ export class LMStudioProvider extends BaseAIProvider {
 
     // Log the request for debugging
     const requestBody: any = {
-      model: this.config.model || 'local-model',
+      model: this.resolvedModel || this.config.model || 'local-model',
       messages: apiMessages,
       max_tokens: this.config.maxTokens || 4096,
       temperature: this.config.temperature || 0.7,
@@ -802,13 +814,7 @@ export class LMStudioProvider extends BaseAIProvider {
    * Get default models
    */
   static getDefaultModels(): AIModel[] {
-    return [{
-      id: ModelIdentifier.create('lmstudio', 'local-model').combined,
-      name: 'Local Model',
-      provider: 'lmstudio' as const,
-      maxTokens: 4096,
-      contextWindow: 4096
-    }];
+    return [];
   }
 
   /**


ArjunDivecha:
Works fine aftere the fix

ghinkle:
This looks directionally right to me. Resolving the actual LM Studio model ID instead of sending the hardcoded `local-model` placeholder seems like the correct fix for the common case where LM Studio has a model loaded and Nimbalyst should just use it.

One thing I want to call out before merging: I think LM Studio can expose multiple loaded models at once, and this change now prefers `data.data[0].id` over the model Nimbalyst/session selected. That means in a multi-model setup we may silently use “first model from `/v1/models`” instead of the one the user picked in settings or the session picker. I’m not sure how often that matters in practice, but I'd welcome feedback from people who run LM Studio with multiple models loaded at the same time.

There’s also a smaller UX edge case now where the connection test can still report success if `/v1/models` returns `200` but no model is loaded, while provider init will fail. That feels less important than the multi-model question, but worth keeping in mind.

I’d be in favor of merging if the intended behavior is “use whatever LM Studio currently has loaded” and we confirm that multi-model selection either doesn’t matter or isn’t something we support yet. If users do rely on selecting among multiple loaded models, we should preserve `config.model` precedence and only fall back to the discovered model when the session is still on the placeholder.

ademczuk:
Hi @ArjunDivecha, this matches the failure mode well and the auto-discovery shape is right. Greg already gave directional approval, so this is just adding depth. Four small things worth checking before merge:

**1. The first-model selection is arbitrary in multi-model setups.**

```ts
this.resolvedModel = data.data[0].id;
```

LM Studio supports loading multiple models at once. `data.data[0]` picks whichever the server lists first, which isn't necessarily the one the user expects. Two reasonable shapes:

- Pick by `config.model` if it appears in the discovered list, fall back to `data.data[0]` if not. That respects an explicit user choice.
- Or accept the first-model behavior but add a code comment so future readers understand the trade-off.

For v1, picking the first is fine if the README/help text mentions it. For v2 it should probably be configurable.

**2. Error-wrapping removed at the bottom of the catch.**

```ts
} catch (error: any) {
  if (error.cause?.code === 'ECONNREFUSED' || ...) {
    throw new Error(`Cannot connect to LMStudio at ${this.baseUrl}. ...`);
  }
- throw new Error(`Failed to connect to LMStudio at ${this.baseUrl}: ${error.message || error}`);
+ throw error;
}
```

The previous form wrapped non-ECONNREFUSED errors with a `Failed to connect to LMStudio at <url>: <message>` prefix. Re-throwing the bare error loses that user-facing context. The new "no model loaded" error you added (good catch, by the way) IS already user-facing, but a generic `TypeError: fetch failed` from some other path would surface to the user without the LM Studio context. Probably worth keeping the wrapping for the non-specific-case path, or adding more `if` branches that throw user-facing errors so the bare `throw error` only fires for genuinely-unexpected cases.

**3. The `'local-model'` fallback in `sendMessage` is now reachable only if initialize is bypassed.**

```ts
model: this.resolvedModel || this.config.model || 'local-model',
```

After your PR, `initialize()` either succeeds with a non-null `resolvedModel` or throws. So if we get to `sendMessage`, `resolvedModel` should be set. The `|| 'local-model'` is defensive but also silently falls back to the broken value the PR is removing elsewhere. If initialize was bypassed (programming error), it'd be cleaner to throw `'LMStudioProvider not initialized'` than to send `local-model` and fail server-side with a confusing error. Or just drop the third fallback and let `model: undefined` produce a clearer downstream failure.

**4. Stale `resolvedModel` if the user swaps the loaded model in LM Studio mid-session.**

The discovery happens once at `initialize()`. If the user unloads the model and loads a different one in LM Studio while the Nimbalyst session is open, subsequent `sendMessage` calls send the stale model ID, and LM Studio returns "Invalid model identifier" again. Probably out of scope for this PR but worth a code comment so future-you doesn't re-debug it. Could be addressed later by re-resolving on first 4xx-with-model-mismatch from `/v1/chat/completions`.

**Minor**: the `console.log('[LMStudio] Resolved model: ...')` could use `logger.main.info` to match the centralized logger pattern documented in `CLAUDE.md` and used by other providers. Small consistency thing.

Good fix overall. (1) and (2) are the ones I'd want answered before merge, (3) and (4) are follow-up territory.

ademczuk:
Quick verification pass on this one. Good news first: the fix is still needed. On current main, `LMStudioProvider.ts` has no model-discovery logic at all. `sendMessage` still sends `model: this.config.model || 'local-model'` and `getDefaultModels()` still returns the `local-model` placeholder. The file has not been touched on upstream since 2026-04-02, before this PR, so nothing has superseded it. The branch is also clean: GitHub reports it mergeable, only one file changed, and both CI checks (TypeScript Check, Unit Tests) are passing.

The one thing holding it up is the open question Greg raised, not a rebase. @ghinkle gave directional approval but flagged that preferring `data.data[0].id` silently overrides the user-selected `config.model` when LM Studio has multiple models loaded. The merge condition was to use the discovered model only as a fallback, not always prefer the first model the server lists.

@ArjunDivecha, would you be open to adjusting the discovery so it picks `config.model` when it appears in the `/v1/models` list and falls back to the first model only when it does not? That, plus restoring the LM Studio context prefix on the non-ECONNREFUSED error path (the catch currently re-throws the bare error), should clear the way to merge. Happy to help if useful. This also fits under the LM Studio Improvements tracking issue (#97).

ademczuk (approved):
Did a full pass on this against current main. The fix is correct and still needed: LMStudioProvider.ts on main has no model-discovery, so it still sends the local-model placeholder that LM Studio now rejects. The /v1/models discovery in initialize() plus resolvedModel || config.model || local-model in sendMessage() is the right shape, and returning [] from getDefaultModels() matches the repo convention of dynamic model fetching with no hardcoded models. CI is green (Unit Tests + TypeScript Check), no conflicts, and @ArjunDivecha confirmed it works after the fix.

One non-blocking note for whoever merges: data.data[0].id takes the first loaded model, so if someone has several models loaded in LM Studio at once it picks the first rather than a user-chosen one. Fine for the common single-model case and strictly better than the old placeholder, just flagging it in case multi-model selection comes up later.

Approving to give it a clear sign-off.

PR #145, opened 2026-05-03
fix: extract Codex child output in meta-agent session results

### LINKED ISSUE #144: Meta-agent get_session_result returns empty fields for OpenAI Codex children
## What's broken

`get_session_result` for a meta-agent child returns null `lastResponse`, empty `recentMessages`, empty `userPrompts`, and null `originalPrompt` whenever the child runs on `openai-codex` (and probably opencode too). The `[Child Session Update]` notifications the parent receives are similarly empty -- no recent messages, no last response. From the parent's API surface the child looks like it never said anything.

## Repro

Spawn a Codex child:

```
mcp__nimbalyst-meta-agent__spawn_session({
  prompt: "In 2 sentences, what does the @nimbalyst/runtime package do? Then say DONE.",
  model: "openai-codex:gpt-5.4",
  isolated: true,
  notifyOnComplete: true
})
```

Wait for completion, then:

```
mcp__nimbalyst-meta-agent__get_session_result({ sessionId })
```

You get back:

```json
{
  "provider": "openai-codex",
  "lastResponse": null,
  "recentMessages": [],
  "userPrompts": [],
  "originalPrompt": null
}
```

Same shape for every Codex child. Claude Code children work fine.

## Root cause

`packages/electron/src/main/services/MetaAgentService.ts` has two parallel gaps:

1. `extractMessageText` only knows Claude's `{type:'assistant', message:{content:[...]}}` shape. Codex stores raw SDK events, so its assistant text lives at `{type:'item.completed', item:{type:'agent_message', text}}` or `{type:'task_complete', last_agent_message}`. None of those match.
2. `extractUserPrompts` does `JSON.parse(content).prompt`. `OpenAICodexProvider.ts` logs Codex inputs as the raw prompt string (not JSON-wrapped), so every input row throws and gets skipped.

Knock-on effect: the dedup signature in `handleChildSessionEvent` (#142) is built from `lastResponse`, so the always-null value makes the signature much less discriminating for Codex children.

## Severity

Medium. Doesn't surface in the user UI, but anyone using meta-agent flows with Codex children loses all parent-side visibility into the child's work. With Codex now a first-class agent provider, that's a real gap.

## Fix

PR coming as a stack on top of #142 (the dedup work shares an import). I'll link it here.

peace, B.Sweet

Dialectician:
## Summary

Meta-agent parents can't see the output of OpenAI Codex children. `get_session_result` returns null `lastResponse` and empty `recentMessages` / `userPrompts` / `originalPrompt`, and the `[Child Session Update]` notifications the parent receives are empty too.

Closes #144.

> Stacked on #142. NIM-11 was developed against the post-#142 `MetaAgentService` and shares the `./metaAgentNotificationSignature` import added there. Please merge #142 first; I'll rebase onto `main` once it lands.

## Two parallel bugs in `MetaAgentService.ts`

1. `extractMessageText` only handled Claude's `{type:'assistant', message:{content:[...]}}` shape. Codex stores raw SDK events, so its assistant text is at `{type:'item.completed', item:{type:'agent_message', text}}` and `{type:'task_complete', last_agent_message}`. Neither matched.
2. `extractUserPrompts` did `JSON.parse(content).prompt`. `OpenAICodexProvider.ts:779` logs Codex inputs as the raw prompt string, not `JSON.stringify({prompt})`. Every input row threw and was skipped.

## What changed

- Pulled both functions into a new `metaAgentMessageText.ts` so they're unit-testable on their own.
- Added the Codex shapes to `extractMessageText`: `task_complete`, `item.completed`/`item.updated` with `agent_message` or `reasoning` items, `event_msg` payloads, root-level `delta`, bare `text`. Tool-only events like `command_execution` and `mcp_tool_call` still return null.
- `extractMessageText` now takes optional `metadata` and skips rows where `metadata.promptType === 'system_reminder'` so session-naming nudges stop leaking into parent notifications.
- `MetaAgentService.extractRecentMessages` and `extractLastAgentResponse` pass `message.metadata` through so the filter actually fires.

## Test plan

- [x] 27 new unit tests in `__tests__/metaAgentMessageText.test.ts` covering Claude shapes, every Codex shape above, system-reminder filtering, and the JSON-with-empty-prompt-vs-raw-text edge case.
- [x] Existing `metaAgentNotificationSignature.test.ts` still passes.
- [x] Full electron service suite: 193 pass, 3 skipped, 0 fail.
- [x] `cd packages/electron && npx tsc --noEmit` -- clean.

## Validation

Built a packaged app locally from `upstream/main` + #142 + this PR. Spawned a Codex child (`openai-codex:gpt-5.4`, `isolated: true`, `notifyOnComplete: true`) with the prompt *"In 2 sentences, what does the @nimbalyst/runtime package do? Then say DONE."*

`get_session_result` after completion:

- `provider: "openai-codex"`
- `lastResponse`: the full assistant answer
- `recentMessages`: 3 entries, all assistant prose, no `<SYSTEM_REMINDER>` rows
- `originalPrompt` and `userPrompts`: the original prompt

I also queried `ai_agent_messages` directly for an older failing Codex session to confirm what the on-disk shapes actually look like: inputs are raw plain-text strings with `metadata.mode = 'agent'` (or `metadata.promptType = 'system_reminder'`), outputs are `JSON.stringify(rawCodexEvent)` with `metadata.codexProvider = true`, and assistant text always sits at `item.text` for `agent_message` items.

## Notes for reviewers

The helper handles only text extraction -- it deliberately does not import `parseCodexEvent` or try to handle tool calls, usage, or reasoning beyond text. The full canonical parser lives at `packages/runtime/src/ai/server/providers/codex/codexEventParser.ts`; this helper just covers the text-bearing subset of shapes. If new Codex shapes get added there, they probably want adding here too.

peace, B.Sweet

### DIFF
--- a/packages/electron/src/main/services/MetaAgentService.ts
+++ b/packages/electron/src/main/services/MetaAgentService.ts
@@ -21,6 +21,7 @@ import {
   shutdownMetaAgentServer,
 } from '../mcp/metaAgentServer';
 import { computeNotificationSignature } from './metaAgentNotificationSignature';
+import { extractMessageText, extractUserPrompts } from './metaAgentMessageText';
 
 type SessionStatusValue = 'idle' | 'running' | 'waiting_for_input' | 'error' | 'interrupted';
 type PromptType = 'permission_request' | 'ask_user_question_request' | 'exit_plan_mode_request';
@@ -882,7 +883,7 @@ export class MetaAgentService {
     }
 
     const messages = await AgentMessagesRepository.list(sessionId, { limit: 500 });
-    const userPrompts = this.extractUserPrompts(messages);
+    const userPrompts = extractUserPrompts(messages);
     const recentMessages = this.extractRecentMessages(messages, 3);
     const pendingPrompt = await this.getPendingInteractivePrompt(sessionId);
 
@@ -1116,27 +1117,11 @@ export class MetaAgentService {
     }
   }
 
-  private extractUserPrompts(messages: Array<{ direction: string; content: string }>): string[] {
-    const prompts: string[] = [];
-    for (const message of messages) {
-      if (message.direction !== 'input') continue;
-      try {
-        const parsed = JSON.parse(message.content);
-        if (typeof parsed.prompt === 'string' && parsed.prompt.trim()) {
-          prompts.push(parsed.prompt.trim());
-        }
-      } catch {
-        continue;
-      }
-    }
-    return prompts;
-  }
-
-  private extractLastAgentResponse(messages: Array<{ direction: string; content: string }>, maxLength: number = 500): string | null {
+  private extractLastAgentResponse(messages: Array<{ direction: string; content: string; metadata?: Record<string, unknown> | null }>, maxLength: number = 500): string | null {
     for (let index = messages.length - 1; index >= 0; index -= 1) {
       const message = messages[index];
       if (message.direction !== 'output') continue;
-      const extracted = this.extractMessageText(message.content);
+      const extracted = extractMessageText(message.content, message.metadata);
       if (extracted) {
         return extracted.length > maxLength ? `${extracted.slice(0, maxLength)}...` : extracted;
       }
@@ -1145,13 +1130,13 @@ export class MetaAgentService {
   }
 
   private extractRecentMessages(
-    messages: Array<{ direction: string; content: string }>,
+    messages: Array<{ direction: string; content: string; metadata?: Record<string, unknown> | null }>,
     limit: number
   ): Array<{ direction: 'input' | 'output'; text: string }> {
     const collected: Array<{ direction: 'input' | 'output'; text: string }> = [];
     for (let index = messages.length - 1; index >= 0 && collected.length < limit; index -= 1) {
       const message = messages[index];
-      const text = this.extractMessageText(message.content);
+      const text = extractMessageText(message.content, message.metadata);
       if (!text) {
         continue;
       }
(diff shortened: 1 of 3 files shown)


Dialectician:
Rebased onto `main` now that #142 has landed. The two child-takeover commits this branch was carrying are part of upstream via #142, so they auto-dropped during rebase. What remains is just the Codex output extraction fix.

All 27 unit tests in `metaAgentMessageText.test.ts` still pass on the rebased branch. The extraction follows the same helper-module pattern #142 introduced for `metaAgentNotificationSignature` and `childSessionTakeover`, so it should fit cleanly alongside that structure.

Conflicts resolved, ready for another look.

Peace B.Sweet

Dialectician:
The failing TypeScript Check isn't from this PR. The two errors are in `ClaudeCodeProvider.ts`, which this PR doesn't touch:

```
../runtime/src/ai/server/providers/ClaudeCodeProvider.ts(94,32): error TS2305:
  Module '"./claudeCode/sdkOptionsBuilder"' has no exported member 'PromptStreamController'.
../runtime/src/ai/server/providers/ClaudeCodeProvider.ts(605,37): error TS2339:
  Property 'promptController' does not exist on type 'BuildSdkOptionsResult'.
```

Reproduces on a clean `main` checkout with no PR changes applied. Filed as #160 with the full root-cause trace (it's an oversight in `ddf92f41` after the earlier `b23282cf` revert).

Unit Tests pass on this branch. Once #160 is resolved I'll re-trigger CI and we should get a clean green.

Peace B.Sweet

Dialectician:
Rebased onto the fixed main (post-5fec84b8 / #160). Both checks green now. Really appreciate the fast turnaround on #160 — twelve minutes from issue filed to fix landed is wild, and the 5s grace-timer approach is a nicer solve than the original revert. Thanks for the great work. Peace B.Sweet

PR #147, opened 2026-05-03
refactor: centralize PGLite TIMESTAMPTZ to epoch-ms via toMillis

Dialectician:
## Summary

Follow up to #140. In the merge thread @ghinkle commented:

> Thanks. good catch. i really need to centralize this timezone handling at some point.

Picking that up here. Scoped tight: single helper consolidation across the existing duplicates, no schema changes.

## What this does

- Makes `packages/electron/src/main/utils/timestampUtils.ts` the only `toMillis()` in the package. Deletes 4 duplicate definitions in `PGLiteQueuedPromptsStore`, `PGLiteSessionWakeupsStore`, `PGLiteSessionStore`, and `PGLiteSessionFileStore`.
- Migrates 9 ad-hoc `new Date(row.x).getTime()` sites in `MetaAgentService.ts`.
- Widens the helper's return type from `number` to `number | null`. The duplicates were silently masking nullable TIMESTAMPTZ columns (`claimed_at`, `completed_at`, `fired_at`) by returning `0` or `Date.now()` for null input. Call sites now thread null through to wire types that already accommodate it (`QueuedPrompt.claimedAt?`, `SessionWakeup.firedAt: number | null`).
- Updates `packages/electron/CLAUDE.md`'s Related-files pointer. It was pointing at one of the deleted duplicates.
- Adds a `timestampUtils` test file. Closes the coverage gap I flagged in #140's notes for reviewers.

## Behavior changes worth surfacing

Three axes of difference get unified across the duplicates, not one. Calling them out so they don't surprise anyone in review:

1. Nullish/invalid input was `0` (post-#140) or `Date.now()`. Now `null`.
2. Numeric input. `PGLiteQueuedPromptsStore` and `PGLiteSessionWakeupsStore` were coercing numbers to strings before parsing. Now passed through directly.
3. Naive string parsing. `PGLiteSessionFileStore` used plain `new Date(value)`, which interprets the string in local time. The canonical helper UTC-normalizes naive strings, so consolidating it silently fixes a latent #139-class bug if that store ever sees string timestamps. It doesn't today, but worth flagging.

Wire types are unchanged. The lie was in the row mappers, not the public types.

## Call site discipline

`!` non-null assertion at NOT NULL column reads. `?? undefined` (or pass-through to `number | null` wire fields) at nullable column reads. Reasoning: schema violations should crash loudly, not paper over with `Date.now()` and silently lie about timestamps.

A handful of `!` sites read columns that are nullable in the schema but always populated via `DEFAULT CURRENT_TIMESTAMP` in practice (`ai_sessions.created_at`, `queued_prompts.created_at`, etc). The assertion encodes the de facto invariant. Tightening the schema to actually be `NOT NULL` is a separate change I didn't want to bundle here.

## Out of scope

A second PR will sweep:

- `ToolCallMatcher.ts` and `ClaudeCodeSessionScanner.ts` ad-hoc sites
- An audit of unconverted read paths in `PGLiteAgentMessagesStore`, `PGLiteDocumentsRepository`, `PGLiteWorkspaceRepository`
- Schema cleanups for columns that should never be null in practice

## Test plan

- [x] `npx vitest run packages/electron/src/main/utils/__tests__/timestampUtils.test.ts`. 5/5 passing.
- [x] `npx tsc --noEmit -p packages/electron`. 0 errors in the diff. Other workspace errors are pre-existing and unrelated.
- [x] Manual smoke. Queue a prompt, schedule a wakeup, confirm timestamps render correctly and the wakeup fires at the expected wall-clock time.

Continues from #140.

Peace B.Sweet

### DIFF
--- a/packages/electron/CLAUDE.md
+++ b/packages/electron/CLAUDE.md
@@ -180,14 +180,15 @@ db.query('INSERT INTO ... VALUES ($1)', [new Date()])
 
 3. **DO**: Retrieve timestamps through `toMillis()` function
 ```typescript
-const createdAt = toMillis(row.created_at);  // Returns epoch milliseconds
+const createdAt = toMillis(row.created_at)!;              // Required timestamp
+const claimedAt = toMillis(row.claimed_at) ?? undefined;  // Nullable timestamp
 ```
 
 4. **DO**: Display with `toLocaleString()` for user's local timezone
 
 **Related files:**
 - `src/main/database/worker.js` - Database schema and comments
-- `src/main/services/PGLiteSessionStore.ts` - toMillis() implementation
+- `src/main/utils/timestampUtils.ts` - canonical toMillis() implementation
 
 ## Renderer State Architecture
 
--- a/packages/electron/src/main/services/MetaAgentService.ts
+++ b/packages/electron/src/main/services/MetaAgentService.ts
@@ -8,6 +8,7 @@ import { ModelIdentifier } from '@nimbalyst/runtime/ai/server/types';
 import { AISessionsRepository, AgentMessagesRepository, SessionFilesRepository } from '@nimbalyst/runtime';
 import { getSessionStateManager } from '@nimbalyst/runtime/ai/server/SessionStateManager';
 import { getDefaultAIModel } from '../utils/store';
+import { toMillis } from '../utils/timestampUtils';
 import { createWorktreeStore } from './WorktreeStore';
 import { GitWorktreeService } from './GitWorktreeService';
 import { database as databaseWorker } from '../database/PGLiteDatabaseWorker';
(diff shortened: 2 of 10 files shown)


Dialectician:
Smoke-tested in a fresh dogfood packaged build off custom/main. All three checks clean:

- Queued prompt timestamps round-trip correctly through the new mapper. Sensible TIMESTAMPTZ values, no anomalies.
- Wakeup firing test (the regression check for #139): scheduled a 2-minute self-wakeup, drift on fire was +17 ms. Far inside any reasonable tolerance.
- Worktree timestamps render normally. created/updated within seconds of each other as expected.

No surprises. Ready when you are.

Peace B.Sweet

ghinkle:
Thanks

PR #162, opened 2026-05-05
fix: read customClaudeCodePath from ai-settings store in path loader

### LINKED ISSUE #58: customClaudeCodePath silently ignored: ClaudeCodeProvider reads from wrong electron-store
## Summary

Nimbalyst's `ClaudeCodeProvider` reads `customClaudeCodePath` and `useStandaloneBinary` from `app-settings.json` via `getAppStore()`, but the Settings UI saves them to `ai-settings.json` via `getSettingsStore()`. The two files are managed by different `electron-store` instances and share no state. As a result, **setting "Custom Claude Installation" via the Settings UI has no effect at startup** — the init code reads the field as `(empty)` even though `ai-settings.json` on disk has the correct value.

This makes the "Use Standalone Binary" toggle and "Custom Claude Installation" path field completely non-functional via the UI. Likely the upstream cause of the symptom in [NIM-761](https://linear.app/nimbalyst/issue/NIM-761/claude-code-exits-with-code-1-on-windows-works-in-cli-fails-in) (Claude Code exits with code 1 on Windows) where the user reported that "forcing standalone binary via config" had no effect.

## Environment

* **Nimbalyst**: 0.56.17
* **macOS**: 26.3.2 (Tahoe) — Apple Silicon
* **Likely also affects**: Windows (per [NIM-761](https://linear.app/nimbalyst/issue/NIM-761/claude-code-exits-with-code-1-on-windows-works-in-cli-fails-in) — same fall-through-to-Electron-binary symptom)

## Steps to Reproduce

1. Install Nimbalyst (any platform with Settings → Claude Code panel)
2. Open Settings → Claude Code
3. Toggle "Use Standalone Binary" ON
4. Set "Custom Claude Installation" path to any wrapper or alternative CLI (e.g., `/usr/local/bin/claude`, or a wrapper script that prepends env vars)
5. Save and quit Nimbalyst
6. Verify on disk: `~/Library/Application Support/@nimbalyst/electron/ai-settings.json` contains your custom path
7. Relaunch Nimbalyst
8. Open `~/Library/Application Support/@nimbalyst/electron/logs/main.log` and grep for `[ClaudeCodeProvider] Initialized settings`

**Expected**:

```
[ClaudeCodeProvider] Initialized settings { useStandaloneBinary: true, customClaudeCodePath: '(set)' }
```

**Actual**:

```
[ClaudeCodeProvider] Initialized settings { useStandaloneBinary: false, customClaudeCodePath: '(empty)' }
```

The custom path you set in the UI is never invoked at runtime. The spawn falls through to the bundled Electron-binary path (`Using Electron binary for Claude Code (standalone not enabled)`).

## Root Cause

The Settings UI `ai:saveSettings` IPC handler writes `customClaudeCodePath` and `useStandaloneBinary` to `ai-settings.json` via the AI settings electron-store:

* **Writer**: `out/main/index.js:L460199-L460205`
* **Store factory**: `getSettingsStore()` at `out/main/index.js:L457140-L457142`
* **File written**: `~/Library/Application Support/@nimbalyst/electron/ai-settings.json`

The `ClaudeCodeProvider.init()` code reads these fields from `app-settings.json` via the **app** electron-store:

* **Reader**: `out/main/index.js:L477452-L477457`
* **Store factory**: `getAppStore()` at `out/main/index.js:L13188-L13207`
* **File read**: `~/Library/Application Support/@nimbalyst/electron/app-settings.json`

The two factories return different electron-store instances pointing at different files. Whatever the UI writes is invisible to `ClaudeCodeProvider` because they share no state.

**Direct reproduction with explicit cross-session evidence**:

I observed the bug across two separate Nimbalyst launches on different days:

```
# main.1.log (rotated, from 2026-04-14 21:46:28):
[ClaudeCodeProvider] Initialized settings { useStandaloneBinary: false, customClaudeCodePath: '(empty)' }

# main.log (current session, 2026-04-15 07:04:53):
[ClaudeCodeProvider] Initialized settings { useStandaloneBinary: false, customClaudeCodePath: '(empty)' }
```

Both times, on-disk state at the moment of init contained the value:

```bash
$ cat ~/Library/Application Support/@nimbalyst/electron/ai-settings.json | grep customClaudeCodePath
"customClaudeCodePath": "/Users/<user>/.local/bin/claude-smart"
```

I verified the bug by writing the value to `app-settings.json` instead and confirming the init then read it correctly:

```
[ClaudeCodeProvider] Initialized settings { useStandaloneBinary: true, customClaudeCodePath: '(set)' }
```

This confirms the reader is looking at `app-settings.json`, not `ai-settings.json`.

## Impact

* **"Use Standalone Binary" toggle and "Custom Claude Installation" path field are non-functional** for any user setting them via the UI
* Affects users who need:
  * Corporate SSO wrapper scripts
  * Multi-profile routing (e.g., `~/.claude` vs `~/.claude-work` based on `$PWD`)
  * Dev/beta builds of Claude Code
  * Alternative Claude Code installations (Homebrew, npm global, etc.)
* Likely the upstream root cause of [NIM-761](https://linear.app/nimbalyst/issue/NIM-761/claude-code-exits-with-code-1-on-windows-works-in-cli-fails-in) (Claude Code exits with code 1 on Windows) — that user reported "Attempted forcing standalone binary via config, but behavior did not change", which is exactly this symptom

## Suggested Fix

The cleanest fix is to make the reader at `L477452-L477457` use `getSettingsStore()` (the AI settings store) instead of `getAppStore()`, matching the writer:

```diff
- const aiStore = getAppStore();
+ const aiStore = getSettingsStore();
  this.useStandaloneBinary = aiStore.get("useStandaloneBinary", false);
  this.customClaudeCodePath = aiStore.get("customClaudeCodePath", "");
```

(Pseudocode — actual minified line is harder to show.)

Alternatively, move the writer to use `getAppStore()` to match the reader. Either direction works as long as both code paths agree on which store holds these fields.

A third option: hoist `customClaudeCodePath` and `useStandaloneBinary` into a shared schema that both stores synchronize on, but this is more invasive.

## Workaround (verified working)

Set the fields in **both** `ai-settings.json` AND `app-settings.json` via direct file edit (NOT via the UI):

```bash
python3 << 'PY'
import json, pathlib
for fname in ['ai-settings.json', 'app-settings.json']:
    p = pathlib.Path.home() / 'Library/Application Support/@nimbalyst/electron' / fname
    d = json.loads(p.read_text())
    d['customClaudeCodePath'] = '/Users/<user>/.local/bin/claude-smart'  # or your custom path
    d['useStandaloneBinary'] = True
    p.write_text(json.dumps(d, indent='\t'))
    print(f'updated: {p}')
PY
```

Both stores will then agree. After restart, `ClaudeCodeProvider` reads the value from `app-settings.json` correctly, and the spawn code (which reads from `ai-settings.json`) also sees the right value.

**Caveat**: this workaround can be silently undone by Nimbalyst's project-level Settings UI rewriting `ai-settings.json` on save — see related issue [NIM-801](https://linear.app/nimbalyst/issue/NIM-801/project-settings-ui-silently-overwrites-user-level) for the writeback bug.

## References

* Related: [NIM-761](https://linear.app/nimbalyst/issue/NIM-761/claude-code-exits-with-code-1-on-windows-works-in-cli-fails-in) (Claude Code exits with code 1 on Windows — same root cause class, different platform/symptom). The user there reports forcing standalone binary "had no effect", which is exactly this bug
* Related: [NIM-591](https://linear.app/nimbalyst/issue/NIM-591/claude-code-authentication-fails-on-windows-nimbalyst-looks-for-macos) (Claude Code authentication fails on Windows — credential lookup bug, similar architectural issue with multiple stores not sharing state)

raheel-lakhpaty:
## Summary

Move `ClaudeCodeProvider.setCustomClaudeCodePathLoader(...)` from `index.ts` into the `AIService` constructor so it reads from the correct electron-store instance (`ai-settings.json`) where the UI persists the custom path. Previously the loader in `index.ts` called `store.get('customClaudeCodePath')` which reads from `app-settings.json`, a separate store that never receives this value, so the custom binary path was always ignored.

## Related issue

Closes #58

## Testing

Manually tested with a corporate SSO wrapper binary (`/usr/local/bin/<path_to_custom_claude>`) set as the custom Claude Code path. Before this fix the app showed "Login Required" despite the path being configured. After the fix, the `[CLAUDE-CODE] Binary path:` log confirms `custom=<path>` rather than `(none)`, and the wrapper is correctly invoked.

## Notes for reviewers

`ClaudeCodeProvider.setCustomClaudeCodePathLoader` is last-write-wins. The stale registration in `index.ts` would always overwrite the correct one set by `AIService` (since `app.whenReady()` fires after the constructor).

### DIFF
--- a/packages/electron/src/main/index.ts
+++ b/packages/electron/src/main/index.ts
@@ -1720,25 +1720,8 @@ app.whenReady().then(async () => {
     const sessionRestored = shouldSkipSessionRestore ? false : await restoreSessionState();
     markEnd('session-restore');
 
-    // Set up a loader that reads customClaudeCodePath fresh from the store on each query,
-    // so changes in the UI take effect without restarting the app. The workspace path is
-    // required: getAIProviderOverridesWithWorktreeFallback handles direct overrides plus
-    // worktree-to-parent inheritance. Only fall through to the global setting when no
-    // project-level override exists at all (which is the intended fallback, not a routing
-    // failure). A missing workspacePath here would mean an upstream wiring bug, so throw.
-    ClaudeCodeProvider.setCustomClaudeCodePathLoader((workspacePath: string) => {
-      if (!workspacePath) {
-        throw new Error('[ClaudeCodeProvider] customClaudeCodePathLoader called without a workspacePath');
-      }
-
-      const projectOverride = getAIProviderOverridesWithWorktreeFallback(workspacePath)?.customClaudeCodePath;
-      if (projectOverride !== undefined) {
-        return projectOverride;
-      }
-
-      return (store.get('customClaudeCodePath', '') as string) || '';
-    });
-    logger.main.info('[ClaudeCodeProvider] Initialized customClaudeCodePath loader (workspace-aware)');
+    // Note: customClaudeCodePathLoader is now set up in AIService constructor,
+    // where it has direct access to the ai-settings store that owns this value.
 
     // Close splash screen now that initialization is done and a real window is about to show.
     // The last restored window activates the app via its own ready-to-show handler.
--- a/packages/electron/src/main/services/ai/AIService.ts
+++ b/packages/electron/src/main/services/ai/AIService.ts
@@ -55,7 +55,7 @@ import {
   shouldShowCommunityPopup,
   wasCommunityPopupShownThisLaunch
 } from '../../utils/store';
-import { mergeAISettings } from '../../utils/aiSettingsMerge';
+import { mergeAISettings, getAIProviderOverridesWithWorktreeFallback } from '../../utils/aiSettingsMerge';
 import { DocumentContextService, type RawDocumentContext, type PreparedDocumentContext } from '@nimbalyst/runtime';
 import { getMessageSyncHandler, getSyncProvider, isDesktopTrulyAway } from '../SyncManager';
 import { normalizeCodexProviderConfig, omitModelsField, stripTransientProviderFields } from '@nimbalyst/runtime/ai/server/utils/modelConfigUtils';
@@ -164,6 +164,22 @@ export class AIService {
     for (const tool of BUILT_IN_TOOLS) {
       toolRegistry.register(tool);
     }
+
+    // Wire up the custom binary path loader so each query reads the current
+    // value fresh from the ai-settings store. This must live here (not in
+    // index.ts) because only AIService owns the ai-settings store; the
+    // store reference in index.ts points to app-settings and would always
+    // return empty string.
+    ClaudeCodeProvider.setCustomClaudeCodePathLoader((workspacePath: string) => {
+      if (!workspacePath) {
+        throw new Error('[ClaudeCodeProvider] customClaudeCodePathLoader called without a workspacePath');
+      }
+      const projectOverride = getAIProviderOverridesWithWorktreeFallback(workspacePath)?.customClaudeCodePath;
+      if (projectOverride !== undefined) {
+        return projectOverride;
+      }
+      return (this.getSettingsStore().get('customClaudeCodePath', '') as string) || '';
+    });
 
     // API keys must be explicitly set by the user in settings.
     // NEVER auto-import keys from process.env. A user's .env file with


omartelo:
we need this

Issue #163, opened 2026-05-05
Bash tool calls fail with "Stream closed" after ~2 long subprocess invocations per session in bypass-all mode


ademczuk:
## Summary

In `permissionMode: bypass-all`, the Bash tool starts returning `Tool permission request failed: Error: Stream closed` after roughly 2-3 long-running subprocess invocations per session. Trivial Bash builtins keep working. Only subprocess-launching invocations and (sometimes) `Write` of new files fail. Restarting Nimbalyst resets the wall for another ~2 calls.

This is on 0.58.21, which already includes the recent shipped "Stream closed" fixes listed below. The reproducer still triggers, and the proximate cause looks different from those fixes - the symptom correlates with PGLite write-lock contention on the audit-log tables.

## Acknowledged context (already shipped)

The following fixes are in 0.58.21 and have been verified present:

- `Stop hook_0 Stream closed cascade on Claude Code 2.1.116` (NIM-340 #1) - `SDK_NATIVE_TOOLS` whitelist expansion for `Monitor`, `ExitWorktree`, `PushNotification`, etc.
- `Bump claude-agent-sdk to 0.2.117` / native binary `2.1.117` (NIM-340 #2) - addresses the binary-side cascade.
- `ClaudeCodeProvider AsyncIterable 5s grace timer` - so late `can_use_tool` control requests can complete on the still-open stdin.
- `Stop renderer freeze and OOM on long Claude Code streams` (NIM-411) - rAF coalescing + per-session Map.

The new evidence below is additive. Filing as a new issue rather than commenting on a closed one because there's no public tracking issue for this behavior.

## Repro

1. Open a session in any workspace with `permissionMode: bypass-all`.
2. From the Claude Code provider, run a Bash tool call that launches a 200+ second subprocess producing moderate stdout, e.g. a Python script that produces ~5 KB of progress prints.
3. After it completes, repeat with a different command. (Works - call #2.)
4. Repeat again. Call #3 (sometimes #2) returns immediately, sub-second, with `Tool permission request failed: Error: Stream closed`.

Reproduced 3 times across 2 Nimbalyst restarts (admin and non-admin). Each restart resets the wall for ~2 calls before re-triggering. Admin elevation does not raise the limit.

## Expected

Tool calls should keep working until the operator stops them, especially in `bypass-all` mode where there's no permission UI to time out.

## Actual

[...]

PR #166, opened 2026-05-05
fix(tool-errors): annotate Stream closed tool errors with audit-log guidance

ademczuk:
## Summary

Replace the cryptic `Stream closed` tool-result error with an operator-actionable diagnostic message that identifies the audit-log-contention pattern and points at the tracking issue. Original error text preserved verbatim. This is the cheapest of the 6 fixes ranked in #163.

## Related issue

References #163. This PR addresses suggestion #1 from the issue body (\"surface a clearer error than `Stream closed`\"). The deeper fixes (audit-write offloading, IPC timeout tuning, separate PGLite connection pool, etc.) remain open for follow-up.

## What changed

- New helper `annotateStreamClosedToolResult` in `claudeCode/toolChunkUtils.ts`. Pattern matches only when `isError=true` AND the string content includes `Stream closed`. Non-matching inputs pass through unchanged.
- `ClaudeCodeProvider` calls the helper once on the `tool_result` branch, immediately before `applyToolResultToToolCall`. The original `item.content` is left untouched so existing `STREAM_CLOSED_DIAGNOSTIC` console logging keeps inspecting the raw signal.
- New test file `__tests__/toolChunkUtils.test.ts` with 6 tests covering the four guard branches plus integration with `applyToolResultToToolCall` (including a regression check that the Edit tool's success-path result-shape preservation is unaffected).
- CHANGELOG entry under `[Unreleased] -> Fixed`.

## What it surfaces to operators

Before:
```
Tool permission request failed: Error: Stream closed
```

After (when the failure pattern matches):
```
Tool call failed in Nimbalyst.

The native binary returned \"Stream closed\". The most common cause in
permissionMode: bypass-all is PGLite write-lock contention on the
audit-log tables (ai_agent_messages, ai_tool_call_file_edits) under
sustained tool-call traffic. Restarting Nimbalyst typically resets
the wall for ~2 more calls.

See nimbalyst/nimbalyst#163 for the full hypothesis and ranked workarounds.

Original tool_result content:
Tool permission request failed: Error: Stream closed
```

## Testing

- `npx vitest run packages/runtime/src/ai/server/providers/claudeCode/__tests__/toolChunkUtils.test.ts` -> 6/6 pass
- `npm run typecheck --prefix packages/runtime` -> clean
- `git merge-tree --write-tree upstream/main HEAD` -> exit 0 (no conflicts)
- Verified locally: in a long bypass-all session that was reproducing the issue, the new wrapped message renders in the chat with the original error preserved at the end.

## Notes for reviewers

- The helper short-circuits on `isError=false`, so the Edit tool's success path that builds a structured `{ message, file_path, old_string, new_string }` result is unaffected. There is an explicit regression test for this.
- `processTeammateToolResult` and `task.summary` are left reading the original `item.content` rather than the annotated string. Single-concern principle: this PR only changes what reaches the chat UI. Happy to extend to those paths in a follow-up if you'd prefer.
- This does not fix the underlying PGLite write-lock contention. It just stops the error message from sending operators down the wrong debugging path. Suggestions 2-6 in #163 (offload audit writes, raise IPC timeout, separate connection pool, periodic VACUUM, coarser flush granularity) remain open and would each be a separate PR.
- The replacement string mentions `nimbalyst/nimbalyst#163` explicitly so the link survives if a user copies the error out of Nimbalyst into their own bug report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Surface a clearer error than `Stream closed` on Claude Code tool-call failures. The native binary's bare error string left operators chasing permission-mode or token issues, but the failure correlates with PGLite write-lock contention on the audit-log tables (`ai_agent_messages`, `ai_tool_call_file_edits`) under sustained tool-call traffic in `permissionMode: bypass-all`. `ClaudeCodeProvider` now passes tool-result content through a new `annotateStreamClosedToolResult` helper that prepends a diagnostic message identifying the audit-log-contention pattern and pointing at #163 for the full hypothesis and ranked deeper fixes. The original error text is preserved verbatim at the end. Existing `STREAM_CLOSED_DIAGNOSTIC` console logging is untouched (it reads the raw chunk content, not the toolCall result). References #163.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/runtime/src/ai/server/providers/ClaudeCodeProvider.ts
+++ b/packages/runtime/src/ai/server/providers/ClaudeCodeProvider.ts
@@ -63,6 +63,7 @@ import {
   prepareClaudeCodeAttachments,
 } from './claudeCode/messagePreparation';
 import {
+  annotateStreamClosedToolResult,
   applyToolResultToToolCall,
   isSearchableAssistantChunk,
 } from './claudeCode/toolChunkUtils';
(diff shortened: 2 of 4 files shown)


ademczuk:
Hi @ghinkle - you mentioned a few small issues first would be great. This is one of two I'm putting up today (the other is #168). It's the cheapest fix listed in the body of #163 - swap the bare "Stream closed" string for an operator-actionable message, no deeper plumbing changed. Existing `STREAM_CLOSED_DIAGNOSTIC` console logging is untouched. Whenever you have a moment, no rush.

ghinkle:
Yea, thanks. I will have a look. I've been working on the root of the stream closed errors for a couple days. It seems to have gotten much better with db write coalescing I pushed earlier today.
https://github.com/nimbalyst/nimbalyst/commit/d0873eb832241bab1f87533f33f71462eb351c53
But I'll take a look at this too.

ademczuk:
Closing in favor of @ghinkle's root-cause fix landed in v0.59.0 (`d0873eb`).

Verified tonight in isolation: tag `13935f6f`, fresh userData, clean throw-away workspace, `permissionMode: bypass-all`, four sequential ~5-minute Python scripts under the Bash tool. 4/4 invocations completed cleanly, zero "Tool permission request failed: Error: Stream closed" anywhere. Issue #163 reports the wall tripping on call 2-3 in this exact shape, so the coalescing fix holds.

Annotation in this PR is now redundant. Thanks for the fast turnaround.

PR #168, opened 2026-05-05
fix(git-watcher): skip empty repos cleanly instead of logging a stack trace

ademczuk:
## Summary

Open a workspace whose `.git` exists but has zero commits without spamming a multi-line `[GitRefWatcher] Failed to start watching: ...` stack trace into `main.log`.

Hit reliably on freshly-init'd workspaces (the user runs `git init` inside a folder, opens it as a workspace before making any commits). `GitRefWatcher.start` was letting the `simple-git` "fatal: your current branch X does not have any commits yet" exception flow into the outer catch as a full error log.

## What changed

- `GitRefWatcher.ts`: pre-flight branch + HEAD lookup is now wrapped in a narrow try/catch that pattern-matches `/does not have any commits yet/i`. On match, logs a one-line `info` message and returns cleanly. Mirrors the existing detached-HEAD short-circuit.
- All other errors still flow to the outer catch unchanged. The fix is intentionally narrow.
- Tests cover three pre-flight paths: empty repo (info, no error), unrelated git error (still hits outer error log), detached HEAD (existing behavior preserved).

## Testing

- `npx vitest run packages/electron/src/main/file/__tests__/GitRefWatcher.test.ts` -> 3/3 pass
- `npm run typecheck --prefix packages/electron` -> clean
- `git merge-tree --write-tree upstream/main HEAD` -> exit 0 (no conflicts)
- Manual repro: `git init` a fresh folder, open it in Nimbalyst, before vs after the patch:
  - Before: 8-line `[GitRefWatcher] Failed to start watching: Error: fatal: ...` stack trace in main.log
  - After: `[GitRefWatcher] Skipping workspace with no commits yet: <basename>`

## Notes for reviewers

- The pattern-match is on the message text rather than an error type because `simple-git` doesn't expose typed errors for this case. The string `"does not have any commits yet"` is git's own canonical wording (`die_if_no_commits` in git's source) and is locale-stable for English locales. If you'd prefer a different detection mechanism (e.g. swallow all errors from log() and rely on the empty `lastCommitHash` fallback already at the call site), happy to switch.
- Once the user makes their first commit, they currently need to reopen the workspace for the watcher to attach. Adding a periodic re-check or a parent-directory watch to auto-attach on first commit is a follow-up; this PR's scope is just the noisy-log fix.
- The CHANGELOG entry is in `[Unreleased] -> Fixed` directly above the `[0.58.21]` block. If PR #166 (open) lands first, there will be a one-line CHANGELOG conflict that's mechanical to resolve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Open a workspace whose `.git` repo has zero commits yet without spamming a multi-line stack trace into `main.log`. `GitRefWatcher.start` was letting the "fatal: your current branch X does not have any commits yet" error from `simple-git`'s `log()` call escape into the outer catch as `[GitRefWatcher] Failed to start watching: ...`. The pre-flight branch + HEAD lookup is now wrapped in a narrow try/catch that pattern-matches that exact error, logs a one-line info message, and returns cleanly. All other errors still surface to the outer catch unchanged. Mirrors the existing detached-HEAD short-circuit. Adds unit tests for the empty-repo, unrelated-error, and detached-HEAD paths.
 
 ### Removed
 <!-- Removed features go here -->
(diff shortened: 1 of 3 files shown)


ademczuk:
Hi @ghinkle - paired with #166 as the second of my "small issues first" set. This one removes the multi-line stack trace that fires when a workspace's `.git` exists but has zero commits yet (common when someone runs `git init` then opens the folder before making the first commit). Mirrors the existing detached-HEAD short-circuit a few lines above. No rush, take your time.

PR #177, opened 2026-05-06
fix(settings): read SDK version from both candidate node_modules locations

### LINKED ISSUE #60: Settings → Claude Agent SDK panel shows "Version: unknown" (hardcoded literal in macOS renderer build)
## Summary

The Settings → Claude Agent SDK panel always shows `Version: unknown` on macOS builds because `BUNDLED_SDK_VERSION` is hardcoded as the literal string `"unknown"` in the bundled renderer JS. There is no IPC, no `package.json` lookup, and no fallback logic — the value is baked in at build time.

Cosmetic only — does not affect functionality. But misleading: users see "unknown" and assume the SDK isn't bundled, when in fact it is (`@anthropic-ai/claude-agent-sdk@0.2.81`).

## Environment

- **Nimbalyst**: 0.56.17 (macOS build)
- **macOS**: 26.3.2 (Tahoe) — Apple Silicon

## Steps to Reproduce

1. Open Nimbalyst
2. Settings → Claude Agent SDK panel
3. Look at the "Version" field

**Expected**: A real SDK version string, matching the bundled `@anthropic-ai/claude-agent-sdk` package version (e.g., `0.2.81`).

**Actual**: `Version: unknown`.

Note: other Macs running the same Nimbalyst `0.56.17` version have been observed showing real version strings like `0.2.69` in the same panel, suggesting some builds substitute the constant with the real value but this build doesn't.

## Root Cause

In the bundled renderer JS at `out/renderer/assets/index-CyheNb8n.js:264841`:

```js
const BUNDLED_SDK_VERSION = "unknown";
```

Used at line `265215`:

```js
children: BUNDLED_SDK_VERSION
```

This is the only definition site and the only usage site — there is no IPC, no lookup, no useEffect, nothing dynamic. The string `"unknown"` is baked into the renderer bundle at build time.

The Windows branch of the same `ClaudeCodePanel` component uses a real lookup at line `265192`:

```js
isWindowsPlatform ? <Windows panel using claudeCodeWindowsStatus.claudeCodeVersion> : <macOS panel using BUNDLED_SDK_VERSION>
```

So the macOS branch goes straight to the hardcoded literal, while the Windows branch consults runtime state.

There IS a `getClaudeCodeVersion` function in the main bundle at `out/main/index.js:L475057-L475066` that reads `manifest.json` from `require.resolve("@anthropic-ai/claude-agent-sdk")` and falls back to `"unknown"` on throw, but it is **only used as a User-Agent header for the usage API** (called from `L475240`). It is never wired to the Settings panel.

The bundled SDK package metadata is real and accessible:

```bash
$ cat /Applications/Nimbalyst.app/Contents/Resources/app.asar.unpacked/node_modules/@anthropic-ai/claude-agent-sdk/package.json
# version: 0.2.81
# claudeCodeVersion: 2.1.81
```

So the data is there, the panel just doesn't read it.

## Impact

Cosmetic only. Does not block any functionality. The SDK and CLI work normally; only the displayed version string is wrong.

## Suggested Fix

Two clean options:

**Option 1: Build-time substitution (preferred)**

Use a Vite `define` substitution to inject the real SDK version at build time:

```js
// vite.config.ts (or rollup.config.js)
import sdkPkg from '@anthropic-ai/claude-agent-sdk/package.json';

export default {
  define: {
    __BUNDLED_SDK_VERSION__: JSON.stringify(sdkPkg.version),
  },
}
```

Then in the renderer source:
```js
const BUNDLED_SDK_VERSION = __BUNDLED_SDK_VERSION__;
```

This adds zero runtime cost and the value is locked to whatever was in `package.json` at build time, which is exactly what "bundled with app" means.

**Option 2: Runtime IPC fetch**

Add an IPC handler `claude-agent-sdk:getVersion` in the main process that reads `package.json` from the bundled SDK location, and have the React component call it on mount via `useEffect`. This matches what the Windows branch does for `claudeCodeWindowsStatus`.

Option 1 is simpler and matches the "Built-in (bundled with app)" framing of the panel.

## Workaround

None — and none needed since this is purely cosmetic. The actual bundled SDK is `@anthropic-ai/claude-agent-sdk@0.2.81` regardless of what the panel says.

## References

No existing issue found via search.

ademczuk:
## Summary

This PR hopefully fixes the `Settings > Claude Agent SDK` panel reading `Version: unknown` on builds where npm workspace dedup hoists `@anthropic-ai/claude-agent-sdk` to the repo-root `node_modules/` instead of leaving it under `packages/electron/node_modules/`.

The build-time read in `electron.vite.config.ts` was hardcoded to `resolve(__dirname, 'node_modules/...')`, with a silent `catch { return 'unknown' }` that swallowed the ENOENT on hoisted installs. The literal string `"unknown"` then got baked into the renderer bundle (`__CLAUDE_AGENT_SDK_VERSION__` define), and the macOS panel had no runtime fallback.

## Related issue

Closes #60.

## What changed

`packages/electron/electron.vite.config.ts`: the IIFE that computes `claudeAgentSdkVersion` now tries the local path first (preserves prior behavior on non-hoisted installs), then the workspace-root candidate. Same `fs.readFileSync` + `JSON.parse` logic; only difference is trying a second location before giving up.

## Testing

- `npm run typecheck --prefix packages/electron` clean.
- Verified locally: on this checkout the SDK is at `packages/electron/node_modules/@anthropic-ai/claude-agent-sdk` and resolves to `0.2.126`; the second candidate is correctly skipped (path doesn't exist). On a hoisted install, the second candidate would resolve and the Settings panel would show the real version instead of `'unknown'`.

## Notes for reviewers

- **Cosmetic fix only** - no behavior change beyond the displayed string. Reporter at #60 confirmed the literal `"unknown"` is the only definition + only usage in the renderer bundle, no IPC.
- **Why try-paths instead of `require.resolve`?** Vite config files have inconsistent ESM/CJS resolution semantics across electron-vite versions and `createRequire(import.meta.url)` adds a small but non-zero risk of breaking on a config that compiles to a target without `import.meta`. The two-candidate try is sufficient because npm workspace hoisting only ever lands the package in those two places.
- **Alternative considered**: replace the build-time define with an IPC + runtime lookup (the existing `getClaudeCodeVersion` in `ClaudeUsageService.ts:69` already does this for User-Agent purposes). Bigger surface than this issue warrants - the literal-define approach is correct in principle, just had a path-resolution gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Settings → Claude Agent SDK panel no longer reads `Version: unknown` on builds where npm workspace dedup hoists `@anthropic-ai/claude-agent-sdk` to the repo-root `node_modules/`. The build-time version read in `electron.vite.config.ts` was hardcoded to `packages/electron/node_modules/...` and silently fell through to `'unknown'` on hoisted installs; it now tries both the local and workspace-root candidates. Closes #60.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/electron.vite.config.ts
+++ b/packages/electron/electron.vite.config.ts
@@ -87,14 +87,28 @@ const isOfficialBuild = process.env.OFFICIAL_BUILD === 'true';
 // IS_DEV_MODE is true only when running `npm run dev`, not for any packaged builds
 const isDevMode = isDev;
 
-// Read Claude Agent SDK version at build time for display in settings
+// Read Claude Agent SDK version at build time for display in settings.
+//
+// The bundled SDK may live in packages/electron/node_modules/ (when not
+// hoisted) OR in the workspace-root node_modules/ (when npm workspace dedup
+// hoists it). Hardcoding the local path to the package was silently falling
+// through to 'unknown' on builds where the install hoisted, so the Settings
+// panel always read 'Version: unknown' on those builds. Try the local path
+// first, then the workspace-root fallback. Closes #60.
 const claudeAgentSdkVersion = (() => {
-  try {
-    const pkgPath = resolve(__dirname, 'node_modules/@anthropic-ai/claude-agent-sdk/package.json');
-    return JSON.parse(fs.readFileSync(pkgPath, 'utf-8')).version;
-  } catch {
-    return 'unknown';
+  const candidates = [
+    resolve(__dirname, 'node_modules/@anthropic-ai/claude-agent-sdk/package.json'),
+    resolve(__dirname, '../../node_modules/@anthropic-ai/claude-agent-sdk/package.json'),
+  ];
+  for (const pkgPath of candidates) {
+    try {
+      const version = JSON.parse(fs.readFileSync(pkgPath, 'utf-8')).version;
+      if (version) return version;
+    } catch {
+      // try next candidate
+    }
   }
+  return 'unknown';
 })();
 const runtimeSrcDir = resolve(__dirname, '../runtime/src');
 const runtimeDistDir = resolve(__dirname, '../runtime/dist');


ademczuk:
Hi @ghinkle - this one's ready for review when you have a moment. Cosmetic fix to the SDK version display in the Settings panel, single config file change.

PR #180, opened 2026-05-06
fix(walkthrough): use 420px width in clamp math when step.wide is true

ademczuk:
## Summary

Wide-variant walkthrough steps render at 420px (`w-[420px]` in `WalkthroughCallout.tsx`) but `calculateCalloutPosition` was always clamping with `CALLOUT_WIDTH = 320`. On wide steps near the right viewport edge, that mismatch let the callout overflow the edge or get clipped, depending on target position.

## Fix

- Add `CALLOUT_WIDE_WIDTH = 420` with a doc-comment coupling it to the Tailwind class on the rendered component
- Plumb `wide: boolean = false` through `calculateCalloutPosition()` and bind a local `calloutWidth`
- Pass `step.wide ?? false` from `WalkthroughCallout.tsx`
- Use `calloutWidth` consistently in the `fitsRight`/`fitsLeft` checks, all four placement branches (top, bottom, left, default), the viewport `left` clamp, and the horizontal arrow-offset clamp

Default of `false` keeps every existing caller behaviorally identical.

## Tests

New `packages/electron/src/renderer/walkthroughs/__tests__/WalkthroughService.test.ts` with 6 cases:

- non-wide: right-edge target gets clamped inside the viewport
- non-wide: right placement honored when there is room
- wide: `fitsRight` uses 420 (falls back to bottom when 320 fits but 420 does not)
- wide: viewport-left clamp uses 420 (compares wide vs non-wide on identical target)
- wide: arrow offset stays within 420-callout bounds
- omitting `wide` is identical to passing `false`

The test file uses `// @vitest-environment jsdom` because `vitest.config.ts` declares `environment: 'node'` and the tests need `document`/`window`. `jsdom` is already hoisted in the workspace root.

```
$ npx vitest run src/renderer/walkthroughs/__tests__/WalkthroughService.test.ts
Test Files  1 passed (1)
     Tests  6 passed (6)
```

## Test plan

- [x] vitest passes locally (6/6)
- [ ] CI green
- [ ] Visual check on a wide walkthrough step near the right viewport edge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Wide walkthrough callouts no longer overflow the right viewport edge or sit clipped on tight layouts. `calculateCalloutPosition` was clamping with the 320px default callout width regardless of `step.wide`, but `WalkthroughCallout.tsx` renders wide steps at 420px (`w-[420px]` Tailwind class). The function now takes a `wide` parameter and binds a local `calloutWidth` (320 or 420) used consistently in fit checks, placement math, viewport clamp, and arrow-offset clamp. Adds `WalkthroughService.test.ts` with unit coverage for both widths.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/renderer/walkthroughs/WalkthroughService.ts
+++ b/packages/electron/src/renderer/walkthroughs/WalkthroughService.ts
@@ -218,6 +218,13 @@ export interface CalloutPosition {
 }
 
 const CALLOUT_WIDTH = 320;
+/**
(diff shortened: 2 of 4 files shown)


ademczuk:
@ghinkle when you have a moment, could you take a look? Tests pass locally and the change is a pure default-preserving plumb of `step.wide` through to the existing clamp math, no behavior change for non-wide callers.

ademczuk:
Closing this as redundant - my mistake.

Looking at it more carefully:
- The actual fix (CALLOUT_WIDE_WIDTH constant + `wide` parameter plumb-through) shipped in #164 and is already on main as commit 6fa01617, with a CHANGELOG entry under v0.59.0.
- The only net-new content in this PR was a `// @vitest-environment jsdom` directive on the test file. I added it because vitest failed locally when I ran it from inside `packages/electron/` (which uses the package-level `vitest.config.ts` with `environment: 'node'`).
- But CI uses the top-level `vitest.config.ts` which has `environmentMatchGlobs` routing renderer tests to jsdom, so CI was already passing for this file on main without the directive.

Net effect: the directive isn't needed and the rest is duplicate. Closing.

PR #181, opened 2026-05-06
fix(workstream-tabs): wait for hydration before restoring tab list (fixes #169)

### LINKED ISSUE #169: All open file tabs disappear when switching between tasks/sessions/files
## Summary
All open file tabs occasionally disappear while switching between tasks, sessions, and files, leaving the editor with no tabs at all.

## Steps to reproduce
1. Have multiple file tabs open in a workspace.
2. Switch around between tasks, sessions, and files.
3. At some point, all tabs vanish.

Not consistently reproducible — happens occasionally.

## Expected behavior
Open file tabs persist while navigating between tasks, sessions, and files.

## Actual behavior
The entire tab strip becomes empty. No tabs are present. Reloading or restarting does not bring them back; the user has to manually reopen files.

## Environment
- App version: 0.58.21
- OS: darwin 25.2.0 (arm64)
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
Not collected — main process logs around the incident showed no errors, warnings, or tab/workspace-state events.

ademczuk:
## Summary

Open file tabs occasionally disappear while switching between tasks, sessions, or files, leaving the workstream editor with no tabs. Reload does not recover them, because the saved tab list has already been overwritten with `[]` by the time the user notices.

Fixes #169.

## Root cause

`WorkstreamEditorTabs` runs its restore effect on mount and reads `workstreamState.openFilePaths`. If the component mounts before `loadWorkstreamStateAtom` has finished hydrating from IPC, that read returns the atom default `openFilePaths: []`. The effect restores zero tabs, marks `restoreStateRef = 'done'`, and the persist effect immediately fires and writes `openTabs: []` back to workspace state, replacing the saved tab list permanently.

`workstreamStatesLoadedAtom` already exists for exactly this race - `AgentWorkstreamPanel` uses it to gate child-session loading - but `WorkstreamEditorTabs` was not consulting it.

## Fix

Add a single guard at the top of the restore effect: if `workstreamStatesLoadedAtom` is still `false`, return early without touching `restoreStateRef`. Once hydration completes, the dep array bumps, the effect re-runs, and restore proceeds normally.

The existing `restoreStateRef === 'done'` gate on the persist effect already blocks IPC writes during the pre-hydration window, so no second guard is needed there.

## Files changed

- `packages/electron/src/renderer/components/AgentMode/WorkstreamEditorTabs.tsx` - import `workstreamStatesLoadedAtom`, gate restore effect on it, add comment + add atom to dep array
- `CHANGELOG.md` - entry under [Unreleased] > Fixed

13 lines of logic in one component, 1 changelog line.

## Testing rationale

No unit test added. The renderer test infrastructure in this package is unit-test-of-pure-functions only - no React Testing Library or component renders set up. Adding component-mount tests would be a much larger lift than the fix itself, with significant Jotai/TabsProvider/IPC mock setup. Happy to add a render test if you'd prefer; just point me at any existing component test that mounts a Jotai-backed renderer component and I will follow that pattern.

## Test plan

- [x] `npx tsc --noEmit -p packages/electron/tsconfig.json` clean
- [ ] CI green
- [ ] Manual: open ~5 file tabs, switch between sessions/workstreams, reload - tabs persist
- [ ] Manual: with empty workstream state file, mount the component - no spurious `openTabs: []` write to workspace state during the load window


ademczuk:
@ghinkle small one for the queue when convenient. Pure additive guard, mirrors the hydration pattern `AgentWorkstreamPanel` already uses for child-session loading.

ademczuk:
Follow-up commit `d9c893ed`: defense-in-depth latch against empty-tab clobber.

The original commit gates the restore effect on `workstreamStatesLoadedAtom`, which fixes the timing race in #169. A deeper pass found two remaining paths to the same data-loss outcome:

1. `loadWorkstreamStates` IPC throws -> the catch at `workstreamState.ts:775` flips the loaded flag to true with an empty workstream map. The gate now passes, restore reads `openFilePaths: []`, marks itself done, persist writes `[]` over saved state.
2. On-demand `loadWorkstreamState(id)` race. `AgentWorkstreamPanel:561` fires it async on workstream switch. `WorkstreamEditorTabs` mounts in parallel and the gate is already open from a prior bulk load, so the restore effect reads the atom before the per-workstream merge completes.

Added a `hasEverHadTabsRef` latch: only allow the persist effect to write `[]` to disk after we've observed at least one non-empty tab state. New workstreams legitimately start empty, so the deferred write is harmless - the first user-opened file flips the latch and persistence proceeds normally. Closing all tabs after that is still allowed (legitimate user action).

Net: even if the upstream loaded-flag invariant is loosened by future error-handling changes, this component cannot wipe saved tabs as long as the user actually has saved tabs.

PR #182, opened 2026-05-06
fix(git-status): walk up to find owning git root for nested repos (fixes #122)

### LINKED ISSUE #122: Editor tab git indicator shows uncommitted state for files in nested repos within a non-git workspace root
## Summary
The editor tab "uncommitted" indicator stays on for files that have already been committed, when the file lives in a nested git repository inside a workspace root that is not itself a git repo.

## Steps to reproduce
1. Open a workspace whose root directory is **not** a git repo, but which contains one or more nested git repos in subdirectories (a superproject-style layout).
2. Open a file from one of the nested repos in the editor.
3. Edit and save the file — the tab correctly shows the uncommitted indicator.
4. Commit the file inside its nested repo (`git -C <nested-repo> commit ...`) and confirm with `git -C <nested-repo> status` that the working tree is clean.

## Expected behavior
The tab indicator clears once the file is committed in its owning nested repo.

## Actual behavior
The tab continues to show the file as uncommitted/modified. Other Nimbalyst tooling that reasons about edited files (e.g. `get_session_edited_files`) is similarly unaware of the nested-repo state — the project's own CLAUDE.md documents this as a known limitation:

> The Nimbalyst Session edits panel and `get_session_edited_files` MCP tool may not track edits here because they look for a single `.git` at the workspace root.

The git integration appears to look only at the workspace root for `.git` and does not walk into subdirectories to find nested repositories.

## Environment
- App version: 0.58.14
- OS: Linux 6.8.0-40-generic (Ubuntu)
- Workspace layout: non-git root with multiple sibling nested git repos

## Logs
Not collected.

ademczuk:
## Summary

Editor tab git indicator stays on indefinitely for files in nested git repos when the workspace root is not itself a git repo. `GitStatusService.getFileStatus` bailed out whenever `workspacePath/.git` did not exist, so the per-file query that drives the tab indicator never consulted git at all and every file came back as `untracked`.

Fixes #122.

## Root cause

`GitStatusService.getFileStatus` had a single `isGitRepository(workspacePath)` guard that only checked for `.git` at the workspace root. A reproducer:

```
super/                 <- workspace root, no .git
  project-a/.git/      <- nested repo
  project-a/src/app.ts
```

Opening `super/project-a/src/app.ts`, editing, saving, and committing inside `project-a` left the tab indicator stuck on. The file was cleanly committed in its owning repo, but Nimbalyst never asked.

## Fix

Add a `findGitRootForFile(filePath, boundary)` helper that walks up from the file's parent directory looking for a `.git` entry, stopping when:

- a `.git` directory or worktree-link file is found (returns that directory)
- we reach the workspace boundary without finding one (returns `null`)
- the file lives outside the workspace (returns `null`, so we never resolve against an unrelated repo elsewhere on disk)

The walk uses proper path-segment boundaries so prefix-colliding siblings (`foo` vs `foo2`) cannot be matched.

`getFileStatus` now:

1. Groups requested files by their owning git root.
2. Runs `git status --porcelain` once per root, with `cwd` = that root.
3. Converts each file's path to root-relative form before looking it up in the parsed status map.
4. Files outside any git repo within the workspace continue to surface as `untracked` (preserves prior non-git-workspace UX).
5. Cache key now includes the sorted list of owning roots so calls hitting different roots are cached separately.

## Out of scope (intentional)

To keep the diff focused, this PR fixes only `getFileStatus` (the per-file query that drives the tab indicator, which is the user-reported symptom). Other entry points are not changed:

- `getUncommittedFiles` and `getAllFileStatuses` are workspace-wide queries; nested-repo support there needs a directory-level repo discovery pass and is a separate, larger change.
- `GitRefWatcher` still only watches the workspace root's `.git/HEAD` and `.git/index`. The 5 second cache TTL on `getFileStatus` means external commits in nested repos still get picked up within ~5 s on the next status query, which is enough for the tab indicator to clear naturally without a watcher overhaul.

If maintainers want me to land those follow-ups in subsequent PRs, happy to.

## Tests

New `packages/electron/src/main/services/__tests__/GitStatusService.findGitRoot.test.ts` with 7 cases over real on-disk tmp dirs:

- workspace IS a git repo - returns workspace
- non-git workspace with a nested repo - returns the nested root
- nested repo inside a git workspace (submodule-like) - returns the deepest repo
- workspace with no git anywhere - returns null
- file outside the workspace boundary - returns null (must not match a repo higher up the filesystem)
- relative file paths resolve against the workspace
- prefix-colliding sibling (`foo2` vs workspace `foo`) does not match

```
$ npx vitest run src/main/services/__tests__/GitStatusService.findGitRoot.test.ts
Test Files  1 passed (1)
     Tests  7 passed (7)
```

## Test plan

- [x] vitest passes locally (7/7)
- [x] `tsc --noEmit` clean
- [ ] CI green
- [ ] Manual: superproject layout from the issue - edit + commit a file in a nested repo, verify tab indicator clears within a few seconds
- [ ] Manual: regular git workspace - confirm no behavior regression for the common single-repo case


ademczuk:
@ghinkle this is the meatiest of the three I just opened (~150 lines + tests). I scoped it to just `getFileStatus` since that's what drives the user-visible tab indicator. `getUncommittedFiles` and `getAllFileStatuses` would need a directory-walk pass for nested-repo discovery, which I can do as a follow-up if useful. Let me know if you'd rather I roll them into this PR or split.

ademczuk:
Two follow-up notes from a deeper review pass:

**clearCache predicate fix** (`8596bd10`): the cache invalidation in `clearCache(workspacePath)` was using `startsWith(workspacePath + ':')` which only matched colon-separator keys (`:uncommitted` and the legacy per-file format). The new nested-repo grouping uses null-byte separators (`workspacePath\0...`), so cache entries from this PR were silently surviving `clearCache` calls and serving stale data for up to the 5 second TTL after a git-ref-watcher invalidation. Fixed - predicate now matches either separator. The `\0all-statuses` key used by `getAllFileStatuses` (predates this PR) also benefits.

**Symlinked workspace path** (known limitation, not patched): `findGitRootForFile` uses `path.resolve()` which normalises `.`/`..` and separators but does not follow symlinks. On macOS where Xcode and Node sometimes expose `/private/var/folders/...` vs `/var/folders/...` style paths, a symlinked workspace boundary won't match the file's real path, and files there silently fall through to "all untracked" - same as the pre-fix behaviour, so it is not a regression. Calling `fs.realpathSync` on both at the top of the helper would resolve it; happy to follow up if useful but it adds an I/O call to every classification, so I left it scoped here.

PR #188, opened 2026-05-06
feat: multi-project rail with keep-warm switching

### LINKED ISSUE #155: [Feature]: Multi-project support: switch between projects without juggling windows
## Problem
I work across multiple projects in Nimbalyst, but each project requires its own window. Switching between projects means alt-tabbing between Nimbalyst windows, which is much heavier than the in-app switching I already get between sessions inside a single project. Managing N windows for N projects gets frustrating quickly.

## Proposed solution
Add a project layer above sessions in the sidebar, so the hierarchy becomes **Project → Sessions**. Switching project would be a single click in the sidebar — same lightweight feel as switching sessions today — without spawning a new window.

## Alternatives considered
- Running one Nimbalyst window per project and alt-tabbing between them (current workaround, frustrating).
- Agent Deck handles this differently: it doesn't tie sessions to projects at all — sessions can come from any repo (or multiple repos), and the user organizes them into self-defined Groups in a single sidebar. That's another shape of the same idea: get out of the "one window per workspace" model.

## Additional context
The session-switching UX inside a single project already feels great; this request is about extending that fluidity one level up to projects.

omartelo:
Closes #155

## Summary

Adds a Discord-style vertical project rail to the left of the workspace, allowing multiple projects to stay warm in a single window. Switching between projects is instant: per-workspace services (DocumentService, FileSystemService, MCP config watcher) and renderer state (tabs, agent layout, navigation history, AI sessions) survive the switch. Window-per-project remains as a fallback via right-click "Open in new window".

The feature is **opt-in** via Settings → Advanced → "Multi-project mode" (default: off, preserves current single-window UX).

## Screenshots

**Rail with multiple projects (active vs inactive)**
image

**Add-project menu (`+` button) with recent projects**
image

**Rail item context menu (right-click)**
image

**Settings → Advanced (Multi-project mode + Restore previous projects)**
image

## What's included

**Rail UI**
- New `ProjectRail` component (56px wide, fixed-left). Active project shows squared corners + accent bar; inactive icons round on hover. Activate / close are sibling `<button>` elements inside a non-interactive wrapper — no nested buttons.
- `+` button opens a floating menu with "Open folder…" and a filtered list of recent projects (excludes already-open ones).
- Right-click on a rail item: Close / Open in new window / Reveal in Finder/Explorer.
- Keyboard: `Cmd/Ctrl+1..9` activates the Nth project.

**Settings**
- `multiProjectMode` (default `false`). When off, rail is hidden and opening a project spawns a new window.
- `restorePreviousProjectsOnLaunch` (default `false`). When off, only the actively chosen project is opened on launch — previous warm projects are not auto-restored.

**Renderer state lift**
- Lifted module singletons and per-component `useState` into atom families keyed by `workspacePath`: `agentModeLayoutAtomFamily`, `navigationHistoryAtomFamily`, `tabsStoreAtomFamily`, `windowModeAtomFamily`, `sidebarWidthAtomFamily`, `aiChatWidthAtomFamily`, `aiChatCollapsedAtomFamily`.
- `gitOperationsAtoms` (status / commits / staged / message / committing) backed by atomFamily so each project has isolated git UI state.
- `collabDocuments` providers stored in `Map<workspacePath, TeamSyncProvider>` instead of singleton.
- `openProjectsAtom` + `activeWorkspacePathAtom` are the single source of truth for the visible project. A subscriber on `activeWorkspacePathAtom` dispatches `workspace:set-active` to main on every change (rail click, close fallback, restore-on-launch) — direct IPC calls in `ProjectRail.tsx` are gone.

**Main process**
- `WindowState` extended with `activeWorkspacePath` and `additionalWorkspacePaths`.
- `WindowManager.referencesWorkspace()` extracted; cleanup now treats both primary and additional paths as references, so services are only torn down when no window references the path.
- New IPC `workspace:register-additional` / `workspace:unregister-additional` create and free per-path services (`documentServices`, `fileSystemServices`, `WorkspaceEventBus`, `nim-asset` roots, navigation history, MCP-config watcher).
- `workspace:set-active` handler is the single transition point for active-only resources: stops the previous workspace watcher, starts a fresh one for the new path, and flips the runtime-global `FileSystemService`. Idempotent when the requested path is already active.
- MCP HTTP resolver routes by `activeWorkspacePath` to prevent tool-call leakage between projects in the same window.

**Event routing**
- `ai-session-state:subscribe` accepts `string | string[]` so the renderer subscribes to all open project paths and receives lifecycle events for inactive projects (streaming continues, transcript updates apply).
- `ai:message-logged` (per-row) and `ai:messages-logged-batch` stamp `workspacePath` so the renderer routes events to the owning workspace without consulting the (active-only) `sessionRegistryAtom`.
- `MessageStreamingHandler` routes batches via `findWindowByWorkspace` (rail-aware) instead of fanning out to every `BrowserWindow`.
- `PanelContainer` keys panels by `workspacePath` to force remount on switch (prevents extension panels — e.g. git — from showing stale cross-project data).

**Cross-workspace activity tracking** (`globalSessionActivityAtom`)
- Maintained by `sessionStateListeners` from `session:started/streaming/waiting/completed/error/interrupted` and `ai:message-logged` events.
- Drives rail badges (`projectActivitySummaryAtom`) and the close-with-streaming confirmation in `ProjectRail.handleClose`.
- Replaces the previous derivation that iterated `sessionRegistryAtom` and silently missed inactive workspaces.

**Persistence**
- `app.openProjects` (paths in rail order) and `app.activeProjectPath` persisted via electron-store.
- `app.multiProjectMode` and `app.restorePreviousProjectsOnLaunch` persisted via electron-store.

## Architecture notes

- **Keep-warm strategy**: per-workspace services live in `documentServices` / `fileSystemServices` maps in main and per-workspace atom families in the renderer. Both survive the visible-project flip. The rail does NOT use `display:none` or `key={path}` for its subtree — the previous body claimed it did, but the implementation hooks the rail's active path into the legacy `workspacePath` useState in `App.tsx` and lets the renderer re-paint with the new path. Components that depend only on workspace-keyed atoms keep their state; components that take `workspacePath` as a prop and re-mount (e.g. `CollabMode`, `EditorMode`) read back from the atom families on remount.
- **AI sessions**: `sessionStoreAtom(sessionId)` and live transcript projection are workspace-agnostic, so streaming sessions in inactive projects keep producing rows on disk and surface to the UI when the user switches back.
- **Workspace-state pruning**: a `workspaceStatePruner` subscriber watches `openProjectsAtom`. When a project is removed (rail X / `Cmd+Shift+W`), it drops every per-workspace cache: tabs slot, sidebar/AI-chat layout, agent-mode layout, navigation history, git operations, file-mention search, collab documents (provider + atom-family entries). Without this the renderer accumulated state across the app's lifetime.
- **Refcount unification**: a single `referencesWorkspace()` helper is used by close-window cleanup and by `unregister-additional` to prevent service double-destroy or leak.

## Subsequent fixes (Copilot review)

After the initial commit, the GitHub Copilot reviewer flagged nine items. All resolved in this PR:

| Severity | Issue | Fix |
| --- | --- | --- |
| Critical | `register-additional` started a single-active-per-window watcher that killed the previous watcher | Watcher transitions only happen in `set-active`; `register-additional` just creates services |
| Critical | `unregister-additional` called the per-window `stopWorkspaceWatcher` | Watcher is stopped only when the path being unregistered was the active one |
| Critical | `setFileSystemService` globally flipped on every `register-additional` | FS global flips on `set-active`; per-path services live in `fileSystemServices` map |
| UX | Rail badges only saw the active project's sessions | New `globalSessionActivityAtom` maintained by listeners across all warm projects |
| UX | Close-with-streaming confirm missed sessions in inactive projects | Reads from `globalSessionActivityAtom` |
| A11y | `<button>` nested inside `<button>` on rail icon | Wrapper is a `<div>`; activate and close are sibling buttons |
| Reliability | `providersByPath.set` happened before `await connect()` | Provider cached only on success; failure destroys it |
| Memory | `persistentSlots` and per-workspace atom families never pruned | `workspaceStatePruner` subscribes to `openProjectsAtom` and drops the slot on close |
| Keep-warm | `CollabMode` `useEffect` cleanup destroyed team-sync provider on every switch | Cleanup removed; provider tear-down moved to the rail close flow |

All nine review threads marked resolved on the PR.

## Tests

- **Unit (vitest, +76 new cases, 766 total PASS)**: `openProjects` reducers (add/dedup/cap, close fall-through, derived atoms); `workspaceLayout` per-path family isolation; `windowState` helpers (`windowReferencesWorkspace`, `anyWindowReferencesWorkspace` with `excludeWindowId`); `MultiProjectRailHandlers` IPC behavior with regression guards for the watcher + FS-global fixes; `OptimizedWorkspaceWatcher` lifecycle; `sessionActivity` reducers; `workspaceStatePruner` atom-family removal.
- **E2E (Playwright)**: `e2e/multi-project/multi-project-rail.spec.ts` covers rail rendering, `register-additional` adding a second project, switching, closing the active project, and the cap-at-8 guard. Requires the dev server (`npm run dev`).

## Test plan

- [x] Toggle "Multi-project mode" off → rail hidden, opening a project from menu/CLI/recent spawns a new window (current behavior preserved).
- [x] Toggle on → rail visible, `+` opens a folder picker and recent-projects dropdown; opening a project adds it to the rail.
- [x] Switch between 2+ projects: tabs, sidebar width, agent/files/plan mode, and navigation history are remembered per project.
- [x] Start a long AI session in project A, switch to project B; messages keep streaming and transcript updates land in A's session in the background.
- [x] Modify a file in project A from the terminal while project B is active; on switching back to A the file tree reflects the change.
- [x] Close a project with an active streaming session → confirm prompt appears (and shows the correct count even when closing an inactive project).
- [x] Right-click rail item → "Open in new window" still spawns a separate `BrowserWindow`.
- [x] Quit + relaunch with `restorePreviousProjectsOnLaunch=false` → only the chosen project opens.
- [x] Quit + relaunch with `restorePreviousProjectsOnLaunch=true` → rail rehydrates in saved order with previous active project.
- [x] `Cmd/Ctrl+1..9` activates the Nth rail position.
- [x] Existing E2E suites (`e2e/monaco/`, `e2e/workspace/`, `e2e/agent/`) pass unchanged with rail off.


karlwirth:
This is a great addition. Thank you!! 

Just two  small bugs and small enhancement

Bug:  In light mode, the text is very hard to read against the background.  
Open in new window

Bug:  The original project name is at the top and the project you have selected is at the next level. This is confusing
image


Enhancement: It would be nice to have the full name of the project appear when I hover.  I have lots of projects and just the two letters isn't enough to keep them straight in my mind so I am finding myself clicking into it to figure out which project it is.

karlwirth:
Interactive prompt indicators broken in multi-project rail (PR #188)

**Summary**
Since enabling the multi-project rail (PR #188), I'm seeing some sessions that are waiting on user input (AskUserQuestion, ExitPlanMode, ToolPermission, GitCommitProposal) are not visually surfacing the "waiting for response" state. The session list shows the spinning "thinking" indicator instead of the warning-coloured "contact_support" prompt indicator, and the in-transcript prompt UI sometimes does not render at all.

**Symptoms**
- Session list item shows the processing spinner ("Thinking…") even though the agent has actually emitted a question/permission/plan-confirm/commit-proposal and is blocked waiting on the user.
- The prompt itself sometimes does not appear inline in the transcript — the user has no way to know a response is needed without scrolling/inspecting.
- Reproduces against sessions running in any project that is open in the rail, including (intermittently) the active project.

**Expected**
- The session list shows the warning-coloured "Waiting for your response" indicator (contact_support icon) as soon as the agent requests interactive input.
- The corresponding prompt widget renders in the transcript.

omartelo:
Thanks for the careful pass — addressed all three in commit 3c8a44dc on this branch.

**Light-mode legibility:** root cause was `ProjectRail.css` referencing non-canonical CSS variables (`--nim-fg`, `--nim-fg-muted`, `--nim-fg-on-accent`, `--nim-accent`, `--nim-danger`) that the theme system never defines. Their dark hardcoded fallbacks happened to look fine in dark mode but collapsed onto light backgrounds in light mode. Switched to the canonical names from `UI_PATTERNS.md` (`--nim-text`, `--nim-text-muted`, `--nim-primary`, `--nim-error`) with light-friendly fallbacks.

**Original project name in title bar:** the title-update effect was short-circuiting in agent mode based on a stale comment claiming `AgenticPanel` set the title itself — nothing does that today, so the title bar froze on the workspace that was active when agent mode was entered. Removed the early return so the title now follows the rail-active workspace regardless of mode.

**Hover tooltip:** the tooltip already showed the project name; extended it to also show the absolute path on a second line so two-letter collisions are easy to disambiguate without clicking in.

omartelo:
Diagnosed and fixed in commit 523a625a.

**Root cause:** the renderer's `sessionStateListeners` was clearing `sessionHasPendingInteractivePromptAtom` on every `session:streaming` event — based on the comment "AI resumed streaming - no longer waiting for user input." That's the wrong signal. `session:streaming` fires for *any* token chunk emitted by the provider, including tail-end chunks that arrive after a `tool_use` for AskUserQuestion / ExitPlanMode / ToolPermission / GitCommitProposal.

In single-project mode pre-rail this didn't matter, because the main-process subscription was scoped to the active workspace, so streaming events for warm-but-inactive sessions never reached this window. With the multi-project rail, `getCurrentSubscriptionPaths()` now subscribes to every open rail project; those tail chunks now reach the renderer and clear the pending flag mid-prompt. Inactive-project transcripts are not mounted, so `refreshPendingPromptsAtom` (which can re-derive the flag from messages) never runs — the indicator stays stuck on "Thinking…" until the user opens the session.

**Fix:**

- Removed the clear from the `session:streaming` case. The pending flag is now only mutated by the explicit resolve events (`ai:askUserQuestionAnswered`, `ai:exitPlanModeResolved`, `ai:toolPermissionResolved`, `ai:gitCommitProposalResolved`, `ai:sessionCancelled`) and the terminal lifecycle events (`session:completed/error/interrupted`), which all clear it correctly.

- Defensive: stamped `workspacePath` on the prompt event payloads that did not carry it (`ai:exitPlanModeConfirm`, `ai:askUserQuestion`, `ai:askUserQuestionAnswered`, `ai:toolPermissionResolved`, `ai:gitCommitProposal`, `ai:gitCommitProposalResolved`). Aligns with the pattern already used by `ai:toolPermission` and `ai:message-logged` (#188 commit `795e4296`). Opens a path for future multi-window routing without changing renderer behavior today.

- Added `sessionStateListeners.test.ts` with 15 cases: regression guards against `session:streaming` clearing the pending flag, verification that resolve/terminal lifecycle events still clear it, and set/clear coverage for all four prompt types.

For the second symptom — "the prompt itself sometimes does not appear inline in the transcript" — the durable widgets render from `sessionMessagesAtom` projected from `ai_agent_messages` via `TranscriptTransformer`. Inline rendering should now work consistently once the indicator drives users to the right session. If you still see missing inline prompts after pulling this, please reopen with reproduction steps and I'll instrument the message-projection path.

Tests: `npx vitest run` 781 ✅, `npx tsc --noEmit` 0 errors.

omartelo:
**Bug found while testing: stale `activeSessionIdAtom` leaks across rail workspace switch**

Repro:
1. Open project A in the rail, run a prompt to completion in some session.
2. Click `+` in the rail and add a brand-new project B (never opened in Nimbalyst before).
3. Rail switches to B but the right-side AGENT panel keeps rendering A's tab title and transcript.
4. Type a prompt in what looks like B's empty session. The message lands in A's session and the main process throws `Error: Session <uuid> not found` from `ai:sendMessage` (cross-workspace lookup rejected by `SessionManager`).

Root cause:

- `activeSessionIdAtom` (`packages/electron/src/renderer/store/atoms/sessions.ts:95`) is a global atom, not workspace-scoped.
- The only writer is `AgentMode.tsx:162-166`:
  ```tsx
  useEffect(() => {
    if (actualActiveSessionId) {
      store.set(activeSessionIdAtom, actualActiveSessionId);
    }
  }, [actualActiveSessionId]);
  ```
  When the rail switches to a workspace whose `selectedWorkstreamAtom(workspacePath)` is null, the derived value flips to `null` but the `if` guard skips the write — the global keeps the previous workspace's session id.
- The companion state is already correctly scoped: `selectedWorkstreamAtom` (`sessions.ts:2097`) and `agentModeLayoutAtomFamily` (`agentMode.ts:109`) are atom families keyed by workspace; `sessionRegistryAtom` is global but `refreshSessionListAtom` (`sessions.ts:1882`) replaces the full Map on workspace change.
- Additionally, the rail-switch effect in `App.tsx:1574-1589` only calls `initNavigationHistory` and `initWindowMode`, leaving a small ordering window where readers fire before AgentMode's mount effect runs `initAgentModeLayout` / `initSessionList` for the new path.

omartelo:
Follow-up to https://github.com/nimbalyst/nimbalyst/pull/188#issuecomment-4412981414 — fix applied in `dc6e2bc1` (merge with main) and the subsequent rail-fix commits on this branch.

**How the fix was applied:**

- **Atom-layer subscriber as single source of cross-workspace hygiene.** Added `attachWorkspaceSwitchCleanup(jotaiStore)` in `packages/electron/src/renderer/store/atoms/openProjects.ts`. It subscribes to `activeWorkspacePathAtom` and zeroes `activeSessionIdAtom` on every flip. `initOpenProjects` wires it via `unsubscribers.push`, so it lives for the renderer's lifetime. Exporting the helper as a stand-alone unit lets jotai-only tests attach it to a `createStore()` instance and assert the clearing behavior without booting the whole IPC stack.

- **Drop the `if` guard in AgentMode.** `packages/electron/src/renderer/components/AgentMode/AgentMode.tsx:162-168` now mirrors `actualActiveSessionId` to the global atom unconditionally — including `null`. AgentMode is no longer the only path that can clear the global; the subscriber covers ordering windows where AgentMode hasn't re-rendered yet.

- **Pre-warm the new workspace from the rail-switch effect.** `packages/electron/src/renderer/App.tsx` rail-switch effect now also calls `initAgentModeLayout(railActivePath)` and `initSessionList(railActivePath)` so the per-workspace layout family and session registry are loaded before any reader fires. Both helpers are idempotent — AgentMode's mount effect remains safe.

**Tests:**

- Three jotai unit tests for `attachWorkspaceSwitchCleanup` in `packages/electron/src/renderer/store/atoms/__tests__/openProjects.test.ts`: clears on flip, clears on flip-to-null, stops clearing once the returned unsubscribe is invoked.
- E2E regression scenario in `packages/electron/e2e/multi-project/multi-project-rail.spec.ts`: after registering a fresh workspace and switching the rail to it, the agent panel must render `.agent-mode-empty` (no transcript leak from the previous workspace).

No new global state introduced; workspace-scoped atom families (`selectedWorkstreamAtom`, `agentModeLayoutAtomFamily`) keep the existing pattern. Promoting `activeSessionIdAtom` itself to an `atomFamily(workspacePath)` was considered and deferred — it would touch every reader (notification listeners, voice mode, VoiceModeButton) without a corresponding correctness benefit once the subscriber is in place.

omartelo:
While wiring the new `exitPlanMode:resolved` listener for the "Thinking…" indicator fix, I matched the file-wide `provider.removeAllListeners(event)` + `on(...)` idiom that every other listener in `MessageStreamingHandler.handle` uses. The idiom is intentional today (every `ai:sendMessage` re-registers, so the blanket removal prevents stacked duplicates on the cached per-session provider) and is safe under current architecture because `MessageStreamingHandler` is the sole subscriber to those events.

It is also a future trap: a new module that subscribes to any of those 13 events would be silently clobbered on the next `sendMessage`. Fixing it for one listener in isolation would only diverge the file's convention without removing the trap, so I deferred the cleanup to a dedicated change. Tracked as #225.

omartelo:
## Hover tooltip on rail icons — properly fixed this time

Earlier on this branch (commit cc00f43c, "richer hover tooltip") the rail tooltip was reworked at the CSS/markup level, but it never actually rendered: the rail container has `overflow-x: hidden`, and the tooltip span lived inside it positioned at `left: 100%`. Anything outside the rail's right edge was silently clipped, so the `:hover > .project-rail-tooltip` opacity gate was never observed. Result: hovering a project icon (or the `+` button) showed nothing.

**Root cause**

`packages/electron/src/renderer/components/ProjectRail.css:13` — `overflow-x: hidden` on `.project-rail`. CSS doesn't allow mixing `overflow-y: auto` with `overflow-x: visible`; the browser coerces both to `auto`, so the cheap escape ("just unhide the X axis") is not available.

**Fix**

Refactor both tooltips (per-project icon and the `+` button) to `@floating-ui/react` with `FloatingPortal`, matching the project's existing rule in CLAUDE.md ("Never manually calculate `position: fixed` coordinates for floating UI elements — always use @floating-ui/react with FloatingPortal").

- Each icon and the add button get a dedicated `useFloating` + `useHover` + `useInteractions` instance.
- Tooltip body renders inside `FloatingPortal`, escaping `overflow: hidden` on the rail container.
- Position via `offset(12) + flip + shift({ padding: 8 })`; placement `right`.
- 200 ms open delay so quick mouse passes don't flicker the tooltip.

CSS now carries appearance only — positioning, opacity transitions, and the parent-hover gate are gone. Added a soft `box-shadow` so the portaled tooltip reads as floating UI rather than inline content.

**Visual confirmation**

Project hover (name + path):

image
(Tooltip looks transparent in the screenshot due to a Hyprland capture artifact — it's opaque in actual use.)

<br/><br/>

Add button hover (`+`):

image

ademczuk:
Hi @omartelo, this is a serious architectural piece and I'm not going to do a +4496-line line-by-line tonight. Reviewed at the architecture and state-isolation layer, since those are the spots where multi-project rail can leak silent cross-workspace pollution. One blocker and two pre-merge concerns:

**1. BLOCKER - runtime AI tool calls route to the active project's file system service, not the session's project.**

`packages/runtime/src/ai/tools/fileTools.ts:41`:

```ts
const fileSystemService = getFileSystemService();
```

This is the no-arg getter, which returns the runtime-global singleton - i.e. the currently active project's service. The PR sets up the `fileSystemServices` map in the main process, keyed by `workspacePath`, which is correct. But the AI tool factory still pulls the global.

Concrete failure mode: user has projects A and B both warm. Session running in project A invokes `Write_file` while project B is the active rail tab. The write goes through `getFileSystemService()`, gets B's service, and writes the file into B's workspace, not A's. Same class of bug as the workspace-scoped IPC rule in `CLAUDE.md` ("workspace-scoped IPC must take `workspacePath` as a required parameter, never fall back to module-level state").

The PR's MultiProjectRailHandlers comment around line 160 acknowledges the routing requirement - "Sessions running in inactive rail projects must resolve their FS service via the per-path map" - but no tool-call path was actually wired to do that. Either:

- Pass `effectiveWorkspacePath` into the tool factory and have tools call `fileSystemServices.get(path)` instead of `getFileSystemService()`, or
- Block multi-project rail from being enabled while any background session is active (feature-flag-with-active-session-guard).

Can't merge as-is without one of those.

**2. PRE-MERGE - E2E test suite uses `page.reload()` between assertions, never exercises the hot path.**

`e2e/multi-project/multi-project-rail.spec.ts`: every meaningful state assertion is preceded by a full renderer reload. The tests validate "state survives an IPC write + renderer boot cycle," but not "state survives a rail-switch in the same renderer process" - which is what users will actually hit. The known session-leak bug you caught manually was the in-memory switch path, and no test exercises it.

Worth one Playwright test that switches projects without a reload and asserts the active session is correctly cleared, the file watcher moves, and the document service routes to the new active path.

**3. PRE-MERGE - `activeSessionIdAtom` cleared on every rail switch creates a transient null window.**

The `attachWorkspaceSwitchCleanup` subscriber zeroes `activeSessionIdAtom` on rail flip, expecting `AgentMode`'s mount effect to repopulate it. The PR body acknowledges this deferral. The window between subscriber-fires and AgentMode-effect-runs is short but not zero, and there's no ordering guarantee on Jotai atom subscribers vs React reconciliation. Any reader during that window (notification listeners, voice mode, `VoiceModeButton`, anything observing session state for UI updates) sees null.

In practice probably a brief flash. In edge cases (slow render, large transcript hydration, voice mode mid-utterance) it could surface as a real bug. Promoting `activeSessionIdAtom` to `atomFamily(workspacePath)` is the proper fix and the PR notes it as deferred - worth either doing it now or filing a tracking issue against the merged PR so it doesn't get lost.

**Nice-to-have, not blocking:**

- `MAX_OPEN_PROJECTS = 8` is enforced renderer-side but not in `workspace:register-additional`. A second window or programmatic IPC could register a 9th path and the renderer would silently lose track. Mirror the cap in the main-process handler.
- PGLite single-database is pre-existing design and not made worse by this PR. Worth a one-line note that future per-project DB isolation would extend the `serviceRegistry` pattern naturally.
- Migration path for existing single-project users is fine - feature flag default-off, all atom families fall back to the primary workspace.

Architecture overall is sound - per-workspace atom families is the right primitive, dual-layer keep-warm (main-process service maps + renderer atom families) is clean, and the `serviceRegistry.ts` extraction breaks the circular import correctly. Just needs the fileTools routing closed.

ademczuk:
Walking back the BLOCKER framing in my prior comment after re-reading the diff. I missed this hunk in `MultiProjectRailHandlers`:

```ts
const fsService = fileSystemServices.get(workspacePath);
if (fsService) {
    setFileSystemService(fsService);
}
```

The PR explicitly wires the active-side flip on rail switch. My "no tool-call path was actually wired" reading conflated the documented constraint (inactive sessions must use the per-path map) with a missing implementation, when the active-project path IS implemented.

The runtime-side question still exists but at concern severity, not blocking: `packages/runtime/src/ai/tools/fileTools.ts:41` uses `getFileSystemService()` no-arg. Whether AI tool dispatch for sessions in inactive rail projects actually fires through the runtime tool registry, or gets fenced off upstream, depends on a code path I didn't trace end-to-end before posting. If you've already verified inactive-project sessions either resolve via the per-path map or can't dispatch tools while inactive, this concern is moot.

Concerns 2 and 3 (e2e test reload pattern, `activeSessionIdAtom` transient null window) still stand at the original pre-merge severity.

Apologies for the overstated initial framing.

omartelo:
Thanks for the architectural review, addressed all three pre-merge concerns in the latest commit on this branch. Quick status:

**1. fileTools per-workspace dispatch (was BLOCKER, walked back to concern)** — fixed.

The vulnerability was reachable, confirmed by a code-trace. Sessions in inactive rail projects could route `searchFiles` / `listFiles` / `readFile` through whichever workspace's `FileSystemService` happened to be on top of the global, because the no-arg `getFileSystemService()` is flipped on rail switch. Added a per-path registry in `runtime/core/FileSystemService.ts` and widened `ToolDefinition.handler` to accept a `ToolContext` carrying `workspacePath`. The electron-side `ToolExecutor` (already keyed by `workspaceId`) now passes that context through, and the rail handlers mirror their `fileSystemServices` map into the runtime registry on register/unregister. `fileTools` resolves the per-path entry first and falls back to the global only when the dispatcher has no workspace context, so legacy single-project paths keep working.

Five new unit tests in `fileTools.test.ts` cover the per-workspace routing (active vs inactive workspace), the legacy fallback, and the missing-path degrade.

SDK-native tools (`Read` / `Write` / `Edit` / `Bash`) were already correctly fenced by the SDK subprocess `cwd`, so this was specifically the runtime-internal tool registry path.

**2. E2E never exercises the in-process rail switch** — fixed.

Added `rail click updates workspace context in-process (no reload)` in `multi-project-rail.spec.ts`. Clicks between two rail icons and asserts `.workspace-summary-header-path` reflects the new workspace plus `.agent-mode-empty` is visible — without any `page.reload()`. The previous tests' reload-driven setup is preserved for boot-cycle coverage.

**3. `activeSessionIdAtom` transient-null window** — fixed (option B from internal discussion).

Rewrote `attachWorkspaceSwitchCleanup` to compute the new value synchronously from `selectedWorkstreamAtom(newPath)` + `workstreamActiveChildAtom`, mirroring AgentMode's own derivation. The earlier "set to null and let AgentMode repopulate on mount" pattern is gone, so readers between the subscriber firing and the React commit see the correct value rather than a transient null. AgentMode's mount-effect write later converges on the same value.

Considered promoting `activeSessionIdAtom` to `atomFamily(workspacePath)` per the original suggestion but deferred — would touch every reader (notification listeners, voice mode, VoiceModeButton, internal writers) for no marginal correctness benefit once option B is in place. Happy to revisit if a workspace-scoped reader pattern shows up that justifies the broader refactor.

**Nice-to-haves (deferred, will address in a follow-up if warranted):**

- `MAX_OPEN_PROJECTS = 8` cap is enforced renderer-side but not in `workspace:register-additional`. A second window or programmatic IPC could push past it. Worth mirroring in the main handler — cheap fix, low priority since the renderer is the only path users currently hit.
- One-line note that future per-project DB isolation extends the `serviceRegistry` pattern naturally — happy to add it to the architecture comment block if useful.

Tests:

- `npx vitest run src/renderer/store/atoms/__tests__/openProjects.test.ts` → 18/18 pass
- `npx vitest run packages/runtime/src/ai/tools/__tests__/fileTools.test.ts` → 20/20 pass (5 new for per-workspace dispatch)
- TypeScript clean in both `runtime` and `electron`

ademczuk:
Thanks @omartelo, this is a substantial response and the fix design looks right to me.

**Fix shape**: The ToolContext widening + per-path registry approach is the right architectural call. `ToolExecutor` passing `{ workspacePath: this.workspaceId }` is the load-bearing piece, and `resolveFileSystemServiceForCall(ctx)` preferring per-path lookup with fallback to the legacy global keeps single-project callers working without forking the runtime tool definitions. The comment in `ToolExecutor.ts:347-353` directly identifies the bug class, which is a nice touch for future readers.

**On severity**: Glad the code-trace confirmed the runtime path was reachable. That is the part I should have verified before stamping BLOCKER, not after the walk-back. The architectural concern was real, the framing was just over-confident.

**Tests**: 5 new unit tests covering active vs inactive workspace dispatch, legacy fallback, and missing-path degrade is the right coverage shape for this kind of cross-package state hygiene fix.

The `atomFamily` deferral on `activeSessionIdAtom` is a sensible call. Option B (synchronous derivation from `selectedWorkstreamAtom` + `workstreamActiveChildAtom`) gets correctness without the broader refactor cost, and the boundary for needing `atomFamily` would be clear if a workspace-scoped reader pattern shows up later.

Reads as ready to me, deferring to @ghinkle on the merge call. Thanks for taking the architectural concerns seriously.

ghinkle:
I just wanted to say that I've been following along with this PR and have tested it locally and I appreciate the work from both of you. 

I've got a couple big items in flight and I'm just trying to decide the ordering on merge back. Hoping to land it tonight or tomorrow.

omartelo:
Merged latest `main` into the branch (commit `fda6bdf1`) to unblock the merge ordering @ghinkle mentioned.

**Conflict resolution status**

No textual conflicts. Git auto-merged two files with overlapping but non-conflicting changes from `main`:

- `packages/electron/src/main/mcp/tools/interactiveToolHandlers.ts` — touched by `6ef48c75` (`PromptForUserInput` tool) and by this branch's interactive-prompt routing work. Edits landed in disjoint regions.
- `packages/electron/src/renderer/store/sessionStateListeners.ts` — touched by `6ef48c75` and by this branch's `session:streaming` regression fix (#188 / `523a625a`). Edits landed in disjoint regions.

**Verification post-merge**

- `npx tsc --noEmit` on `packages/electron` → 0 errors
- `npx tsc --noEmit` on `packages/runtime` → 0 errors

Branch is `MERGEABLE` / `CLEAN` against `main` and the test commits from the prior rail-fix iterations are preserved.

---

Thanks to everyone who put time into this one:

- @karlwirth — for the careful UX pass that surfaced the light-mode legibility regression, the confusing title-bar source-of-truth, and the two-letter project-icon ambiguity. All three pointed at real architectural defects, not just polish.
- @ademczuk — for the architecture-level review and the state-isolation lens. The `fileTools` routing concern was the right thing to escalate; the walked-back framing didn't take away from the fact that the underlying vulnerability was reachable and got the per-workspace `ToolContext` plumbing it needed. The e2e hot-path and transient-null concerns were equally load-bearing.
- @ghinkle — for testing the branch locally and for the merge-ordering visibility. Glad to have this stay coordinated with your in-flight work rather than land out-of-order.

Implementation is validated and ready when the ordering works for you, @ghinkle.

PR #193, opened 2026-05-07
fix(automations): detect automationStatus + tolerate CRLF on Windows (fixes #67, #68)

### LINKED ISSUE #67: Tracker view ignores automationStatus-keyed automation files
**Title:** Tracker view ignores `automationStatus`-keyed automation files

**Body:**

```markdown
## Summary

The Tracker view is empty even when the workspace contains valid automation files in `nimbalyst-local/automations/`. The automations themselves run fine and appear in the document header — they just never show up as Tracker rows.

Root cause is an asymmetry between two frontmatter parsers in the renderer: `detectTrackerFromFrontmatter` recognises the `automationStatus` extension-owned key (and lifts its nested fields), but `resolveTrackerFrontmatter` (used by the Tracker view loader) only recognises `trackerStatus.type`. So files that the automations extension itself produces are invisible to the Tracker view.

**Version:** 0.57.33 (Windows 11)

## Reproduction

1. Create a file at `nimbalyst-local/automations/test.md` via the `/automation` command (or write one by hand) with the standard frontmatter the extension emits:
   ```yaml
   ---
   automationStatus:
     id: test
     title: Test
     enabled: true
     schedule:
       type: daily
       time: "09:00"
   ---
   ```
2. Open the Tracker view.
3. **Expected:** the automation appears as a row with type "automation".
4. **Actual:** Tracker view is empty. The file's document header (top of the editor) correctly shows the automation schedule, confirming the file is parsed and recognised elsewhere.

## Root cause

In the renderer bundle `index-CAJHKe60.js` (~line 3198):

```js
function resolveTrackerFrontmatter(frontmatter, trackerType) {
  if (!frontmatter) return null;
  if (frontmatter.trackerStatus && typeof frontmatter.trackerStatus === "object") {
    const trackerData = frontmatter.trackerStatus;
    if (trackerData.type === trackerType) {
      const { trackerStatus: _2, ...topLevel } = frontmatter;
      return { ...trackerData, ...topLevel };
    }
  }
  return null;
}
```

This is the gate that `convertFullDocumentToTrackerItems` uses to filter the metadata list. Since automation files write `automationStatus`, never `trackerStatus`, they're filtered out.

Compare with `detectTrackerFromFrontmatter` in `registerExtensionSystem-CZr3Upen.js` (~line 169529), which already does the right thing using:

```js
const EXTENSION_OWNED_KEYS = {
  automationStatus: "automation"
};
```

…and lifts `extData` into the merged record. The Tracker view loader is missing this exact mapping.

## Suggested fix

Apply the `EXTENSION_OWNED_KEYS` mapping inside `resolveTrackerFrontmatter` before the `trackerStatus` check, mirroring `detectTrackerFromFrontmatter`. Roughly:

```js
function resolveTrackerFrontmatter(frontmatter, trackerType) {
  if (!frontmatter) return null;

  for (const [extKey, mappedType] of Object.entries(EXTENSION_OWNED_KEYS)) {
    if (mappedType === trackerType && frontmatter[extKey] && typeof frontmatter[extKey] === "object") {
      const extData = frontmatter[extKey];
      const { [extKey]: _x, trackerStatus: _ts, ...topLevel } = frontmatter;
      return { ...extData, ...topLevel };
    }
  }

  if (frontmatter.trackerStatus && typeof frontmatter.trackerStatus === "object") {
    const trackerData = frontmatter.trackerStatus;
    if (trackerData.type === trackerType) {
      const { trackerStatus: _2, ...topLevel } = frontmatter;
      return { ...trackerData, ...topLevel };
    }
  }
  return null;
}
```

This also fixes a secondary issue: even after a user manually adds `trackerStatus: { type: automation }` to their file as a workaround, the Tracker columns (title, schedule, lastRun, runCount) come up blank because the loader doesn't lift fields out of the nested `automationStatus` block — only the file-name fallback shows. Lifting `extData` as above populates the columns automatically.

## Workaround

Add `trackerStatus: { type: automation }` to the top of each automation file's frontmatter alongside the existing `automationStatus` block. Both keys must coexist — the extension keeps reading `automationStatus`, the Tracker view picks up `trackerStatus`. Note this leaves columns empty (only file name shows).
```

### LINKED ISSUE #68: Frontmatter regex breaks on CRLF — automations and Tracker miss files on Windows
## Summary

Several frontmatter parsers in the renderer and the bundled `automations` extension use `/^---\n([\s\S]*?)\n---/`, which fails to match when files have CRLF line endings. On Windows this means automation `.md` files checked out from git (or saved by some external tools) become invisible to the parser entirely — the automations extension can't see them, the Tracker view can't see them, and the document header doesn't recognise the file as an automation.

The main-process parser uses `/^---\s*\n([\s\S]*?)\n---/` (tolerant of `\r` before the newline) and works correctly. The inconsistency between parsers is the bug.

**Version:** 0.57.33 (Windows 11)

## Reproduction

1. On Windows, clone or pull a repo that contains automation files in `nimbalyst-local/automations/*.md` without an explicit LF rule in `.gitattributes`. Git's `core.autocrlf` will check them out as CRLF.
2. Open Nimbalyst.
3. **Expected:** automations appear in the Tracker view and run on schedule.
4. **Actual:** the extension's `automations.list` returns nothing for those files; Tracker view is empty; the document header doesn't show the schedule strip.
5. Convert one file to LF (`sed -i 's/\r$//' file.md` or re-save in an editor with LF). Within ~30s the automation appears in `automations.list` and the document header populates.

## Root cause

Three parsers share the same brittle pattern:

**Extension** — `resources/extensions/automations/dist/index.js` (~line 1343):
```js
const te = /^---\n([\s\S]*?)\n---/;
```

**Renderer (TrackerPlugin / extension system)** — `registerExtensionSystem-*.js`, `extractFrontmatter$1` (~line 169505):
```js
const frontmatterRegex = /^---\n([\s\S]*?)\n---/;
```

The same file's `updateFrontmatter` (~line 169611) and `updateTrackerInFrontmatter` use the variant `/^---\n[\s\S]*?\n---\n?/`, which has the same problem.

**Main process** — `out/main/index.js` `extractFrontmatter` (~line 234511) — already correct:
```js
const yamlMatch = frontmatterContent.match(/^---\s*\n([\s\S]*?)\n---/);
```

So `document-service:metadata-list` correctly parses CRLF files into a JS object, but the renderer-side gates that consume that data (and the extension's own file reader) reject it before any UI code runs.

## Suggested fix

In each renderer/extension parser, change `\n` to `\r?\n`:

```js
// extension parser
const te = /^---\r?\n([\s\S]*?)\r?\n---/;

// renderer extractFrontmatter$1
const frontmatterRegex = /^---\r?\n([\s\S]*?)\r?\n---/;

// updateFrontmatter / updateTrackerInFrontmatter
const frontmatterRegex = /^---\r?\n[\s\S]*?\r?\n---\r?\n?/;
```

Or align all parsers on the main process's `\s*\n` style. Either is fine; consistency is the point.

## Workaround

Pin automation files to LF via `.gitattributes` at the repo root:

```
nimbalyst-local/automations/*.md text eol=lf
```

Then re-save any existing files (`sed -i 's/\r$//' nimbalyst-local/automations/*.md`) to convert in place. Files reappear within the next 30s rescan.

## Why this matters

Silent failure mode — no error in the console, no indication in the UI that the file is malformed. A user who clones their workspace on a fresh Windows machine sees empty Tracker view and no scheduled automations, with no obvious cause. Pairs poorly with the related bug (`automationStatus` ignored by Tracker view) where the symptom looks identical but the cause is different.
```

### PR DESCRIPTION
**Pull request by ademczuk**
## Summary

Two related Windows-specific symptoms reported by the same user that combine to keep automation files invisible to the Tracker pipeline:

- Fixes #67: documents with an `automationStatus` frontmatter block do not appear as rows in the Tracker view, even though `detectTrackerFromFrontmatter` already classifies them correctly.
- Fixes #68: on Windows with `git config core.autocrlf=true`, frontmatter regexes that require bare LF after `---` silently fail. Files arrive as `---\r\n...` and consumers treat them as if they have no frontmatter.

### #67 - tracker resolver ignored extension-owned keys

`resolveTrackerFrontmatter` in `TrackerTable.tsx` only checked `frontmatter.trackerStatus?.type`. `detectTrackerFromFrontmatter` in `frontmatterUtils.ts` already had an `EXTENSION_OWNED_KEYS` map (`{ automationStatus: 'automation' }`) that the rest of the system used, but `resolveTrackerFrontmatter` never imported it. Two diverging copies of the classification logic; the resolver lost.

### #68 - bare-LF frontmatter regexes (repo-wide audit)

The reporter pointed at three sites. A repo-wide audit (`grep -rn '/\^---' packages`) turned up 13 frontmatter regex sites in total. Sites already tolerant of CRLF (left unchanged):

- `FileHtmlExporter.ts` (`\r?\n`)
- `frontmatterReader.ts` (`\s*\n`)
- `CommandFileParser.ts` (`\s*\n`)
- `ElectronDocumentService.ts` (`\s*\n` / `[\s\S]*?`)
- `editor/markdown/FrontmatterUtils.ts` (`\s*\n`)

Sites patched in this PR (all changed from `^---\n` to `^---\r?\n` on the open and `\n---` to `\r?\n---` on the close):

- `packages/extensions/automations/src/frontmatter/parser.ts:8`
- `packages/runtime/src/plugins/TrackerPlugin/documentHeader/frontmatterUtils.ts:45, 160, 232`
- `packages/electron/src/main/services/ai/MessageStreamingHandler.ts:339`
- `packages/electron/src/renderer/hooks/useIPCHandlers.ts:791`
- `packages/electron/src/renderer/services/AIToolService.ts:188`
- `packages/runtime/src/ai/tools/index.ts:398`
- `packages/runtime/src/editor/plugins/DiffPlugin/core/diffFrontmatter.ts:5`
- `packages/runtime/src/editor/plugins/FloatingDocumentActionsPlugin/TrackerTypeHelper.ts:38, 152, 197`
- `packages/runtime/src/plugins/FrontmatterPlugin/fieldUtils.ts:150, 348`

13 sites in total - the reporter's three plus 10 more I missed in the first pass and added after a wider review.

## Fix

- Export `EXTENSION_OWNED_KEYS` from `frontmatterUtils.ts` so the Tracker resolver can share one source of truth with `detectTrackerFromFrontmatter`.
- Add an extension-owned-key lookup at the top of `resolveTrackerFrontmatter`, mirroring the priority order in `detectTrackerFromFrontmatter`.
- Replace bare `\n` with `\r?\n` in 13 frontmatter regex sites across 6 packages. Output is still canonicalised to LF on save (jsyaml default), so only the read path changes. Round-trip on CRLF input produces LF output that subsequent reads tolerate.

## Tests

22 vitest cases:

- `frontmatterUtils` extension-owned tests gain 3 CRLF round-trip cases
- `parser.test.ts` (new) adds 6 cases for the automations extension parser on CRLF input
- `resolveTrackerFrontmatter.test.ts` (new) adds 6 cases for the #67 resolver contract including a no-false-positive case for non-matching tracker types

```
$ npx vitest run \
    packages/runtime/src/plugins/TrackerPlugin/documentHeader/__tests__/extensionOwnedFrontmatter.test.ts \
    packages/runtime/src/plugins/TrackerPlugin/components/__tests__/resolveTrackerFrontmatter.test.ts \
    packages/extensions/automations/src/frontmatter/__tests__/parser.test.ts

Test Files  3 passed (3)
     Tests  22 passed (22)
```

## Test plan

- [x] all 22 vitest cases pass locally
- [x] `tsc --noEmit` clean for both `packages/electron` and `packages/runtime`
- [ ] CI green
- [ ] Manual on Windows: open a workspace with at least one `/automation` file, confirm it shows in the Tracker view; create a fresh automation with `core.autocrlf=true`, confirm `parseAutomationStatus` round-trips it.

## Regression risk

Low. The CRLF change is purely additive (`\r?` makes `\r` optional; LF input unchanged). The resolver change adds a check before the existing one, with the existing path preserved verbatim. The new check returns null when the extension type does not match the requested tracker type, so it cannot misclassify a non-automation document.

Out-of-scope follow-up that this PR does not attempt: writing CRLF output on round-trip when input was CRLF (we always emit LF). The motivation here is to make the read path tolerant; rewriting line endings is a separate concern.


ademczuk:
@ghinkle two related Windows bugs from the same reporter, batched together. The CRLF regex is purely additive (`\r?\n`), the tracker resolver change uses the same `EXTENSION_OWNED_KEYS` map `detectTrackerFromFrontmatter` already used so the two functions stop diverging. 22 vitest cases covering both fixes.

ademczuk:
Added a follow-up commit (`03cdfb9e`) extending the CRLF tolerance to 10 more frontmatter regex sites I missed in the first pass. The reporter pointed at 3, but `grep -rn '/\^---' packages` shows 13 total - sites unchanged are the ones already using `\s*\n` or `\r?\n` (FileHtmlExporter, frontmatterReader, CommandFileParser, ElectronDocumentService, editor markdown FrontmatterUtils). PR description updated with the full site list. CI re-running.

ademczuk:
Two follow-up notes after a deeper review pass:

**Inline resolver** (`d1e90610`): `ElectronDocumentService.parseInlineTrackerCandidate` had its own copy of the tracker classification logic that didn't consult `EXTENSION_OWNED_KEYS`. That meant the document-service IPC import path would reject automation documents even after the resolver fix in `TrackerTable` landed. Patched to use the same map and priority order. Three resolver sites are now in sync (`detectTrackerFromFrontmatter`, `resolveTrackerFrontmatter`, `parseInlineTrackerCandidate`).

**CRLF -> LF write-side normalisation** (known trade-off, not patched): `jsyaml.dump()` always emits `\n` line endings, so when a CRLF input file is updated through any of the patched paths, the written-back file is canonicalised to LF. For users with `* text=auto` in `.gitattributes` this triggers a one-time `git status` modification on first save. There is no CRLF-output mode in jsyaml; preserving CRLF would need a separate post-dump conversion that detected the original ending and put them back. I think LF canonicalisation is the right default but it is a real side effect worth knowing about.

**Cosmetic** (out of scope): `NewFilePreview.stripFrontmatter` uses `indexOf('\n---', 3)` and would leave a stray `\r` at body start on CRLF input. The leading `\r` is invisible in rendered preview, so the practical impact is zero. Easy follow-up if you want it tightened.

ghinkle:
Yea, trackers is an inprogress feature with more changes likely, but merging this as is.

PR #194, opened 2026-05-07
fix(tray): support Linux and Windows in TrayManager (fixes #39)

### LINKED ISSUE #39: Linux: Tray icon not showing — TrayManager skips non-macOS platforms
## Description

The system tray icon does not appear on Linux. `TrayManager.initialize()` in the main Electron process exits early on non-macOS platforms:

```js
if (process.platform !== "darwin") {
  logger.info("[TrayManager] Skipping initialization on non-macOS platform");
  return;
}
```

## Additional issue

The tray template icon files (`trayTemplate.png`, `trayTemplate@2x.png`) are not shipped in the `resources/` directory of the AppImage, so even removing the platform check alone wouldn't be enough — the icons need to be bundled as well.

## Steps to reproduce

1. Install Nimbalyst on Linux (AppImage)
2. Launch the app
3. No tray icon appears in the system tray

## Expected behavior

A tray icon should appear in the system tray on Linux, similar to macOS.

## Suggested fix

1. Remove or relax the `process.platform !== "darwin"` guard in `TrayManager.initialize()`
2. Guard the macOS-specific `systemPreferences.subscribeNotification("AppleInterfaceThemeChangedNotification", ...)` call behind a `process.platform === "darwin"` check (this API is macOS-only and would crash on Linux)
3. Bundle `trayTemplate.png` and `trayTemplate@2x.png` in the `resources/` directory of the Linux build

## Environment

- **OS:** Ubuntu Linux (x86_64)
- **Nimbalyst version:** Latest AppImage from GitHub releases
- **Desktop:** GNOME with system tray support

ademczuk:
## Summary

The system tray entry has been macOS-only since the feature shipped. Linux AppImage and Windows builds had no tray icon despite Electron's `Tray` API working on all three platforms and the icon assets being in the bundle already.

Fixes #39.

## Root cause

`TrayManager.initialize()` started with:

```ts
if (process.platform !== 'darwin') {
  logger.main.info('[TrayManager] Skipping initialization on non-macOS platform');
  return;
}
```

A few lines below the guard, the method called `systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', ...)`, which is genuinely macOS-only - removing the guard alone would crash on Linux/Windows. So both halves needed handling.

`trayTemplate.png` and `trayTemplate@2x.png` were already in `packages/electron/resources/` and already declared in `extraResources` (no platform filter), so they ship to every OS today. No new assets needed.

## Fix

- Drop the darwin early-return.
- Wrap `systemPreferences.subscribeNotification` and `unsubscribeNotification` in `if (process.platform === 'darwin')`. The teardown closure now checks if a subscription id was captured before unsubscribing.
- Gate `tray.setTitle(...)` (called three times in `updateIcon`) on darwin too. Per Electron docs, `setTitle` is macOS-only; the explicit guard keeps intent obvious.

The cross-platform `nativeTheme.on('updated', ...)` listener stays unguarded, since it's the icon-redraw signal on every OS.

## Tests

New `TrayManager.platform.test.ts` mocks `process.platform` plus the full Electron module surface and asserts:

```
$ npx vitest run packages/electron/src/main/tray/__tests__/TrayManager.platform.test.ts

 [PASS] does not return early on Linux
 [PASS] does not return early on Windows
 [PASS] still subscribes the macOS appearance notification on darwin

Test Files  1 passed (1)
     Tests  3 passed (3)
```

Each platform stub:
- restores `process.platform` after the case (no global state leak between tests),
- resets the `TrayManager` singleton so each `it()` runs fresh,
- mocks `getSessionStateManager`, the database, `electron-store` getters, and the rest of the dependency surface to keep the test focused on the platform branch.

## Files changed

- `packages/electron/src/main/tray/TrayManager.ts` - drop darwin early-return, wrap two macOS-only API call sites
- `packages/electron/src/main/tray/__tests__/TrayManager.platform.test.ts` - new, three platform cases
- `CHANGELOG.md` - entry under [Unreleased] > Fixed

## Test plan

- [x] vitest passes locally (3/3)
- [x] `tsc --noEmit` clean
- [ ] CI green
- [ ] Manual on Linux: install the AppImage, confirm a tray entry appears with the correct icon and that no `AppleInterfaceThemeChangedNotification` errors land in the main-process log
- [ ] Manual on Windows: build, run, confirm tray entry appears
- [ ] Manual on macOS: confirm tray still updates on system appearance changes (template icon flip on light/dark)

## Regression risk

Low. The two newly-active platforms had no tray functionality before, so there is no behaviour to regress on Linux/Windows. macOS behaviour is preserved by the explicit darwin paths around `systemPreferences.subscribeNotification` and `tray.setTitle`. The `nativeTheme` listener was already cross-platform and is unchanged.


ademczuk:
@ghinkle small platform-guard fix - removes the darwin-only early return, wraps the two macOS-only API calls (systemPreferences subscription + tray.setTitle). Icon assets already shipped to all platforms, so no resource changes needed. Three vitest cases mock process.platform to verify Linux/Windows no longer bail out and macOS still subscribes the appearance notification.

ademczuk:
Caveat worth flagging: the bundled `trayTemplate.png` is a macOS template image (monochrome, designed for system tinting). On Linux/Windows where `setTemplateImage(true)` is a documented no-op, it renders as-is. The tray ENTRY appears, the menu works, and badge counts function on macOS - but the icon may look basic on Linux/Windows until a colour variant is added. Out of scope for this PR; happy to follow up with platform-specific icons if useful.

ademczuk:
Two follow-up notes from a deeper review pass:

**Linux runtime dependency (not crashes, but invisible).** Electron's `Tray` constructor on Linux requires `libayatana-appindicator3-1` (or `libappindicator3-1` on older distros). On a fresh Ubuntu/Debian install without the package, `new Tray(icon)` silently produces no visible icon. The fix in this PR makes the code run on Linux and Windows, but Linux users on minimal installs will need that package. Worth a one-line note in the README install section if you want to ship this end-to-end - happy to add it as a follow-up commit if useful.

**Icon visual quality on non-darwin.** The bundled `trayTemplate.png` is a monochrome macOS template image. `setTemplateImage(true)` is a no-op on Linux/Windows, so the icon renders as solid black - fine on light Linux panels, hard to see on dark panels (GNOME dark theme, KDE Breeze Dark). The error/attention states use the explicit blue-dot bitmap path which renders better on dark panels. A proper colour-adapted Linux/Windows icon set is a separate concern (new asset + `nativeTheme` swap logic) and out of scope here.

The fix itself is correct - Linux/Windows users get a tray entry + functional menu. These are quality polish items rather than blockers.

Issue #197, opened 2026-05-07
Bundle Node.js runtime with MCP so it doesn't require a system install


ablac:
## Problem
Nimbalyst's MCP functionality depends on a system-installed Node.js. When Node.js isn't already on the machine, this creates a setup blocker — the user has to install Node.js manually before MCP features work, which isn't obvious from the Nimbalyst install flow.

## Proposed solution
Bundle a Node.js runtime as part of the MCP component of Nimbalyst, so the MCP server uses the bundled runtime instead of requiring a system-wide Node.js installation. This would let users install Nimbalyst and have everything work out of the box.

## Alternatives considered
- Documenting Node.js as a prerequisite (easy to miss, and shifts setup burden onto the user)
- Detecting missing Node.js on first run and prompting/auto-installing it

## Additional context
Encountered on Windows 11 — Node.js was not installed, and had to be installed manually via `winget install OpenJS.NodeJS.LTS` before things worked.

ademczuk:
Posted PR [#198](https://github.com/nimbalyst/nimbalyst/pull/198) which addresses the underlying symptom (MCP servers using `command: "node"` need a system Node install) by routing those through Electron's bundled Node runtime via `ELECTRON_RUN_AS_NODE=1` - no separate Node binary needed in the installer. Awaiting maintainer review on whether this scope matches what you had in mind, or whether they'd prefer a literal Node bundle.

ademczuk:
Thanks for filing this, @ablac. The end-user outcome you described (MCP servers using `command: "node"` working without a system-wide Node install) shipped in v0.60.0 via #198. The implementation took a slightly different path than the bundled-Node approach you suggested: instead of including a separate Node binary in the installer, Nimbalyst now routes bare `node` commands through Electron's bundled runtime via `ELECTRON_RUN_AS_NODE=1`. Same outcome, no extra installer size.

Absolute Node paths (system Node, nvm, custom installs) are deliberately left untouched, so anyone who pinned a specific Node version keeps that.

Closing as completed. If you hit a Node-not-found error on v0.60.0 or later, please reopen with the MCP server config and OS details.

PR #198, opened 2026-05-07
feat(mcp): use Electron's bundled Node runtime for stdio MCP servers (refs #197)

ademczuk:
## Summary

Issue #197 reports that fresh installs without a system-wide Node.js install can't run MCP servers configured with `command: "node"`. The OP had to `winget install OpenJS.NodeJS.LTS` on Windows 11 before Nimbalyst's MCP features worked. The OP's suggested fix was to bundle a Node.js binary in the installer; this PR takes a smaller-scope path that achieves the same end-user outcome at zero installer-size cost.

Refs #197.

## Approach

Electron itself is a Node binary. With `ELECTRON_RUN_AS_NODE=1`, the Electron executable acts as a plain Node process for the child. So instead of shipping a 30-50 MB Node binary per platform, route bare `node` commands through the runtime that's already in the bundle.

`MCPConfigBuilder.processServerConfig` now rewrites bare `node` tokens to `process.execPath` and injects `ELECTRON_RUN_AS_NODE=1` into the spawn env.

## Substitution rules

Triggers when ALL conditions hold:

- transport is stdio (not SSE)
- `command` is the bare token `node` / `node.exe` (case-insensitive, with optional `.exe`)
- we're running inside Electron (`process.versions.electron` is a non-empty string)

Does NOT touch:

- absolute paths (`/usr/local/bin/node`, `/Users/me/.nvm/versions/node/v22.11.0/bin/node`, `C:\Program Files\nodejs\node.exe`) - a user who pinned a specific Node version keeps that choice
- other runtimes (`npx`, `npm`, `bun`, `deno`, `python`) - the bundled Node runtime can't stand in for those
- non-Electron environments (vitest, CI, mobile builds) - behaviour unchanged

## Tests

New `MCPConfigBuilder.bundledNode.test.ts`, 14 cases:

- `isBareNodeCommand` helper: matches `node`, `node.exe`, case variants, whitespace; rejects absolute paths, other runtimes, undefined, empty
- substitution: triggers in Electron, preserves user env vars, leaves `args` unchanged
- substitution: skipped for absolute paths, other runtimes, SSE transport, non-Electron environments

```
$ npx vitest run packages/runtime/src/ai/server/mcp/__tests__/MCPConfigBuilder.bundledNode.test.ts
Test Files  1 passed (1)
     Tests  14 passed (14)
```

## Files changed

- `packages/runtime/src/ai/server/mcp/MCPConfigBuilder.ts` - 45 lines (substitution block + static `isBareNodeCommand` helper)
- `packages/runtime/src/ai/server/mcp/__tests__/MCPConfigBuilder.bundledNode.test.ts` - new, 172 lines, 14 cases
- `CHANGELOG.md` - entry under `[Unreleased] > Changed`

## Test plan

- [x] 14 vitest cases pass locally
- [x] `tsc --noEmit` clean for both `packages/runtime` and `packages/electron`
- [ ] CI green
- [ ] Manual on a fresh Windows 11 box without Node installed: verify a `command: "node"` MCP server (e.g., something from `mcpServers` in `~/.claude.json`) spawns correctly via Electron's bundled runtime

## What this does NOT cover (out of scope)

The OP's full ask was "bundle Node.js." This PR addresses the symptom (MCP servers using `command: "node"` work without system Node) without literally bundling a separate Node binary. Two follow-ups remain that may need maintainer design buy-in:

- **`npx`-spawned servers still need system Node + npm.** Many MCP servers in the wild use `command: "npx"` to run via `npx -y @modelcontextprotocol/server-...`. Supporting those without a system Node install is a separate question: does Nimbalyst ship a managed package cache? Where do new packages get installed? Probably worth its own discussion.
- **No PATH discovery / "which node" fallback.** If a user has `command: "node"` and is running OUTSIDE Electron (vitest, CI, mobile build), behaviour is unchanged - we just look up `node` on PATH the same as before.

Happy to follow up with either if maintainers want them.

## Regression risk

Low. The substitution is additive (new `if` branch in an existing method). It only fires under all three of [stdio + bare-node + electron], so:

- Outside Electron: zero change. No-electron test verifies this.
- SSE servers: untouched.
- Absolute-path Node references: untouched.
- Non-Node commands: untouched.
- Bare `node` in Electron: now uses `process.execPath` with `ELECTRON_RUN_AS_NODE=1`. The latter is a documented Electron contract that's been stable for years (used by Electron's own internal tooling).

If the Electron version ever drops `ELECTRON_RUN_AS_NODE`, the substitution would still spawn `process.execPath` but the child would be Electron-trying-to-run-as-Electron, not Node. That would fail loudly at startup rather than silently. We can add a feature-detect / opt-out env var if maintainers want belt-and-braces.

### DIFF
--- a/packages/electron/src/main/services/MCPConfigService.ts
+++ b/packages/electron/src/main/services/MCPConfigService.ts
@@ -822,6 +822,8 @@ export class MCPConfigService {
    * Process a server config for runtime use.
    * On Windows, converts npm/npx/etc commands to their .cmd equivalents.
    * Transparently wraps HTTP transport with mcp-remote.
+   * Routes bare `node` commands through Electron's bundled Node runtime so
+   * MCP servers do not require a system-wide Node install (see #197).
    */
   processServerConfigForRuntime(serverConfig: MCPServerConfig): MCPServerConfig {
     // First, convert HTTP to stdio with mcp-remote wrapper
@@ -832,6 +834,38 @@ export class MCPConfigService {
       return config;
     }
 
+    // Substitute bundled Electron Node runtime for bare `node` commands.
+    // Electron itself is a Node binary; setting `ELECTRON_RUN_AS_NODE=1`
+    // and pointing `command` at `process.execPath` makes the spawn behave
+    // exactly like invoking `node`. Costs zero extra installer size
+    // (Electron is already shipped) and unblocks fresh Windows / macOS /
+    // Linux installs that don't have Node.js on PATH. See #197.
+    //
+    // Substitution rules:
+    //   - Only triggers when running inside Electron (`process.versions.electron`).
+    //     Outside Electron (vitest, CI, mobile builds), behaviour is unchanged.
+    //   - Only matches the bare token `node` (or `node.exe` on Windows).
+    //     A user who specifies an absolute path (`/usr/local/bin/node`,
+    //     `/Users/me/.nvm/versions/node/...`) keeps their original choice.
+    //   - Does NOT touch `npx`, `npm`, `bun`, `deno`, etc. Those are separate
+    //     tools that the bundled Node runtime cannot stand in for.
+    if (
+      MCPConfigService.isBareNodeCommand(config.command) &&
+      typeof process !== 'undefined' &&
+      typeof process.versions === 'object' &&
+      typeof process.versions.electron === 'string' &&
+      process.versions.electron.length > 0
+    ) {
+      return {
+        ...config,
+        command: process.execPath,
+        env: {
+          ...(config.env || {}),
+          ELECTRON_RUN_AS_NODE: '1',
+        },
+      };
+    }
+
     // Resolve command for current platform
     const resolvedCommand = this.resolveCommandForPlatform(config.command);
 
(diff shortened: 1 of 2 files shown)


ademczuk:
@ghinkle scoped take on #197 - rather than ship a separate Node binary in the installer (30-50 MB per platform), this routes bare `command: "node"` MCP servers through Electron's existing Node runtime via `ELECTRON_RUN_AS_NODE=1`. Zero installer-size cost, addresses the OP's actual blocker. `npx` is left as out-of-scope follow-up since that needs a separate design call about managed package caching. Happy to revisit if you'd rather literally bundle a Node binary.

ghinkle:
Yea, this makes sense to use electron's run as node feature, but unfortunately this landed in dead code. MCPConfigBuilder was an incomplete refactor.  Sorry about that, will clean this up.

This fix needs to move into MCPConfigService.processServerConfigForRuntime

ademczuk:
@ghinkle moved the fix from `MCPConfigBuilder.processServerConfig` to `MCPConfigService.processServerConfigForRuntime` per your review. Force-pushed in [`d583b04`](https://github.com/nimbalyst/nimbalyst/pull/198/commits/d583b049). Same substitution logic, same 14 unit tests (relocated to `packages/electron/src/main/services/__tests__/MCPConfigService.bundledNode.test.ts`), but now in the live path that the four call sites in `src/main/index.ts` actually go through. Tests green locally on Windows.

Also: spotted that my earlier #194 squash-merge left literal merge conflict markers in `CHANGELOG.md` lines 19/21/24 on `main` (sorry, that's on me). Filed a tiny separate PR [#203](https://github.com/nimbalyst/nimbalyst/pull/203) to remove them and reorder the four `Unreleased > Fixed` entries cleanly. No code change in #203 - just markdown hygiene.

PR #200, opened 2026-05-07
fix(claude-provider): omit temperature for Opus 4.7+ to unbreak Test Connection

### LINKED ISSUE #199: Claude Chat fails with "temperature is deprecated" against default Opus 4.7 model
## Summary

Claude Chat unconditionally sends `temperature` on every API request. The default model is `claude-opus-4-7`, which Anthropic no longer accepts `temperature` for. The Test Connection button reports "Failed" for any valid Anthropic API key, and real chat messages return the same 400.

## Steps to reproduce

1. Open Settings, go to the Claude Chat panel.
2. Paste a valid Anthropic API key.
3. Click Test.

Same 400 also reproduces by skipping the Test button, leaving the default model, and sending a normal chat message.

## Expected behavior

Test reports success on a valid key. Real chat messages stream a response.

## Actual behavior

Test reports "Failed" with:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"`temperature` is deprecated for this model."},"request_id":"req_..."}
```

This looks like an API key problem to users. Multiple users have already reported it as "API key not being accepted."

Workaround: switch the default Claude Chat model to a Sonnet variant. Sonnet still accepts `temperature`, so Test passes and chat works.

## Where it lives in the code

`packages/runtime/src/ai/server/providers/ClaudeProvider.ts` around lines 284-291 builds the request:

```ts
const apiRequest: any = {
  model: modelId,
  max_tokens: this.config.maxTokens || 4000,
  temperature: this.config.temperature || 0,
  system: systemPrompt,
  messages: apiMessages,
  ...(tools.length > 0 ? { tools } : {}),
};
```

`temperature` is always present. The default model that triggers the error is set in `packages/runtime/src/ai/modelConstants.ts`:

```ts
export const DEFAULT_MODELS = {
  claude: 'claude:claude-opus-4-7',
  ...
};
```

The Test Connection handler at `packages/electron/src/main/services/ai/AIService.ts` around line 2787 picks that default and posts through the same provider, which is why every fresh install reports a misleading "Failed" on the Test button.

The Claude Code panel's Test button hits the same provider path, so it shows the same misleading "Failed" too.

## Environment

- App version: 0.59.2 (and earlier on this code path)
- OS: any (the bug is on the request side, not platform-specific)
- Anthropic API key tier: any (Console keys with Opus access)

## Logs

Not collected. The 400 above is the full error surfaced in the UI.

ademczuk:
## Summary

Fixes #199. Anthropic deprecated `temperature` for `claude-opus-4-7`; sending it returns HTTP 400 `"temperature is deprecated for this model"`. The default Claude Chat model on a fresh install is Opus 4.7, so every new user hit a misleading "Test Connection: Failed" against valid Anthropic API keys and blamed their key. Real chat sends failed the same way.

`ClaudeProvider` now omits `temperature` for Opus 4.7+ via a new `ClaudeProvider.supportsTemperature(modelId)` denylist. Older Opus (4 / 4.1 / 4.5 / 4.6), every Sonnet variant, every Haiku variant, and legacy Claude 3 models stay on the existing temperature path, mirroring the pattern `OpenAIProvider` already uses for `o1` / `gpt-5` / `gpt-4.5` (`OpenAIProvider.ts:262-269`).

## Root cause

`packages/runtime/src/ai/server/providers/ClaudeProvider.ts:287` was building the request unconditionally:

```ts
const apiRequest: any = {
  model: modelId,
  max_tokens: this.config.maxTokens || 4000,
  temperature: this.config.temperature || 0,  // <-- always sent
  system: systemPrompt,
  messages: apiMessages,
  ...(tools.length > 0 ? { tools } : {}),
};
```

The `|| 0` fallback meant `temperature` was always present in the API request, so the moment Anthropic flipped `claude-opus-4-7` to reject it the Test Connection button broke for every fresh install.

## Fix

```ts
// Build request without `temperature`...
const apiRequest: any = {
  model: modelId,
  max_tokens: this.config.maxTokens || 4000,
  system: systemPrompt,
  messages: apiMessages,
  ...(tools.length > 0 ? { tools } : {}),
};

// ...then add it back only when the model accepts it.
if (ClaudeProvider.supportsTemperature(modelId)) {
  apiRequest.temperature = this.config.temperature || 0;
}
```

The classifier is a small static method:

```ts
static supportsTemperature(modelId: string | undefined): boolean {
  if (!modelId || typeof modelId !== 'string') return true;
  const id = modelId.trim().toLowerCase();
  if (!id) return true;

  const opusMinor = id.match(/^claude-opus-4-(\d{1,2})(?:-|$)/);
  if (opusMinor) {
    const minor = parseInt(opusMinor[1], 10);
    return minor < 7;
  }

  return true;
}
```

The regex is anchored on `(?:-|$)` to avoid mis-parsing the 8-digit date suffix on Opus 4.0 (`claude-opus-4-20250514`) as a minor version. The `\d{1,2}` cap keeps two-digit minors (`claude-opus-4-10`) classified correctly without colliding with the date.

## Design choice: denylist over allowlist

Took both options to a multi-model consensus check. The trade-off is:

- **Denylist** (chosen): omit `temperature` for known-rejecting models, default to including for new models. If Anthropic deprecates `temperature` on a future model the denylist needs a follow-up PR, but the failure mode is a loud HTTP 400 with a clear message.
- **Allowlist**: include `temperature` only for known-good models, default to omitting for new models. Every new Claude release silently strips `temperature` from API calls until the allowlist is updated, regressing user-set values without any visible signal.

Picked denylist because the silent regression mode of an allowlist is the worse failure for users who set temperature deliberately, and because `OpenAIProvider` already uses the same pattern in this codebase.

## Tests

`packages/runtime/src/ai/server/providers/__tests__/ClaudeProvider.temperature.test.ts` adds 18 unit tests:

- Opus 4.7 / hypothetical 4.8 / 4.9 rejected
- Two-digit minors (4-10, 4-25) rejected
- Opus 4.0 (`claude-opus-4-20250514`) accepted - the 8-digit date suffix must not be parsed as a minor
- Opus 4.1 / 4.5 / 4.6 accepted (with and without date suffix)
- All Sonnet variants (3.7, 4, 4.5, 4.6) accepted
- Hypothetical `claude-sonnet-4-7` accepted (denylist scoped to Opus only)
- All Haiku variants accepted
- Legacy `claude-3-opus-20240229` accepted
- Case-insensitive matching
- Whitespace-trimmed input
- Malformed input (undefined, null, empty string) safely returns `true`

```
Test Files  1 passed (1)
     Tests  18 passed (18)
```

`npm run typecheck` passes. The two pre-existing test failures on Windows (`ClaudeCodeProvider.bashParser.test.ts`, `claudeCodeEnvironment.test.ts`) reproduce on `upstream/main` and are unrelated.

## Files

- `packages/runtime/src/ai/server/providers/ClaudeProvider.ts` - omit `temperature` for Opus 4.7+, add `supportsTemperature` static method, update the diagnostic `console.log` to render `(omitted - not supported by model)` instead of `undefined`
- `packages/runtime/src/ai/server/providers/__tests__/ClaudeProvider.temperature.test.ts` - 18 new tests
- `CHANGELOG.md` - entry under `[Unreleased] > Fixed`

## Related

- Reproducible by setting Claude Chat to Opus 4.7 (the default) and clicking Test in Settings, or by sending any chat message
- Workaround currently in use: switch the Claude Chat default model to a Sonnet variant
- The Claude Code panel's Test button hits the same provider path and was broken the same way; this fix covers it too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Claude Chat no longer fails Test Connection (and real chat sends) against the default `claude-opus-4-7` model. Anthropic deprecated `temperature` for Opus 4.7 and the provider was sending it unconditionally, so every fresh install hit `400 "temperature is deprecated for this model"` on the first API call and reported a misleading "Test Connection: Failed" against valid keys. `ClaudeProvider` now omits `temperature` for Opus 4.7+ via a new `ClaudeProvider.supportsTemperature(modelId)` denylist (matches `claude-opus-4-N` for N >= 7, while keeping Opus 4 / 4.1 / 4.5 / 4.6, every Sonnet variant, every Haiku variant, and legacy Claude 3 models on the existing temperature path). Mirrors the pattern `OpenAIProvider` already uses for `o1` / `gpt-5` / `gpt-4.5`. Adds 18 unit tests covering the Opus version boundary, the 8-digit date suffix on Opus 4.0 (`claude-opus-4-20250514`), case-insensitive matching, and malformed input. Fixes #199.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/runtime/src/ai/server/providers/ClaudeProvider.ts
+++ b/packages/runtime/src/ai/server/providers/ClaudeProvider.ts
@@ -284,11 +284,21 @@ export class ClaudeProvider extends BaseAIProvider {
       const apiRequest: any = {
         model: modelId,
         max_tokens: this.config.maxTokens || 4000,
-        temperature: this.config.temperature || 0,
         system: systemPrompt,
         messages: apiMessages,
         ...(tools.length > 0 ? { tools } : {}),
       };
+
+      // Anthropic deprecated `temperature` for `claude-opus-4-7` and is
+      // expected to do so on future reasoning Opus releases. Sending it
+      // returns HTTP 400 `"temperature is deprecated for this model"`,
+      // which surfaces in Nimbalyst as a misleading "Test Connection:
+      // Failed" against valid API keys -- users blame their key. The
+      // default model on a fresh install is Opus 4.7, so every new user
+      // hits this until they switch to Sonnet. See nimbalyst#199.
+      if (ClaudeProvider.supportsTemperature(modelId)) {
+        apiRequest.temperature = this.config.temperature || 0;
+      }
 
       // Apply response format if specified (extension chat completions)
       if (this.config.responseFormat && this.config.responseFormat.type !== 'text') {
@@ -305,7 +315,7 @@ export class ClaudeProvider extends BaseAIProvider {
       console.group('🚀 [ClaudeProvider] Final API Request');
       console.log('Model:', apiRequest.model);
       console.log('Max tokens:', apiRequest.max_tokens);
-      console.log('Temperature:', apiRequest.temperature);
+      console.log('Temperature:', 'temperature' in apiRequest ? apiRequest.temperature : '(omitted - not supported by model)');
       console.log('Has tools:', !!apiRequest.tools);
       console.log('Number of tools:', apiRequest.tools?.length || 0);
       console.log('System prompt length:', apiRequest.system.length);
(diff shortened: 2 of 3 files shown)


ademczuk:
Hey @ghinkle - small one for the queue. Closes #199 (Karl filed it earlier today). Pulls a fresh-install showstopper out of Claude Chat: every new user with the default Opus 4.7 model hits a 400 on Test Connection because we send `temperature` unconditionally, and they blame their API key.

Same pattern `OpenAIProvider` already uses for o1/gpt-5/gpt-4.5. 18 unit tests, single file change in the provider. Took the denylist vs allowlist trade-off to a multi-model consensus before settling on denylist (silent regression on an allowlist would be worse for users than a future loud 400 that we patch when it happens).

PR #210, opened 2026-05-08
fix(agent-view): initialize unified workstream state on drag-drop conversion

SheldonCurtiss:
## Summary

When a user drag-drops one session onto another to create a workstream, the unified workstream state for the new parent is not initialized. This produces inconsistent UI when more sessions are subsequently dragged into the workstream and the parent does not inherit the dragged session's UI settings (layout mode, split ratio, files sidebar visibility) the way the sibling-creation path does.

## Repro

1. In the agent-view sidebar, drag one session row onto another. A new workstream parent is created with the dropped session as its first child.
2. Drag a second session into the same workstream.
3. Inconsistent rendering of the parent workstream: layout, sidebar visibility, etc. fall back to defaults instead of the dragged session's settings; subsequent reparent operations on the parent operate against uninitialized fields.

## Root cause

In `packages/electron/src/renderer/store/atoms/sessions.ts:1420-1428`, the call to `convertToWorkstreamAtom` (in `workstreamState.ts`) was gated:

```ts
// Update unified workstream state (only when sibling was created)
if (siblingResult.success && siblingResult.sessionId) {
  const { convertToWorkstreamAtom } = await import('./workstreamState');
  set(convertToWorkstreamAtom, { sessionId, parentId: parentSessionId, siblingId: siblingResult.sessionId });
}
```

Drag-drop sets `skipSiblingCreation: true`, so `siblingResult.sessionId` is never produced and the `if` block was skipped. As a result `workstreamStateAtom(parentSessionId)` was never written — only `setWorkstreamActiveChildAtom` ran (which only patches `activeChildId`). The new parent ended up with default values for everything else, including:
- `childSessionIds: []` (default), so the type auto-detect on read stayed `'single'` until a later reparent populated children indirectly.
- All UI settings (`layoutMode`, `splitRatio`, `filesSidebarVisible`, etc.) at defaults instead of inherited from the dragged session.

The signature of `convertToWorkstreamAtom` in `workstreamState.ts` required `siblingId: string`, which is why the call site couldn't simply be moved out of the conditional.

## Fix

**`packages/electron/src/renderer/store/atoms/workstreamState.ts`**
- Make `siblingId` optional in the action's input type.
- Derive `childSessionIds`: `[sessionId, siblingId]` when a sibling exists, otherwise `[sessionId]`.
- Derive `activeChildId`: `siblingId ?? sessionId`.
- Skip the `set(workstreamStateAtom(siblingId), createDefaultState(siblingId))` call when `siblingId` is omitted (nothing to initialize).

The original-session-cleared block and the parent-state UI-inheritance block are unchanged so the existing sibling-creation path is preserved exactly.

**`packages/electron/src/renderer/store/atoms/sessions.ts`**
- Remove the `if (siblingResult.success && siblingResult.sessionId)` gate around the `convertToWorkstreamAtom` call so it runs in both paths.
- Spread `siblingId` into the action input only when present, so the no-sibling case stays clean.

`markSessionReadAtom` is intentionally left gated on sibling existence — there's no need to mark the original session as read since it already was.

**Tests** (new file: `packages/electron/src/renderer/store/atoms/__tests__/workstreamState.convertToWorkstream.test.ts`)
- Pins the with-sibling behavior (parent state + cleared original + sibling defaults + UI inheritance).
- Pins the new no-sibling drag-drop behavior (parent has only the original as a child, original session is the active child, original session state cleared, UI settings inherited).
- Follows the existing standalone-Jotai-store pattern from `fileTree.visibleNodes.test.ts`.

All 7 cases pass: `npx vitest run packages/electron/src/renderer/store/atoms/__tests__/workstreamState.convertToWorkstream.test.ts`.

## Test plan

- [x] `electron-vite build` succeeds.
- [x] New unit tests pass (7/7).
- [x] Manual test: drag-create a workstream from a session → drag more sessions into it → child list visually consistent.
- [x] Existing sibling-creation conversion path unchanged.

## Related

Companion to #208 (fixes the rename live-update + adds Rename to workstream parent menu). That PR's "out of scope" note pointed at exactly this code path. PRs are independent and can be merged in either order.

### DIFF
--- a/packages/electron/src/renderer/store/atoms/__tests__/workstreamState.convertToWorkstream.test.ts
+++ b/packages/electron/src/renderer/store/atoms/__tests__/workstreamState.convertToWorkstream.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { createStore } from 'jotai';
+import {
+  convertToWorkstreamAtom,
+  initWorkstreamState,
+  workstreamStateAtom,
+} from '../workstreamState';
+
+/**
+ * Unit tests for convertToWorkstreamAtom covering both:
+ *   - the sibling-creation path (existing behavior)
+ *   - the drag-drop path (new: siblingId omitted)
+ *
+ * The drag-drop path was previously gated out at the call site in sessions.ts,
+ * so workstreamStateAtom(parentId) was never initialized. This test pins the
+ * fully-initialized behavior so future refactors can't silently regress it.
+ */
+
+describe('convertToWorkstreamAtom', () => {
+  let store: ReturnType<typeof createStore>;
+  const sessionId = 'session-original';
+  const parentId = 'parent-workstream';
+  const siblingId = 'session-sibling';
+
+  beforeEach(() => {
+    store = createStore();
+    // Required by workstreamStateAtom's writer (schedulePersist asserts a workspace path is set).
+    initWorkstreamState('/test-workspace');
+  });
+
+  describe('with siblingId (normal conversion)', () => {
+    it('initializes parent state with both children and sibling as active', () => {
+      store.set(convertToWorkstreamAtom, { sessionId, parentId, siblingId });
+
+      const parent = store.get(workstreamStateAtom(parentId));
+      expect(parent.type).toBe('workstream');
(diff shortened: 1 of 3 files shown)


SheldonCurtiss:
Closing — reopening from GigaBigBoss fork. Will link the replacement PR.

PR #221, opened 2026-05-09
feat(mcp/sessions): add searchField=title to list_recent_sessions

### LINKED ISSUE #83: Feature request: list_recent_sessions should support title-only search
## Description

`mcp__nimbalyst-session-context__list_recent_sessions` accepts a `search` parameter, but it matches on session **content** (the conversation transcript), not just the session title.

This makes it hard to find a specific session by name when the search term appears frequently in conversations.

## Example

Searching for `"Budget"`:
- **Expected:** the session titled "Budget tracker setup"
- **Actual:** every session that mentions "Budget" anywhere in its conversation — planning sessions, sessions where budgets came up in passing, etc.

## Request

Add a `searchField` (or `titleOnly: true`) parameter to `list_recent_sessions`:

```
list_recent_sessions({ search: "Budget", searchField: "title" })
// or
list_recent_sessions({ search: "Budget", titleOnly: true })
```

Default behavior can stay as-is for backward compatibility.

## Environment
Nimbalyst v0.57.33, macOS.

ademczuk:
## Summary

Fixes #83. Reporter wants `mcp__nimbalyst-session-context__list_recent_sessions` to support a title-only search mode so models can find a session by name without getting drowned in content matches.

## Root cause

The current `query` parameter goes through `AISessionsRepository.search`, which in `PGLiteSessionStore` runs two FTS queries and unions the results: one against `s.title` and one against `ai_transcript_events.searchable_text`. So searching `"Budget"` returns:
- the session titled "Budget tracker setup" (good)
- every session that mentioned a budget anywhere in its conversation (noise)

There was no way to scope the query to titles only.

## Fix

Adds a `searchField` parameter to the tool schema with values `'title' | 'content' | 'both'`. Default is `'both'`, which preserves the existing behavior.

When `searchField: 'title'` is passed:

```ts
const all = await AISessionsRepository.list(workspaceId, options);
const needle = trimmedQuery.toLowerCase();
sessions = all.filter((s) =>
  (s.title ?? "").toLowerCase().includes(needle)
);
```

The handler calls `list()` (which is what the no-query path already calls) and filters by case-insensitive substring on `s.title`. The FTS path is bypassed entirely. Pagination via `slice(offset, offset + limit)` happens after the filter, same as before.

The result label and no-results message both append `in titles` when the title-only mode is active so the model can tell the modes apart in the output:

```
Recent sessions (showing 1 of 1 matching "Budget" in titles):
1. "Budget tracker setup" (uuid) - 2 hours ago
   Provider: claude-code
```

## Why a JS substring filter and not a SQL title-only branch

The store's existing title query at `PGLiteSessionStore.ts:644-678` uses `plainto_tsquery('english', ...)` which stems and tokenizes. For "find by name" the user wants substring fidelity, not stemming -- searching `"Budg"` should match `"Budget tracker setup"`. Adding a separate parametrized SQL path would mean either teaching the SessionStore interface a new flag or introducing an `ILIKE` branch that has different semantics from the main FTS path. Filtering in JS keeps the change scoped to the MCP handler, makes the semantics obvious, and matches how `SessionManager.test.ts:95-110` already mocks `search` for tests.

## Class-of-bug check

Grepped `AISessionsRepository.search` across the MCP package -- only `handleListRecentSessions` calls it. `VoiceModeService.ts:586` also calls it, but that's the voice-agent search-recent-sessions tool, which is a different surface and not what the reporter raised.

## Test plan

- [ ] Manual: `list_recent_sessions({ query: "Budget" })` -- existing FTS behavior, matches title and content
- [ ] Manual: `list_recent_sessions({ query: "Budget", searchField: "title" })` -- only sessions whose title contains "Budget"
- [ ] Manual: `list_recent_sessions({ searchField: "title" })` with no query -- behaves like list-all (the searchField is ignored when query is empty, per the description)
- [ ] Manual: invalid `searchField` value falls back to `'both'`

## Closes

Fixes #83.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-<!-- New features go here -->
+- Title-only mode for `mcp__nimbalyst-session-context__list_recent_sessions`. The existing `query` parameter goes through `AISessionsRepository.search`, which uses Postgres full-text search across both session titles and `ai_transcript_events.searchable_text`. That made it hard to find a session by name when the search term appeared frequently in conversations (e.g. searching `"Budget"` returned every session that mentioned budgets in passing, not just the session titled "Budget tracker setup"). Adds a `searchField` parameter accepting `'title'`, `'content'`, or `'both'` (default `'both'` keeps existing behavior). When `searchField: 'title'` is passed, the handler calls `AISessionsRepository.list` and filters in memory by case-insensitive substring match on `s.title`, bypassing the FTS path entirely. The result label and no-results message both show `in titles` when the title-only mode is active so the model can tell the modes apart. Fixes #83.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/electron/src/main/mcp/sessionContextServer.ts
+++ b/packages/electron/src/main/mcp/sessionContextServer.ts
@@ -395,13 +395,26 @@ async function handleListRecentSessions(
   offset: number,
   workspaceId: string,
   currentSessionId: string,
-  includeArchived: boolean
+  includeArchived: boolean,
+  searchField: "title" | "content" | "both" = "both"
 ): Promise<string> {
   let sessions: SessionMeta[];
 
   const options = { includeArchived };
-  if (query && query.trim().length > 0) {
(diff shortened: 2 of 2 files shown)


ademczuk:
Hey @ghinkle, small MCP-handler-only change for #83 if you have a minute. Fork PR so I can't request review formally.

PR #222, opened 2026-05-09
fix(transcript): add chatShowToolCalls user setting to honour issue #118

### LINKED ISSUE #118: Add option to fully hide tool calls from chat view (not just collapse them)
## Problem
When using Claude Code in Nimbalyst, the chat view fills with tool call rows (Read, Bash, Grep, ToolSearch, TodoWrite, Skill, etc.) — even with the `showToolCalls: false` setting in `ai-settings.json`, the rows still appear in collapsed form. For users who only care about the conversation between themselves and the agent, this clutters the view and makes it hard to scroll back through the actual dialogue.

## Proposed solution
A toggle (toolbar button or keybind) to fully hide tool call rows in the chat view, leaving only user messages and assistant text responses. Could be:
- A view-level toggle in the chat panel header (eye icon, or "compact / conversation-only" mode)
- A keybind such as `Ctrl+Shift+T`
- Or honoring the existing `showToolCalls: false` setting more strictly so rows are removed from the DOM, not just collapsed

## Alternatives considered
- The `showToolCalls: false` setting in `ai-settings.json` is already off but rows still render collapsed.
- Manually scrolling past them — works for short sessions, breaks down on long ones with many tool calls.

## Additional context
Use case: non-technical operators who use Nimbalyst as a chat interface and don't need to see the underlying tool plumbing. In a typical session, 25+ collapsed tool-call rows can separate two short conversational messages, making it hard to find the actual dialogue.

ademczuk:
## Summary

Fixes #118. Reporter on the issue set `showToolCalls: false` in `ai-settings.json` expecting tool-call rows to disappear from the chat view, but they kept rendering as collapsed bars.

## Why the original approach (single-flag) was wrong

The existing `showToolCalls` field is **dev-mode-only**:
- Default `false` in the electron-store schema (with comment "Hidden by default, developer mode only").
- The UI toggle in `AgentFeaturesPanel` is gated behind an `{isDevelopment && ...}` block.

Naively making the chat view honour `showToolCalls` from the atom (as the original draft of this PR did) would silently hide tool rows for **every** production user after upgrade, since their stored value is the default `false`. The reporter who manually set `false` is a power user; the rest of the user base never sees the toggle.

I caught this on a second-pass review, moved the PR back to draft, and reworked.

## What this PR does (approach A)

Adds a separate user-facing `chatShowToolCalls: boolean` (default `true`) that the chat transcript routes through. The dev-only `showToolCalls` field is untouched.

**Main-process schema and IPC** (`packages/electron/src/main/services/ai/AIService.ts`):
- New `chatShowToolCalls` boolean in the electron-store schema with `default: true`.
- Parallel reads in `ai:getSettings` and `ai:getEffectiveSettings`, parallel write block in `ai:saveSettings`.

**Renderer atom** (`packages/electron/src/renderer/store/atoms/appSettings.ts`):
- `AIDebugSettings` interface gains `chatShowToolCalls: boolean`.
- Default `true` in `defaultAIDebugSettings` and `initAIDebugSettings`.
- Persisted via the existing debounced `scheduleAIDebugPersist`.
- New `chatShowToolCallsAtom` slice export, peer to `showToolCallsAtom`.

**UI toggle** (`packages/electron/src/renderer/components/Settings/AgentFeaturesPanel.tsx`):
- "Show Tool Calls in Chat" toggle placed in the always-visible Agent Features section (NOT in the `isDevelopment` block).
- The existing dev-only "Show All Tool Calls" toggle is unchanged and stays in `Developer Options`.

**Chat transcript routing** (`packages/electron/src/renderer/components/UnifiedAI/SessionTranscript.tsx`):
- Reads `chatShowToolCallsAtom` and passes it into `initialSettings.showToolCalls` (replaces the hardcoded `true`).

**Renderer guards** (`packages/runtime/src/ui/AgentTranscript/components/RichTranscriptView.tsx`):
- Two render paths previously called `renderToolCard` unconditionally:
  1. Orphan tool cards (tool ran but no following assistant message yet).
  2. `toolMessagesBefore` cards rendered alongside the next assistant turn.
- Both now check `settings.showToolCalls` and skip non-interactive tools when off.
- Interactive widgets (`ToolPermission`, `ExitPlanMode`, `AskUserQuestion`, `GitCommitProposal`) always render so the user can act on prompts even when tool rows are hidden.
- `INTERACTIVE_WIDGET_TOOLS` is lifted from an inline array (line 1848 on main) to a module-scope `Set`, reused at all three guard sites.

## Behaviour table

| User state | Before | After |
|---|---|---|
| Hasn't touched setting (default) | sees tool rows | sees tool rows (default `true`) |
| Manually set `chatShowToolCalls: false` in `ai-settings.json` (reporter on #118) | sees tool rows | tool rows hidden, interactive widgets still shown |
| Has dev mode on + toggles dev-only `showToolCalls` | unchanged | unchanged |

## Test plan

- [x] `tsc --noEmit` clean across `packages/runtime` + `packages/electron` (3 pre-existing `@dnd-kit/*` module-resolution errors are unrelated to this PR; same on `upstream/main`).
- [x] `npm run test:unit -- --run packages/runtime/src/ui/AgentTranscript/components/__tests__/TranscriptWidgets.test.tsx`: 57/57 pass.
- [ ] Manual: fresh install, agent session with tool calls visible by default.
- [ ] Manual: toggle off in Agent Features, agent session shows no tool rows but `ToolPermission`/`ExitPlanMode`/`AskUserQuestion`/`GitCommitProposal` widgets still render and accept input.
- [ ] Manual: set `chatShowToolCalls: false` directly in `~/.claude/ai-settings.json`, confirm matches the in-app toggle off.

## Notes

- Default `true` preserves UX for every existing user who hasn't touched the setting. The migration safety pattern (`?? true` in `initAIDebugSettings`) handles users whose stored settings predate this field.
- The original draft's "honour the existing dev-only field" approach was wrong. Walked back, reworked. Greg, I commented this rationale on the original draft (Sat 09:18Z) and proposed approach A; absent a different opinion this is what shipped.

Closes #118.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-<!-- New features go here -->
+- Honour a new user-facing `chatShowToolCalls` setting in `ai-settings.json` (default `true`) so the AI chat view can hide tool-call rows entirely instead of just collapsing them. The original `showToolCalls` field stays a developer-mode-only toggle (default `false`, gated behind `isDevelopment` in `AgentFeaturesPanel`), so production users get a discoverable "Show Tool Calls in Chat" toggle in Agent Features that defaults to on. The renderer's tool-row guards in `RichTranscriptView` (orphan-tool path and `toolMessagesBefore` path) consume the new value flowing through `initialSettings.showToolCalls`. Interactive tool widgets (`ToolPermission`, `ExitPlanMode`, `AskUserQuestion`, `GitCommitProposal`) always render so the user can still act on prompts even when tool rows are hidden. Default `true` preserves UX for everyone who has not manually set the value. Fixes #118.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/electron/src/main/services/ai/AIService.ts
+++ b/packages/electron/src/main/services/ai/AIService.ts
@@ -429,6 +429,10 @@ export class AIService {
             type: 'boolean',
             default: false  // Hidden by default, developer mode only
           },
+          chatShowToolCalls: {
+            type: 'boolean',
+            default: true  // User-facing chat toggle; defaults true to preserve current UX
+          },
           aiDebugLogging: {
             type: 'boolean',
             default: false  // Hidden by default, developer mode only
@@ -2570,6 +2574,7 @@ export class AIService {
       const apiKeys = this.getSettingsStore().get('apiKeys', {}) as Record<string, string>;
       const providerSettings = this.getNormalizedProviderSettings();
       const showToolCalls = this.getSettingsStore().get('showToolCalls', false) as boolean;
+      const chatShowToolCalls = this.getSettingsStore().get('chatShowToolCalls', true) as boolean;
       const aiDebugLogging = this.getSettingsStore().get('aiDebugLogging', false) as boolean;
       const showPromptAdditions = this.getSettingsStore().get('showPromptAdditions', false) as boolean;
       const showUsageIndicator = this.getSettingsStore().get('showUsageIndicator', true) as boolean;
@@ -2592,6 +2597,7 @@ export class AIService {
         apiKeys: this.maskApiKeys(apiKeys),
         providerSettings,
         showToolCalls,
+        chatShowToolCalls,
         aiDebugLogging,
         showPromptAdditions,
         showUsageIndicator,
@@ -2689,6 +2695,10 @@ export class AIService {
 
       if (settings.showToolCalls !== undefined) {
         this.getSettingsStore().set('showToolCalls', settings.showToolCalls);
+      }
+
+      if (settings.chatShowToolCalls !== undefined) {
+        this.getSettingsStore().set('chatShowToolCalls', settings.chatShowToolCalls);
       }
 
       if (settings.aiDebugLogging !== undefined) {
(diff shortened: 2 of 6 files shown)


ademczuk:
Hey @ghinkle, the chat view fix for #118. Two render paths in RichTranscriptView were calling renderToolCard unconditionally, plus SessionTranscript hardcoded showToolCalls=true. Settings panel checkbox now actually does what it says.

ademczuk:
@ghinkle moving this to draft. Self-caught a regression on second-pass review before you merged.

## What I missed

`showToolCalls` is gated as dev-mode-only by `AgentFeaturesPanel.tsx:229` (the toggle lives inside an `isDevelopment` block). Schema at `AIService.ts:428-431` has `default: false` with the comment "Hidden by default, developer mode only". Production users never see this toggle, never set the value, and `getSettingsStore().get('showToolCalls', false)` returns false for all of them.

Today the chat view hardcodes `initialSettings.showToolCalls: true`, which is why production users still see tool rows even though their stored value is false. My PR makes the chat view honor that hidden default-false, which means **every production user would silently lose tool-call rows after upgrade**. That's the opposite of a bug fix.

The reporter on #118 is a power user who manually wrote `showToolCalls: false` into `ai-settings.json`. Their explicit intent should be honored. Production users who never touched the setting should not have their chat silently transformed.

## Proposed re-do

Two approaches, both safer than what I shipped:

**A. New chat-facing setting.** Add a separate `chatShowToolCalls: boolean` defaulted to `true` in both main (`AIService.ts`) and renderer (`appSettings.ts`), expose a non-dev-mode toggle in a sensible Settings panel, and route the chat view through that. Leave the existing `showToolCalls` dev-only field alone for what it actually means in your model. Reporter's `chatShowToolCalls: false` in ai-settings.json works, default-true preserves current UX for everyone else.

**B. Dev-mode gate in the chat view.** Honor `showToolCalls` from the renderer atom only when `isDevelopment` is true. Production users keep current UX (always show), dev users get the working toggle. Reporter on #118 would need to be in dev mode for their explicit-false to take effect, which probably isn't what they want.

I think A is the right answer. I'll rework the PR against A unless you have a different opinion.

The renderer changes in `RichTranscriptView` (orphan path + `toolMessagesBefore` guard, `INTERACTIVE_WIDGET_TOOLS` lifted to module scope) are still useful regardless of which setting drives them, so I'll keep that part.

PR #223, opened 2026-05-09
fix(auto-updater): suppress transient network errors on background polls

### LINKED ISSUE #56: Update error, can it be disabled?
Windows version.
I constantly get this error even if I am connected to the network via LAN cable.
Can the update check be disabled somehow? There is no entry in the settings.

Update Error
net::ERR_NAME_NOT_RESOLVED
You can [download the latest version manually](https://nimbalyst.com/download).
Dismiss

ademczuk:
## Summary

Fixes #56. Reporter on Windows is behind a network that cannot resolve the GitHub release feed. The auto-updater polls every hour, every poll fails with `net::ERR_NAME_NOT_RESOLVED`, every failure fires `update-toast:error`, and there is no setting to opt out. The same pattern hits anyone on a captive portal, restrictive corporate firewall, or LAN-only network.

## Root cause

`AutoUpdaterService.setupEventHandlers()` at `packages/electron/src/main/services/autoUpdater.ts:158-196` fires `update-toast:error` to the frontmost window for every `error` event from `electron-updater`, regardless of:
- the stage (check vs download)
- the cause (network, permission, signature, disk)
- whether the check was user-initiated or the hourly auto-poll

The codebase already has a `classifyUpdateError(err)` helper at line 33 that returns `'network' | 'permission' | 'disk_space' | 'signature' | 'unknown'`. It is currently only used for analytics. This PR uses the same classifier to gate the toast.

## Fix

In the existing `error` handler, capture `wasManualCheck` before the reset (the existing code resets the flag without remembering its prior value), then suppress `update-toast:error` when all three are true:

```ts
const isTransientNetworkCheckError =
  stage === 'check' && errorType === 'network' && !wasManualCheck;
if (!isTransientNetworkCheckError) {
  this.sendToFrontmostWindow('update-toast:error', { message: err.message });
}
```

What is preserved exactly as before:
- `log.error('Update error:', err)` still fires
- `AnalyticsService.sendEvent('update_error', ...)` still fires with the same shape
- `this.sendToAllWindows('update-error', err.message)` still fires (any renderer subscriber listening on `update-error` continues to receive it)
- Manual checks via `checkForUpdatesWithUI` set `isManualCheck = true` and still see the toast on failure
- Download errors (any stage that is not `'check'`) still see the toast
- The Windows EPERM rename retry path at `autoUpdater.ts:167` is reached before the toast logic and is unchanged
- The dev-mode-skip toast at line 572 (which emits `update-toast:error` with a literal "Update checking is not available in development mode") is in a different code path (`checkForUpdatesWithUI`, manual flow), unchanged

## Why not add a hide-update-errors setting

Considered. The reporter explicitly asked "Can the update check be disabled somehow?" - meaning they want to opt out. A setting toggle would address that ask but introduces a real footgun: a user toggling it off would also miss download failures, signature failures, and disk-space failures, which are real problems they should hear about. Suppressing only the transient-network-check class targets the actual pain (recurring noise on a network that cannot reach the feed) without hiding errors that warrant action. Open to adding a setting if you'd prefer that direction.

## Class-of-bug check

`grep -rn "update-toast:error" packages/electron/src` finds three sites:
- `autoUpdater.ts:191` - the auto-poll error handler (this PR's site)
- `autoUpdater.ts:434` - download retry failure (different stage, intentionally surfaces)
- `autoUpdater.ts:572,592` - manual-check-flow errors (intentionally surfaces, user clicked)

Only the first one was the recurring-noise source.

## Test plan

- [ ] Manual: simulate offline state (disconnect Wi-Fi), let auto-poll fire, no toast appears, debug log still records `Update error: ...` and analytics still records `update_error`
- [ ] Manual: click `Check for Updates` while offline, error toast appears (manual check flow preserved)
- [ ] Manual: simulate signature failure (e.g. tamper with installer), error toast appears (non-network errors still surface)
- [ ] Manual: download error mid-flight, toast appears (download stage still surfaces)

## Closes

Fixes #56.


ademczuk:
Hey @ghinkle, small one for #56. Reuses the existing `classifyUpdateError` helper so it is a single condition added in the error handler. Manual checks and download errors still surface normally.

Issue #207, opened 2026-05-08
Workspace file watcher does not honor .gitignore files of nested git repos, causing main-process OOM in non-git workspace roots


woytowg:
## Summary

In a non-git workspace root that contains nested git repos (the same superproject layout described in #122), Nimbalyst's workspace file watcher does not honor the `.gitignore` files of the nested repos. As a result, opening such a workspace can flood the watcher with events from build-output trees that the nested repo already ignores, trip the workspace event-bus circuit breaker, and then crash the Electron main process with a V8 "JavaScript heap out of memory" error in a tight restart loop. I have not been able to test a build with the fix from bug #122, so this may be resolved already.

## Steps to reproduce

1. Use a workspace whose root is **not** a git repo but contains nested git repos in subdirectories (the same layout as #122).
2. In one of the nested repos, have a sizable build-output directory (in my case roughly 4.7 GB / 100k+ files — an unpacked Linux rootfs) that is correctly listed in that nested repo's `.gitignore` (e.g. `/rootfs` in `<nested-repo>/.gitignore`).
3. Open the workspace root in Nimbalyst.

## Expected behavior

The workspace file watcher should treat the nested repo's `.gitignore` as authoritative for paths inside that nested repo, and skip the ignored build-output directory. Equivalently: the watcher should discover nested `.git`/`.gitignore` pairs the same way `git status` would, rather than only looking at the workspace root.

## Actual behavior

The watcher walks into the ignored build-output directory anyway. Within seconds, the workspace event bus trips its circuit breaker, the main process climbs to ~3.6–3.7 GB, and Electron crashes with `OOM error in V8: ... JavaScript heap out of memory`. On restart, the same workspace is restored from session state and the cycle repeats indefinitely.

## Evidence (from `~/.config/@nimbalyst/electron/logs/main.log` and `~/.xsession-errors`)

```
[WorkspaceEventBus] Circuit breaker tripped for "<WORKSPACE>" —
  received 5000 events in 5000ms. Killing watcher to protect the process.
  This workspace may be too large or missing a .gitignore.
```

Followed by a flood of:

```
Error reading folder contents: Error: EACCES: permission denied,
  scandir '<WORKSPACE>/<nested-repo>/rootfs/etc/credstore'
... (many more under <nested-repo>/rootfs/etc, /home, /lib, /persist, /root)
```

Then in `~/.xsession-errors`:

```
<--- Last few GCs --->
[...]

PR #226, opened 2026-05-09
fix(file-tree): move all selected items on multi-file drag, not just one

### LINKED ISSUE #31: Dragging multiple selected files only moves one file at a time
## Description

When multiple files are selected in the editor view and the user attempts to move them by dragging to a different folder, only one file is moved per drag operation. The rest of the selected files remain in their original location. To move all selected files, the user must repeat the drag-and-drop process individually for each remaining file, which is tedious and error-prone.

## Steps to reproduce

1. Open the editor view.
2. Select two or more files using Ctrl+Click or Shift+Click.
3. Drag the selection to a different folder.
4. Observe the result, only one file is moved to the target folder.
5. The remaining selected files stay in their original location.
6. Repeat the drag operation to move the next file and so on.

## Expected Behavior

All selected files should be moved to the target folder in a single drag and drop operation. The multi selection state should be respected and applied to the move action as a batch.

## Actual Behavior

Only one file is moved per drag operation, regardless of how many files are selected. The drag handler appears to process only a single file from the selection, ignoring the rest.

ademczuk:
## Summary

Fixes #31. Reporter selects N files via Ctrl-Click in the workspace tree, drags them to a target folder, and only the file they actually grabbed moves. The other N-1 files stay in place.

## Root cause

Two bugs in series in `FlatFileTree.tsx`:

**1. `handleDragStart` ignored the multi-selection.** Line 476 hardcoded `sourcePaths: [node.path]`. Even when the user had Ctrl-Click selected 5 files and dragged one of them, the drag state only carried that one path.

**2. `handleDrop` only read the first path.** Line 610 had `const sourcePath = dragState.sourcePaths[0]`. So even if I fixed `handleDragStart` to populate the array, drop only ever moved `[0]` and dropped the rest on the floor.

Either bug alone would have produced "only one file moves". Both together made the bug fully invisible to single-bug-fix attempts.

## Fix

**`handleDragStart`**:

```ts
const sourcePaths = selectedPaths.has(node.path) && selectedPaths.size > 1
  ? Array.from(selectedPaths)
  : [node.path];
```

When the dragged item is part of the multi-selection: drag the whole selection. When the user grabs an unselected item (clicked once, then dragged): drag only that item. This matches macOS Finder and Windows Explorer behavior. The drag-preview text shows `N items` instead of the filename when N > 1.

**`handleDrop`**:

Iterates `sourcePaths`, copies/moves each, collects per-item failures, refreshes the tree once if any succeeded, and surfaces a single summary `ErrorDialog` for partial failures. Mirrors the pattern `handleDeleteMultiple` already uses (shipped in #216, with the matching reasoning that one error dialog per failed file is hostile UX).

```ts
const sourcePaths = dragState.sourcePaths.filter((p) => p && p !== node.path);
// ...iterate, collect failures, single summary dialog
```

The dialog title flips between `Move failed` (all failed), `Some files could not be moved` (partial), and the verb swaps to `copied` when Alt/Cmd is held during drop. Same per-item failure-summary format `handleDeleteMultiple` uses.

## What stays the same

- Single-item drag (the user grabs an unselected file) is unchanged.
- `handleDragOver` / `handleDragEnter` / `handleDragLeave` are unchanged. The existing child-folder-of-source guard at `dragState.sourcePaths.some(src => ...)` already iterates the array, so it was already correct for the multi-source case.
- External drag from Finder (the `e.dataTransfer.types.includes('Files')` branch) is unchanged.
- Copy-on-Alt/Cmd modifier is unchanged.
- Tree refresh on success is unchanged.

## Class-of-bug check

`grep -n "sourcePaths\[" packages/electron/src/renderer/components/FlatFileTree.tsx` finds zero remaining hardcoded-index reads. Other consumers of `dragState.sourcePaths` already use `.some()` / `.includes()` (the over-folder validation), so they were already array-correct. Only the drag-start setter and the drop reader were buggy.

## Test plan

- [ ] Manual: select 3 files via Ctrl-Click, drag one of them into a folder, all 3 move
- [ ] Manual: select 3 files, hold Alt while dragging, all 3 copy (originals remain)
- [ ] Manual: drag an unselected file (no Ctrl-Click prior), only that file moves
- [ ] Manual: drag a multi-selection that includes the destination folder, the destination is filtered out (existing single-item behavior preserved for that edge case)
- [ ] Manual: induce a failure on one file mid-batch (e.g. permission denied on one of the selected files), the others still move and a single summary dialog lists the failure

## Closes

Fixes #31.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Drag-and-drop on a multi-file selection now moves or copies every selected item, not just one. `handleDragStart` in `FlatFileTree.tsx` was setting `sourcePaths: [node.path]` regardless of the current `selectedPaths` set, and `handleDrop` was reading only `sourcePaths[0]`, so dragging a Ctrl-Click selection of N files always silently dropped N-1 of them. `handleDragStart` now uses the full selection when the dragged item is part of it (and falls back to single-item drag when the user grabs an unselected item, matching Finder/Explorer behavior). `handleDrop` iterates every dragged path, collects per-item failures, refreshes the tree if any succeeded, and surfaces a single summary `ErrorDialog` for partial failures (mirroring the multi-delete pattern from #216). The drag preview shows `N items` for multi-selection drags. Fixes #31.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/renderer/components/FlatFileTree.tsx
+++ b/packages/electron/src/renderer/components/FlatFileTree.tsx
@@ -468,19 +468,31 @@ export function FlatFileTree({
       return;
     }
 
+    // When the user drags an item that is part of the current multi-selection,
+    // include every selected path so the drop handler moves/copies them all
+    // in a single operation. When the dragged item is NOT in the selection
+    // (the user grabbed an unselected file/folder), drag only that item -
+    // matches the typical macOS Finder / Windows Explorer behaviour where
+    // dragging an unselected item ignores any prior selection. See #31.
+    const sourcePaths = selectedPaths.has(node.path) && selectedPaths.size > 1
+      ? Array.from(selectedPaths)
+      : [node.path];
+
     e.dataTransfer.effectAllowed = 'copyMove';
     e.dataTransfer.setData('text/plain', node.path);
     // Also set custom MIME type so AI input can accept this as an @-file mention
     e.dataTransfer.setData('application/x-nimbalyst-file-mention', node.path);
     setDragState({
-      sourcePaths: [node.path],
+      sourcePaths,
       dropTargetPath: null,
       isCopy: false,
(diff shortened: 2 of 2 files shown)


ademczuk:
Hey @ghinkle, two bugs in series for #31. Both `handleDragStart` (hardcoded `[node.path]`) and `handleDrop` (hardcoded `sourcePaths[0]`) needed fixing for multi-file drag to work. Single-item drag and external Finder drag both unchanged. Same multi-failure summary-dialog pattern as `handleDeleteMultiple` from #216.

ademczuk:
Heads up — the failing CI here is inherited from main, not my change. `WorkspaceEventBus-nested-gitignore.test.ts` (added in `5c0cf068` for #207) has 3 failing tests with `Error: No watcher callback registered` plus one with a `gitignoreBypassed` arg mismatch. Reproduced locally on a clean main checkout: 3 failed / 1 passed. My change touches `FlatFileTree.tsx` only and is unrelated to the WorkspaceEventBus.

Happy to take a swing at fixing the test isolation if it's still on the to-do list, or leave it for whoever's actively in that file. Mention if you'd like a separate PR for it.

PR #227, opened 2026-05-09
feat(tracker): allow tracker_update to change the primary type of an item

### LINKED ISSUE #79: Feature request: Tracker primary type should be changeable after creation
The primary type of a tracker item (`task`, `idea`, `plan`, `bug`) is set at creation time via `tracker_create` and is immutable thereafter.

## What's broken

- `tracker_update` accepts `typeTags`, but those are added as **secondary tags** — they don't replace the primary type.
- There's no UI mechanism to change the primary type either.
- The only workaround is to **create a new item with the correct type and archive the old one**, which loses comments, attachments, and session links.

## Use case

Tracker items often start as one type and evolve. An `idea` becomes a concrete `task`. A `task` turns out to be a `bug`. A loose `task` grows into a `plan`. Today, that natural evolution is blocked unless you're willing to abandon the existing item's history.

## Request

Either:
1. Allow `tracker_update` to replace the primary type (e.g. `primaryType: "task"`), or
2. Add a dedicated `tracker_change_type(trackerId, newType)` tool, **and**
3. Surface the same in the UI — a type selector on the tracker item that allows changing.

## Environment
Nimbalyst v0.57.33, macOS.

ademczuk:
## Summary

Fixes #79. Tracker items often start as one type (idea) and evolve into another (task / bug / plan). `tracker_update` previously only accepted `typeTags` as a secondary-tag list. The primary `type` column was set at creation and immutable. The only workaround was archive-and-recreate, which loses comments, attachments, and session links.

## Fix

Adds a `primaryType` string parameter to the `tracker_update` MCP tool:

```ts
mcp__nimbalyst-tracker__tracker_update({
  id: "...",
  primaryType: "bug"
})
```

The handler:

1. **Validates** the new type against `globalRegistry.has(newType)`. Same registry that `tracker_create` validates against. Unknown types return a clear error.
2. **Persists** `tracker_items.type` via its own UPDATE before the existing data write, so the in-memory `row.type` and subsequent code (typeTags rebuild, activity log, sync) see the new value.
3. **Re-validates** `data` against the new type's schema. If the new type has a required field the old item lacks, the error surfaces as a normal validation error rather than silently writing a half-typed item.
4. **Logs activity** with action `type_changed` so the change appears in the item's history.
5. **Rebuilds `type_tags`**: the new primary type goes at the front, the old primary type is removed (it was leaking as a secondary tag with the previous code path), other secondary tags are preserved in order.
6. **Respects explicit typeTags**: if the user passes both `primaryType` and `typeTags`, the explicit `typeTags` array wins, consistent with the existing "replace the array" semantic.

## What stays the same

- `tracker_create`, validation rules, sync flow, activity-log shape - unchanged
- Item ID, comments, attachments, session links - all preserved (only the `type` column changes)
- `typeTags` semantics when `primaryType` is not provided - unchanged
- All 21 existing `trackerToolHandlers.test.ts` tests pass

## What is intentionally NOT in this PR

The reporter asks for the same UI surface ("a type selector on the item that allows changing"). That requires UI design - where the selector lives in the tracker detail panel, whether it confirms before changing, etc. - and is a separate ticket. This PR closes the MCP-side half so an agent (or a power user with MCP access) can flip the type today without losing history. Happy to follow up on the UI surface in a separate PR once @ghinkle picks the placement.

## Test plan

- [x] `npx vitest run packages/electron/src/main/mcp/tools/__tests__/trackerToolHandlers.test.ts` - 21/21 pass
- [ ] Manual via MCP: `tracker_update({ id: X, primaryType: "bug" })` on an existing `task`. The item now shows in `bug` view, history is intact, activity log records the change.
- [ ] Manual: pass an unknown type, get a clean error rather than a 500.
- [ ] Manual: pass a new type whose required fields the existing data lacks, get a normal validation error.
- [ ] Manual: pass both `primaryType: "bug"` and `typeTags: ["regression"]`, the resulting `type_tags` is `["bug", "regression"]` (explicit `typeTags` wins, primary type at the front).

## Note about CI

The Unit Tests workflow currently fails on main (since `5c0cf068` for #207, `WorkspaceEventBus-nested-gitignore.test.ts` has 3 isolation failures and one assertion mismatch). My PR will inherit that failure. The TypeScript Check passes cleanly. Verified locally that this PR does not introduce any additional test failures - the 21 tracker tests still pass.

## Closes

Fixes #79.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-<!-- New features go here -->
+- Allow `mcp__nimbalyst-tracker__tracker_update` to change the primary type of a tracker item. `tracker_update` previously only accepted `typeTags` as a secondary-tag list - the primary `type` (the thing that determines whether an item shows in the `task` / `idea` / `plan` / `bug` view) was set at creation time and immutable, so the only workaround for a `task` that turned out to be a `bug` was to archive it and recreate, losing comments, attachments, and session links. Adds a `primaryType` parameter that (a) validates against the tracker registry via `globalRegistry.has(newType)`, (b) updates `tracker_items.type` in a separate UPDATE before the existing data write, (c) re-validates the item's data against the new type's schema, (d) appends a `type_changed` activity-log entry, (e) rebuilds `type_tags` so the new primary type is at the front and the old primary type is removed (preserving any other secondary tags). When the user passes both `primaryType` and `typeTags` the explicit `typeTags` array wins. Fixes #79.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/electron/src/main/mcp/tools/trackerToolHandlers.ts
+++ b/packages/electron/src/main/mcp/tools/trackerToolHandlers.ts
@@ -669,6 +669,11 @@ export const trackerToolSchemas = [
           items: { type: "string" },
           description: "Set type tags (replaces existing type tags). Primary type is always included.",
         },
+        primaryType: {
+          type: "string",
+          description:
+            "Replace the item's primary type (e.g. change a 'task' into a 'bug', or an 'idea' into a 'plan'). Must be a registered tracker type. Item history (comments, attachments, session links) is preserved. The primary type is also auto-merged into type_tags so it remains the canonical primary tag.",
+        },
         fields: {
           type: "object",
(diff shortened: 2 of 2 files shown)


ademczuk:
Hey @ghinkle, MCP-side half of #79. Adds `primaryType` to `tracker_update`. UI-side (a type selector in the tracker detail panel) is a separate ticket - happy to follow up once you pick where the selector lives. All 21 existing `trackerToolHandlers.test.ts` tests still pass; CI Unit Tests will be red because of the inherited `WorkspaceEventBus-nested-gitignore.test.ts` failures from main, but TypeScript Check is green.

Issue #135, opened 2026-05-02
[Feature]: Programable action buttons


arcenik86:
### Describe the feature request

 The ability to create actions (buttons) above/below the message bar that let you assign certain things like a snippet of code you frequently run.

ghinkle:
Any of the designs below look right?

<img width="2346" height="2016" alt="Image" src="https://github.com/user-attachments/assets/58f24e4d-c194-40b5-ac3f-d7ee096ff2f6" />

<img width="754" height="310" alt="Image" src="https://github.com/user-attachments/assets/ef1860c7-75b9-443a-8c49-a4c0c2546f79" />

arcenik86:
B. Below the input field looks most natural. Less clutter for sure. C.
Would be the next best choice.

arcenik86:
Is it possible for this to go beyond slash commands? There is some input that gets repetitive and having this loaded on a button would be great.  A bonus would be the ability to have this on the terminal pane too. Likely toggle controlled so people can decide if they want to burn the vertical space or not.

ghinkle:
Implemented on main in commit 6f4d978ba (May 16, 2026).

Nimbalyst now has an Actions control in the AI composer that reads reusable prompts from nimbalyst-local/ai-actions.md and inserts the selected prompt/snippet into the draft verbatim, with undo support, file watching, help text, and tests.

That covers the core request for reusable composer actions beyond slash commands. The shipped UX is a dropdown in the composer controls row rather than always-visible buttons above/below the input. If chip-style buttons or terminal-pane support are still wanted, those should be tracked as follow-up issues.

PR #228, opened 2026-05-09
fix(file-watcher): make supportsRecursiveWatch lazy and use path.sep in boundary check

ademczuk:
## Summary

Unblocks the broken Unit Tests CI on `main` since `5c0cf068` (the #207 fix). All four `WorkspaceEventBus-nested-gitignore.test.ts` cases now pass; production code behaves identically.

## Root cause

Two compounding platform-mocking bugs in `WorkspaceEventBus.ts`. Neither affects production runtime (`process.platform` doesn't lie there) but both break the test harness, and the test failure pattern depends on which platform CI runs on.

**1. `supportsRecursiveWatch` evaluated at module load**

```ts
const supportsRecursiveWatch = process.platform === 'darwin' || process.platform === 'win32';
```

The test forces `process.platform = 'darwin'` via `vi.hoisted` to exercise the recursive-watch code path. But the `const` is computed once when the module first loads. If any earlier test in the same vitest worker imports `WorkspaceEventBus` before the override takes effect, the value freezes to the real OS. On Linux CI this means `supportsRecursiveWatch` stays `false`, `subscribe` routes through chokidar instead of the mocked `fs.watch`, the test's `mockWatcherCallbacks` array stays empty, and `fireWatchEvent` throws `Error: No watcher callback registered`.

**2. Boundary check uses a `process.platform`-derived separator**

```ts
const sep = process.platform === 'win32' ? '\\' : '/';
const boundaryWithSep = workspaceAbs.endsWith(sep) ? workspaceAbs : workspaceAbs + sep;
if (absolutePath !== workspaceAbs && !absolutePath.startsWith(boundaryWithSep)) {
  return null;
}
```

When the test mocks `process.platform = 'darwin'` on a Windows host, this branch picks `sep = '/'`. Real `path.join` and `mkdtempSync` keep returning Windows-style paths with `\`. The boundary becomes `C:\...\Tkz1xE/`, the `absolutePath` is `C:\...\Tkz1xE\nested\...`, the `startsWith` fails, the function returns null, the nested-repo `.gitignore` never gets consulted, and the test's "drops content events for files inside a nested-repo gitignored dir" sees `onChange` fire when it shouldn't.

## Fix

Both are small. Both make the implementation more correct against any future platform mocking.

```ts
function supportsRecursiveWatch(): boolean {
  return process.platform === 'darwin' || process.platform === 'win32';
}
```

8 call sites updated to call the function. And inside `findGitRootForPathCached`:

```ts
const sep = path.sep;
```

`path.sep` is locked to the actual OS at module load via the path module variant (`path.win32` vs `path.posix`) and is independent of `process.platform` runtime mocking.

## What stays the same

- Every existing call site keeps its semantic meaning - `if (supportsRecursiveWatch)` -> `if (supportsRecursiveWatch())`. No control-flow change in production.
- `path.sep` and `process.platform` agree at runtime, so production behavior is byte-for-byte identical.
- No behavior change in the recursive-watch path, the chokidar path, the gitignore filter, the nested-repo discovery, or the event dispatcher.
- `loadGitignoreFilter`, `loadGitignoreFilterSync`, `isGitignoredScoped`, `getGitignoreAction`, and the chokidar `ignored` callback are all unchanged.

## Verification

Before this PR (on `main` at `e387c040`):
- `WorkspaceEventBus-nested-gitignore.test.ts`: 3 failed / 1 passed (Windows local), 4 failed / 0 passed (Linux CI per `CI #133` / `CI #135` for v0.59.3 release)
- Total file/ tests: 17 failed / 38 passed (Windows local)

After this PR:
- `WorkspaceEventBus-nested-gitignore.test.ts`: 4/4 pass
- Total file/ tests: 14 failed / 41 passed (3 net improvement, 0 regressions)

The remaining 14 failures (`FileSnapshotCache.test.ts` and `WorkspaceEventBus-gitignore-bypass.test.ts`) are pre-existing on `main` and unrelated to this PR.

## Test plan

- [x] `npx vitest run packages/electron/src/main/file/__tests__/WorkspaceEventBus-nested-gitignore.test.ts` - 4/4 pass
- [x] No new failures introduced in `packages/electron/src/main/file/` test suite
- [ ] CI Unit Tests on Linux turns green for this PR (and consequently for #226 and #227 once they rebase against the merged main)

## Note on context

@ghinkle, this is the test-architecture fix I offered on PR #226's CI-failure note. Happy for you to take it in a different direction if you have a preferred shape - e.g. you might want to keep the `const` and rework the test to use `vi.resetModules()` instead. The lazy-function form is the minimum-change approach that works on all three CI platforms (macOS, Windows, Linux) and against the existing test mocking strategy without touching the test file at all.

## Closes

Refs #207.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- `WorkspaceEventBus` now evaluates `supportsRecursiveWatch` lazily and uses `path.sep` (not a `process.platform`-derived separator) inside `findGitRootForPathCached`, so the `WorkspaceEventBus-nested-gitignore.test.ts` suite added with #207 passes regardless of which test file imports the module first or which OS runs the tests. Previously the module-level `const supportsRecursiveWatch = process.platform === 'darwin' || process.platform === 'win32'` froze the value at module load, so vitest workers that imported the bus before the test's `vi.hoisted` `process.platform = 'darwin'` override took effect locked the value to the real platform (false on Linux CI), routing `subscribe` through the chokidar path instead of the mocked `fs.watch` and producing `Error: No watcher callback registered`. Separately, the boundary check at line 320 used a `process.platform === 'win32' ? '\\' : '/'` separator that clashed with the actual OS-style separator returned by real `path.join` on Windows when `process.platform` was mocked to `darwin`, causing the nested-repo discovery to give up and treat `nested/rootfs/etc/foo.txt` as not-ignored. Both the `supportsRecursiveWatch` accessor and the separator are now resolved at call-time. Production code behaves identically because `process.platform` and `path.sep` agree there. Refs #207.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/file/WorkspaceEventBus.ts
+++ b/packages/electron/src/main/file/WorkspaceEventBus.ts
@@ -12,8 +12,18 @@ import { isPathInWorkspace } from '../utils/workspaceDetection';
  * macOS uses FSEvents (1 FD for the entire tree).
  * Windows uses ReadDirectoryChangesW (1 handle for the entire tree).
  * Linux does NOT support recursive: true and throws ERR_FEATURE_UNAVAILABLE_ON_PLATFORM.
+ *
+ * Implemented as a function (rather than a module-level const) so test
+ * setups that override `process.platform` via `vi.hoisted` (see
+ * `WorkspaceEventBus-nested-gitignore.test.ts`) pick up the override
+ * regardless of which test file imports this module first. A const would
+ * lock the value at module-load time, so any earlier test importing this
+ * module before the override ran would freeze the real platform value
+ * for every subsequent test in the same worker.
  */
-const supportsRecursiveWatch = process.platform === 'darwin' || process.platform === 'win32';
+function supportsRecursiveWatch(): boolean {
+  return process.platform === 'darwin' || process.platform === 'win32';
+}
 
 /**
  * .git is always ignored — it's an internal data structure, never user content.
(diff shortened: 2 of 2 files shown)


ademczuk:
@ghinkle this is the test-architecture fix I offered on PR #226. Two compounding `process.platform`-mocking bugs in the #207 commit; production behavior unchanged. 4/4 nested-gitignore tests now pass. The other file-watcher test failures on main (`FileSnapshotCache`, `WorkspaceEventBus-gitignore-bypass`) are pre-existing and not addressed here.

If you'd rather keep the `const` and rework the test (e.g. `vi.resetModules()` before the WorkspaceEventBus import), say the word and I'll close this.

ademczuk:
@ghinkle moving this to draft, the fix isn't sufficient.

**What I confirmed**: the `process.platform === 'win32' ? '\\' : '/'` boundary-separator issue at line 320 IS real and breaks the tests on Windows local. After the fix, all 4 nested-gitignore tests pass on Windows local (3 failed / 1 passed before -> 4/4 pass after).

**What I did NOT fix**: Linux CI still shows the same `Error: No watcher callback registered` for all 4 tests after my push (`67ab4b0a`). The lazy `supportsRecursiveWatch` change doesn't actually unblock Linux CI - which means there's a separate Linux-only issue with how `Object.defineProperty(process, 'platform', { value: 'darwin', writable: true })` interacts with Vitest's `vi.mock('fs', ...)` on Linux that I haven't pinpointed. On Windows, the fix happens to work because real `process.platform === 'win32'` already satisfies `supportsRecursiveWatch`, so the mock is functionally a no-op for the routing decision.

**What I'd suggest instead** if you want to take a swing at it:
- The tests might need `vi.resetModules()` followed by a fresh `await import('../WorkspaceEventBus')` inside `beforeEach`, plus moving the platform override into the same scope so the WorkspaceEventBus module loads fresh against the mocked platform per test.
- Or a `__setSupportsRecursiveWatchForTest()` exported helper that the test can poke directly (less elegant, but explicit).
- Or an architectural shift: pass the platform decision into `subscribe` as an option so the test never has to mock `process.platform` at all.

Holding this branch in case the small parts (the `path.sep` boundary fix and the lazy accessor) are useful as a partial defense even if they don't unblock CI on their own. Veto and I'll close.

ademczuk:
@ghinkle your `9141dc3a` (nest tmpdir to clear the workspace depth check) is the right fix. Different root cause than my supportsRecursiveWatch / path.sep diagnosis - the depth check rejects the unsafe path before subscribe ever calls fs.watch, which is why the mock callback never registered and why my partial fix only worked on Windows local. One-line test setup change vs my eight-site implementation refactor. Cleaner. Closing this draft, ta.

PR #291, opened 2026-05-14
fix(onboarding): bound onboarding:get IPC with 3s timeout to prevent picker hang

### LINKED ISSUE #260: Beach ball / hang on developer-vs-standard mode picker in initial setup (cannot complete onboarding)
## Summary
On initial setup, when the picker for "developer mode vs standard mode" is shown, the app hangs (macOS spinning beach ball) and must be force-quit. Onboarding cannot be completed.

## Steps to reproduce
1. Install latest stable Nimbalyst on macOS.
2. Launch the app for the first time and proceed through initial setup.
3. Reach the screen offering "developer mode" or "standard mode".

## Expected behavior
Clicking either option proceeds to the next onboarding step and the app finishes setup.

## Actual behavior
The app freezes on the mode-select screen, the macOS beach ball appears, and the only way out is Force Quit. The user has never been able to get past this step across multiple launches.

## Environment
- App version: latest stable (0.60.0)
- OS: macOS (Apple Silicon, exact version unknown)
- Build: official packaged build
- Reported by a second user (not the filer)

## Logs
Main process logs across 3 cold-start attempts show clean startup with no errors up to and past `[Boot] Main process initialization complete`. The hang appears to be entirely renderer-side at the mode-select screen — no useful main-process signal in the logs we have. Renderer debug logs were not available from the affected machine.

ademczuk:
Fixes #260.

## Symptom

@karlwirth's reporter (a second user, not Karl himself) hit a macOS beach ball on the developer-vs-standard mode picker during initial setup. The dialog opened, neither mode card highlighted, the "Get Started" button never became reachable, and the only escape was Force Quit. Reproducible across multiple cold-start attempts on macOS 0.60.0.

## Root cause

Two `useEffect` hooks fire the same `onboarding:get` IPC channel on dialog open:

1. `packages/electron/src/renderer/components/UnifiedOnboarding/UnifiedOnboarding.tsx:86-101` - the in-component `checkExistingUser` effect.
2. `packages/electron/src/renderer/hooks/useOnboarding.ts:263-294` - the outer gate that decides whether to open the dialog at all.

The main-process handler at `packages/electron/src/main/ipc/SettingsHandlers.ts:375-378` uses a dynamic `await import('../utils/store')` on every invocation. On a fresh cold start the first import resolves slowly, and concurrent calls from the two renderer paths can race on the lock for the underlying electron-store file.

Both renderer call sites awaited the IPC unconditionally with no timeout. When either promise never resolved, React kept rendering the dialog's initial state (`isExistingUser = false`, `developerMode = null`, all form fields blank, "Get Started" button disabled because no mode is selected). The mode-card click handlers are pure local `setState` and DO work even during the hang, but visually the user gets no feedback because both effects are still mid-flight and the dialog never reconciles its "loaded" state.

## Fix

Race the IPC against a 3-second timeout at both call sites. On timeout:

- `UnifiedOnboarding.tsx`: fall back to `setIsExistingUser(false)` (the conservative new-user path). The dialog renders fully and becomes interactive.
- `useOnboarding.ts`: fall through to the "show dialog" branch. The user gets the picker instead of a silent no-op.

```ts
const state = await Promise.race([
  window.electronAPI.invoke('onboarding:get'),
  new Promise<null>((resolve) => setTimeout(() => resolve(null), 3000)),
]);
if (state === null) {
  console.warn('[onboarding] onboarding:get IPC timed out after 3s; proceeding...');
  // fallback path
}
```

The happy path (IPC resolves under 3s) is byte-identical to the old code. The bad path now degrades to "interactive dialog with new-user defaults" instead of "permanently stuck render."

## Tests

3 cases in `packages/electron/src/renderer/components/UnifiedOnboarding/__tests__/UnifiedOnboarding.ipcTimeout.test.tsx`:

1. Never-resolving IPC times out at 3s and the component completes its render path.
2. Rejecting IPC still proceeds (existing catch block preserved).
3. Happy-path IPC continues to populate `isExistingUser` correctly (regression guard).

Tests use `vi.useFakeTimers()` to advance past the timeout deterministically. Cross-platform - no macOS required.

## Regression risk

Low. Behaviour when the IPC resolves under 3s is unchanged. Behaviour when it doesn't changes from "permanent beach ball" to "interactive dialog assuming new user." Even on the worst case (IPC eventually returns "existing user" data after the 3s cutoff) the user is in a worse onboarding flow than ideal but is no longer stuck. The data the IPC returns is informational (it controls which fields appear) and the user can still complete onboarding successfully from the new-user path.

A proper longer-term fix should de-duplicate the two parallel IPC calls and ensure the main-process handler's dynamic import does not race on store lock. That is out of scope for this PR; the timeout is the defensive guard that unblocks the user today.

## Changelog


ghinkle:
I don't think this is likely to fix the problem seen by the user. onboarding:get would have fired before opening the dialog anway. These dynamic imports for utils/store were never really necessary though and i'm going to try and clean up the settings patterns here soon. Wil come back to this.

ademczuk:
@ghinkle thanks for the review, and apologies for the noise - this PR (and 6 others of mine) was auto-closed when the fork relationship on `ademczuk/nimbalyst` was reset earlier today to support an unrelated kimiclaw-frontend experiment. The branches survive and I've resubmitted from a fresh fork at #313.

On your substantive point: you're right that `onboarding:get` fires before the dialog opens, so a timeout on `onboarding:get` itself wouldn't directly fix a hang INSIDE the dialog. The repro path I had in mind was different though - @karlwirth's #260 shows the dialog rendering its initial "no mode selected" state with no interactivity, which is what you see if the `onboarding:get` promise never resolves and the renderer holds its initial state (no error, no fallback). The 3s timeout makes that path bail to the conservative new-user / show-dialog fallback so the dialog stays interactive.

Happy to defer to your settings-pattern cleanup if you'd rather land the structural fix and skip this one. Two options for you:
1. Land #313 as a short-term safety net until the dynamic-imports cleanup ships (then revert this once the cleanup makes the timeout moot)
2. Close #313 as superseded and wait for your cleanup

Either works for me. Flagging that #260 stays open for affected users in the interim either way.

PR #292, opened 2026-05-14
fix(updater): drop redundant checkForUpdates that tore down Squirrel.Mac proxy

### LINKED ISSUE #245: Auto-update fails with "The command is disabled and cannot be executed" immediately after download completes (macOS arm64)
## Summary
In-app auto-update consistently fails on macOS arm64: the update zip downloads successfully, then `autoUpdater.quitAndInstall()` immediately errors with `"The command is disabled and cannot be executed"`. The user is left with the update toast showing an error, and the only workaround is to manually download and replace the bundle.

## Steps to reproduce
1. Run a packaged Nimbalyst build on macOS (Apple Silicon).
2. Wait for the in-app update prompt, or trigger an update check.
3. Click "Update now" on the update toast.
4. Observe: zip downloads to `~/Library/Caches/@nimbalystelectron-updater/pending/`, then the toast flips to an error within ~1ms of "Update downloaded".

Reproduces on every attempt across multiple release versions. Persists across:
- Manually replacing `/Applications/Nimbalyst.app` with a freshly downloaded build (multiple times)
- `launchctl remove com.nimbalyst.electron.ShipIt`
- Deleting `~/Library/Caches/com.nimbalyst.electron.ShipIt/` (stuck `ShipItState.plist` from a prior aborted install)
- Deleting `~/Library/Caches/@nimbalystelectron-updater/pending/`
- Stripping `com.apple.quarantine` xattrs from `/Applications/Nimbalyst.app` (originally applied by Arc browser on download)

Code signature verifies clean (`codesign --verify --deep --strict` -> "valid on disk", "satisfies its Designated Requirement"), notarization stapled, Developer ID intact.

## Expected behavior
After clicking "Update now", the app quits and relaunches at the new version.

## Actual behavior
The download completes, then `autoUpdater.quitAndInstall()` synchronously throws the Squirrel.Mac error `"The command is disabled and cannot be executed"`. The renderer receives `update-toast:error`. App stays at the old version. No update is ever applied via the in-app flow.

The error appears to originate from Squirrel.Mac's `SQRLUpdater` (returned when its update downloader is nil at check-for-updates time), suggesting the electron-updater HTTP proxy server is being torn down before Squirrel can consume it. Note the `"Proxy server for native Squirrel.Mac is closed"` log line immediately preceding the error.

History: there was a prior aborted install on 2026-04-30 where ShipIt failed with `SQRLInstallerErrorDomain` (source bundle disappeared mid-install, see logs below). ShipIt aborted after 3 retries but never cleared its `ShipItState.plist`, which may have contributed to subsequent failures. Clearing that state did not fix the current symptom; the proxy-server-closed-before-quitAndInstall failure reproduces from a fully clean slate.

## Environment
- App version: 0.59.2
- Platform: darwin arm64
- OS: 25.4.0 (macOS)
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153
- Install source: GitHub release download via Arc browser
- App location: `/Applications/Nimbalyst.app` (owned by user, writable)
- Not Homebrew-managed, not MDM-managed as a Library Item

## Logs

Most recent failure (`~/Library/Application Support/@nimbalyst/electron/logs/main.log`):

```
[2026-05-11 17:14:46.083] Sending update-toast:show-available to frontmost window
[2026-05-11 17:14:46.084] Downloading update from Nimbalyst-macOS-arm64.zip, ...
[2026-05-11 17:14:46.243] Proxy server for native Squirrel.Mac is closed
                          (fileToProxy=https://github.com/nimbalyst/nimbalyst/releases/download/v0.59.2/Nimbalyst-macOS-arm64.zip)
[2026-05-11 17:14:46.244] Update downloaded: {
  downloadedFile: '~/Library/Caches/@nimbalystelectron-updater/pending/Nimbalyst-macOS-arm64.zip'
[2026-05-11 17:14:46.244] Sending update-toast:show-ready to frontmost window
[2026-05-11 17:14:46.244] Error: The command is disabled and cannot be executed
[2026-05-11 17:14:46.245] Update error: Error: The command is disabled and cannot be executed
[2026-05-11 17:14:46.245] Failed to check and download latest: Error: The command is disabled and cannot be executed
```

Earlier ShipIt failure (`~/Library/Caches/com.nimbalyst.electron.ShipIt/ShipIt_stderr.log`, 2026-04-30):

```
ShipIt[3524] Installation error: Error Domain=SQRLInstallerErrorDomain Code=-1
  "Failed to copy bundle ~/Library/Caches/com.nimbalyst.electron.ShipIt/update.0F1B4JL/Nimbalyst.app/
   to directory /var/folders/.../com.nimbalyst.electron.ShipIt.a9C8EMTn/Nimbalyst.app"
  NSUnderlyingError=... "The file 'Nimbalyst.app' couldn't be opened because there is no such file."
[3 install attempts all failed with NSCocoaErrorDomain Code=260]
ShipIt[6609] Too many attempts to install, aborting update
```

`ShipItState.plist` was never cleared after that abort, leaving stale `updateBundleURL` pointing at the deleted `update.0F1B4JL/Nimbalyst.app/`.

ademczuk:
Fixes #245.

## Symptom

@adambhenry reported the auto-update flow failing on macOS arm64 immediately after the download completes, with a toast showing "The command is disabled and cannot be executed." The log line `"Proxy server for native Squirrel.Mac is closed"` fires at the same millisecond timestamp as `"Update downloaded"`, before `quitAndInstall` is invoked. Multiple workaround attempts (manual restart, re-download) all produced the same NSException.

## Root cause

`AutoUpdaterService.checkAndDownloadLatest` was calling `autoUpdater.checkForUpdates()` immediately before `autoUpdater.downloadUpdate()` to "get the absolute latest version" (lines 637-661). On macOS this is fatal: each `checkForUpdates()` call spins up a new Squirrel.Mac proxy server, and the new proxy tears down the prior one that Squirrel's internal `SQRLUpdater` already holds a reference to from the original `update-available` event.

The sequence:

1. `update-available` event fires `pendingUpdateInfo = {version, ...}`. Squirrel.Mac's `SQRLUpdater` has the proxy URL.
2. User clicks "Install now." `checkAndDownloadLatest` runs.
3. `await autoUpdater.checkForUpdates()` opens a NEW proxy, closes the previous one. SQRLUpdater's reference is now dangling.
4. `await autoUpdater.downloadUpdate()` runs (succeeds on the new proxy).
5. `update-downloaded` fires, toast shows "ready to install."
6. User clicks "Install." `quitAndInstall(true, true)` runs.
7. Squirrel's internal downloader tries to hand off to the proxy it was given in step 1 - which is closed. NSException: "The command is disabled and cannot be executed."

arm64 reproduces it more reliably than x86_64 because process scheduling differences make the proxy-teardown race fire earlier.

## Fix

Two parts in `packages/electron/src/main/services/autoUpdater.ts`:

**1. Drop the redundant pre-download check in `checkAndDownloadLatest`.** The `update-available` event has already populated `pendingUpdateInfo` with the version we want to download; there is no reason to re-check before downloading. Just call `downloadUpdate()` directly so the single Squirrel session that fired `update-available` stays alive through `quitAndInstall`:

```ts
private async checkAndDownloadLatest() {
  try {
    log.info('Starting update download (single-check path to avoid Squirrel.Mac proxy race)...');
    await autoUpdater.downloadUpdate();
  } catch (error) {
    // ... existing error handling
  }
}
```

**2. Add `squirrel_install_disabled` to `classifyUpdateError`.** Today this NSException collapses into the generic "unknown" bucket, which means analytics cannot detect a regression and the toast has nothing to surface. The new branch matches both the full phrase and the shorter "cannot be executed" suffix so future Electron / electron-updater versions that re-wrap the NSException stay classified correctly. `classifyUpdateError` is exported so the test does not need the full service.

## Tests

8 cases in `packages/electron/src/main/services/__tests__/autoUpdater.classifyUpdateError.test.ts`:

- Full Squirrel NSException maps to `squirrel_install_disabled`
- Trailing-fragment "cannot be executed" also matches (forward compatibility)
- Network / permission / disk / signature classifications still work (regression guards)
- Generic "unknown" fallback still triggers
- Precedence: network beats squirrel if both phrases appear (documents branch order)

## Regression risk

Low.

- The proxy-teardown bug only manifested when `checkAndDownloadLatest` ran (which is only invoked from the "Install" toast handler, not from background checks). Removing the pre-download check has no effect on the background `update-available` flow.
- The behavior change for users is: "Install now" used to silently re-check, now it goes straight to download. The version is already known from the `update-available` event that triggered the toast, so the user gets the version they were prompted about.
- `classifyUpdateError` returns a new string in one previously-unknown case. Any code that switches on the result of this function would only see the new value when the new failure mode occurs (i.e. the failure mode where the old "unknown" was already useless).

## Why no Mac was needed

The diagnosis came from reading the existing logs in the issue ("Proxy server for native Squirrel.Mac is closed" at the same millisecond as "Update downloaded"), the electron-updater source, and the call chain. The fix is purely a code restructuring; no signing or notarization is involved. Upstream nimbalyst's macOS CI (per `docs/E2E_TESTING.md` and `build:mac:notarized`) will validate the build on PR open.

## Changelog


ghinkle:
This extra check was added to eliminate the scenario where an update is released while the user has the last update notice open. (e.g. they were away from their computer) This causes the user to go through the whole update process and then they get another update immediately on restart.

Alternatively we could reduce the updater friction by switching to an auto-download policy and only show the user a single "Restart to Update" message. I think that is becoming more common these days. Thoughts?

ademczuk:
@ghinkle thanks for the context, and apologies for the noise - this PR (and 6 others of mine) was auto-closed when the fork relationship on `ademczuk/nimbalyst` was reset earlier today to support an unrelated kimiclaw-frontend experiment. Branches survive; I've resubmitted at #314.

On the design question: I had not realized the second `checkForUpdates()` was there to catch a release that landed while the user was away. That's a real scenario. The Squirrel.Mac proxy tear-down on macOS still seems like the wrong tradeoff to me though - users on adambhenry's reported config (arm64) hit it reliably and the recovery is "force-quit + reinstall", which is worse than seeing the version-mismatch banner one launch later.

Auto-download + single "Restart to Update" is the right direction, agreed. It also sidesteps the proxy issue entirely because the second check moves to a context where we are not mid-quitAndInstall.

Two questions before I rework:
1. Do you want the auto-download switch in the same PR as the Squirrel race fix, or separate? The race fix is the smaller surface and could land first to unblock adambhenry; the policy change is a bigger UX call.
2. For the auto-download model - is there a precedent in the codebase (e.g., extension updater) for the cancellation + retry path, or would this be a new pattern?

Happy to take either path. If you want me to ship the policy change in a fresh PR I can draft it; if you want to handle this one yourself (you may have more context on what's already in flight) just close #314 and I'll pivot.

Issue #231, opened 2026-05-10
Worktree child sessions can appear non-running while actively streaming


ghinkle:
## Summary
Sibling worktree sessions launched from `/nimbalyst-planning-launch-new-session` can remain visible as non-running while actively executing.

## Symptoms
A child session launched from planning was actively streaming output, but the session/workstream UI did not show it as `running`.

Example affected session:
- Session ID: `a8282d0a-6442-4dc9-92ff-10c703ab152e`
- Canonical workspace: `<WORKSPACE>`
- Worktree path used by the running child: `<WORKSPACE>_worktrees/bright-tide`

In the observed case, the UI stayed in a non-running state and could remain stuck on the waiting-for-input indicator even though the session state manager reported the child as active, `status: running`, and `isStreaming: true`.

## Expected Behavior
If a sibling/child session is actively executing in a worktree, the session list and workstream UI should reflect that it is running.

## Reproduction
1. Start from a planning/workstream session.
2. Launch a sibling session via `/nimbalyst-planning-launch-new-session` into a worktree.
3. Let the child begin streaming.
4. Observe that the child can continue running while the parent workspace UI does not update its running state.

## Root Cause
Session lifecycle events for worktree child sessions were routed using the event's `workspacePath`, which was the worktree path. The session list/workstream UI for the parent session was subscribed under the parent project's canonical workspace path.

That mismatch caused `session:started` and `session:streaming` events to be filtered out before they reached the renderer. As a result, `sessionProcessingAtom` and `sessionHasPendingInteractivePromptAtom` could remain stale for those child sessions.

## Proposed Fix
Use the session's canonical `ai_sessions.workspace_id` as the owning workspace when available, allow workspace-scoped subscriptions to accept worktree-originated events for sessions that belong to the subscribed parent workspace, and normalize forwarded session-state events to the canonical workspace before renderer-side filtering.

## Notes
A focused fix was implemented locally in:
- `packages/electron/src/main/ipc/SessionStateHandlers.ts`
- `packages/electron/src/renderer/store/sessionStateListeners.ts`
- `packages/electron/src/shared/sessionWorkspaceRouting.ts`

A focused unit test was added for the routing logic.

ademczuk:
Pre-emptive code-read. The routing-normalization approach is the right fix. One regression risk worth verifying:

[...]

PR #293, opened 2026-05-14
fix(sessions): clear processing atom on terminal events even without workspacePath

### LINKED ISSUE #116: "Thinking…" indicator stays on after Claude Code response completes; requires Cancel to clear
## Summary
After a Claude Code response fully completes in Nimbalyst, the "Thinking…" indicator remains visible and only goes away when the Cancel Request button is clicked.

## Steps to reproduce
1. Open a Claude Code agent session in Nimbalyst (latest build).
2. Send any prompt and wait for the response to finish.
3. Observe the chat area after the response is fully rendered and any tool calls have completed.

## Expected behavior
The "Thinking…" indicator clears automatically once the response is complete, and the input is ready for the next message without manual intervention.

## Actual behavior
The "Thinking…" indicator continues to display indefinitely after the response is complete. Clicking the Cancel Request button is required to clear it. Cancel cleanly clears the state and the next message can be sent normally. Reproduces on every response.

## Environment
- App version: latest build (noticed today, 2026-04-30)
- Agent: Claude Agent · Opus 4.7 (1M)
- Thinking mode: High
- OS: Windows 11 Pro

## Logs
Not collected (user opted out of log gathering).

ademczuk:
Fixes #116.

## Symptom

@arcenik86 reported that the "Thinking..." indicator stays on after Claude Code finishes responding. The assistant's reply renders correctly, but the spinner pinned at the top of the session never clears until the user clicks Cancel. Reproducible on extended-thinking Opus sessions where slower turns make the underlying race more likely.

## Root cause

Commit `4e5dd9e7` (May 11, fixing #231 multi-project rail routing) added an `ownedWorkspacePath` null-guard to `sessionStateListeners.ts:217-234`:

```ts
const ownedWorkspacePath = resolveOwnedWorkspacePath({
  eventWorkspacePath,
  sessionWorkspacePath,
});
if (!ownedWorkspacePath) {
  return;   // <- silently drops the event
}
```

`resolveOwnedWorkspacePath` returns `null` when:
- The IPC event payload has no `workspacePath` field, AND
- The session is not yet in `sessionRegistryAtom` (no `sessionMeta.workspaceId`)

This combination fires during:
- A startup race where the lifecycle event arrives before the session list is hydrated
- Post-HMR re-evaluation that re-mounts the listener after the session was registered
- Fresh-session edge cases where `update-available` (sorry, `session:completed`) arrives faster than `session:list` populates

When that combination hits a terminal event (`session:completed` / `session:error` / `session:interrupted`), `sessionProcessingAtom(sessionId)` is never set to `false` and `RichTranscriptView.isWaitingForResponse` keeps returning `true`. The spinner is pinned forever.

Extended-thinking sessions reproduce the race more reliably because longer turns increase the chance of HMR, focus changes, or registry-population delays during the turn.

## Fix

For terminal lifecycle events, clear `sessionProcessingAtom(sessionId)` and `sessionHasPendingInteractivePromptAtom(sessionId)` **unconditionally before** the `ownedWorkspacePath` null-guard:

```ts
const isTerminalEvent =
  type === 'session:completed' ||
  type === 'session:error' ||
  type === 'session:interrupted';
if (isTerminalEvent) {
  store.set(sessionProcessingAtom(sessionId), false);
  store.set(sessionHasPendingInteractivePromptAtom(sessionId), false);
}

if (!ownedWorkspacePath) {
  return;
}
```

These atoms are keyed by `sessionId` alone (no workspace component) and represent "is this turn still running" / "is there a pending interactive prompt." A missing `workspacePath` is no reason to keep them `true`. The workspace-routed branches below (the safety-net reload via `reloadSessionDataAtom`, the streaming-bucket cleanup via `clearSessionStreamingAtom`) still run only when the path is known, so multi-project rail routing semantics from #231 are preserved.

## Tests

3 new regression cases in `packages/electron/src/renderer/store/__tests__/sessionStateListeners.test.ts`, one per terminal event type. Each asserts that dispatching `{ type, sessionId }` with **no `workspacePath`** and **no registry entry** still clears `sessionProcessingAtom` AND `sessionHasPendingInteractivePromptAtom`. All 18 tests in the suite pass locally.

## Regression risk

Low.

- Behaviour with a resolvable `workspacePath` is unchanged - the existing terminal-event handler still runs and the workspace-routed branches still execute. The new pre-guard clear duplicates the work already done inside the switch, but the operation is idempotent (`store.set(atom, false)` is a no-op when already false).
- Multi-project rail routing semantics from #231 are preserved: events for sessions in unrelated workspaces don't run the workspace-routed branches because `ownedWorkspacePath` is still required for those.
- The only behaviour change is for the previously-broken case: terminal events with no workspace path now clear the processing flag instead of dropping silently.

## Changelog


ghinkle:
Thanks for the fix here — clearing the stuck per-session processing state is the right direction. I do see one blocker before merge: in packages/electron/src/renderer/store/sessionStateListeners.ts, terminal events still return before clearSessionStreamingAtom runs when the event payload and registry lookup both lack workspacePath, so cross-workspace rail activity can stay stale even after the session ends. Please also clear the streaming activity in that path, and maybe add a regression test for the no-workspacePath case.

Separately, we do want to start tightening this up at the source: the longer-term direction is that all emits should carry a real workspacePath instead of relying on fallbacks. That doesn’t need to block this PR by itself, but as we touch these paths we want to move toward enforcing that invariant rather than depending on renderer-side recovery.

ademczuk:
@ghinkle thanks for the thorough review, and apologies for the noise - this PR (and 6 others) was auto-closed when the fork relationship on `ademczuk/nimbalyst` was reset earlier today to support an unrelated kimiclaw-frontend experiment. Resubmitted at #315 with the blocker addressed.

You were right: `clearSessionStreamingAtom` only fired inside the switch case AFTER the no-workspacePath early-return, so cross-workspace rail "streaming" badges stayed stuck on terminal events for sessions in inactive projects. Fixed in #315:

- Moved the `clearSessionStreamingAtom` call into the unconditional terminal-event block above the early-return. The atom already accepts an optional `workspacePath` and falls back to `sessionActivityIndexAtom` lookup when omitted, so this works in the no-workspacePath case.
- Removed the duplicate call inside the switch case. Idempotent but obscures the invariant.
- Added a regression test that seeds the activity index via `markSessionStreamingAtom` for an inactive workspace, then fires a terminal event without workspacePath, and asserts the streaming flag is cleared.

On the longer-term direction (all emits should carry a real workspacePath): agreed. Worth noting for whoever picks that up - it's a multi-call-site change in the main process (`MessageStreamingHandler` and the lifecycle emitters in `AIService`/`SessionManager`), but it would let us drop the renderer-side fallback entirely and replace it with a hard assertion. Happy to scope a follow-up PR for that if you'd like to keep this one focused on the symptom fix.

Issue #272, opened 2026-05-13
[Bug]: Nimbalyst desktop crashes ~20–30 minutes into a session with Sync enabled


AnisminC:
## Summary
Nimbalyst desktop app crashes consistently roughly 20–30 minutes into a continuous task when Sync is enabled. Disabling Sync stops the crashes.

## Steps to reproduce
1. Enable Sync in Nimbalyst.
2. Start a Claude session and let it run a continuous task.
3. Leave the machine and monitor the session from the mobile app.
4. After ~20–30 minutes, the desktop app crashes.

## Expected behavior
The desktop app stays running for the duration of a long session with Sync enabled.

## Actual behavior
The desktop app crashes every time, ~20–30 minutes into the session / ~20–30 minutes after Sync was enabled. No visible error dialog was observed (user was away from the machine, monitoring via mobile). Turning Sync off eliminates the crashes.

## Environment
- App version: 0.59.2
- OS: Windows 11 (10.0.26200), x64
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
Not collected.

ademczuk:
@AnisminC thanks for filing. A 20-30 minute reproducible crash with Sync enabled is something I'd love to PR a fix for, but without a stack trace the root cause space is large (collabv3 sync writer, websocket reconnect storm, periodic background flush, mobile-mirror push). Could you grab a few things on the next repro so we can pin it down?

1. Crash dumps from `%LOCALAPPDATA%\@nimbalyst\electron\Crashpad\reports\` (Windows path). The most recent `.dmp` file after a crash, if Electron's crashpad caught it.
2. Renderer console + main log around the crash time. Easiest way: launch from a terminal so stderr is captured, run until crash, paste the last 50-100 lines.
   - `%AppData%\Nimbalyst\logs\main.log` should have the main-process tail
   - Renderer log lives at `%AppData%\Nimbalyst\nimbalyst-debug.log` in dev mode
3. Approximate memory usage at the moment of crash if you can spot it (Task Manager > Details > Nimbalyst processes). The 20-30 minute timing is suspicious for a slow leak.
4. Whether the crash is the main process (whole app window disappears) or just the renderer (window stays, content goes blank / "Aw, snap"). The difference narrows the suspect list a lot.

[...]

PR #294, opened 2026-05-14
fix(database): PGLite lock self-heals on Windows after PID reuse (#272)

ademczuk:
Addresses the stuck-lock symptom AnisminC reported on #272. The underlying main-process crash that left the lock orphaned in the first place is a separate bug we are still gathering data on; this PR ensures the lock self-heals on the next launch regardless of whether we reproduce the crash itself.

## Symptom

@AnisminC reported the PGLite lock file did NOT self-heal after a main-process crash on Windows. From his comment on #272:

> [PGLite Worker] Failed to initialize: Error: Failed to initialize PGlite
> Database path: ~/AppData/Roaming/@nimbalyst/electron/pglite-db
> Error: Database is locked by another Nimbalyst process.
> Lock holder PID: 6732
> Lock acquired: 2026-05-13T21:09:49.396Z
> Code: DATABASE_LOCKED

Original Nimbalyst PID 6732 crashed at 21:09Z. User relaunched at 00:56Z (3h47m later, no reboot). The new launch was blocked with `DATABASE_LOCKED` until the lock file was deleted manually. The original Nimbalyst process had been dead for almost four hours.

## Root cause

`worker.js`'s lock liveness check used `process.kill(lockPid, 0)` and interpreted ANY `EPERM` error as "process still running":

```js
catch (e) {
  isRunning = e.code === 'EPERM';
}
```

On Windows, `process.kill(pid, 0)` against a PID owned by another user or privilege level throws `EPERM` regardless of whether that PID is the original Nimbalyst process or a totally unrelated system / service process that recycled the number. Windows PID namespace is 16-bit (~32k) and recycling happens aggressively within a session, especially after 3+ hours of activity. AnisminC's PID 6732 was recycled to a system process the worker thread cannot signal. EPERM was thrown. The "still running" branch fired. Launch blocked.

The `process.kill(pid, 0)` API cannot distinguish "same process, different user" from "different process that reused the PID" using error code alone.

## Fix

Extract the decision to a pure helper `decideLockIsRunning` (`packages/electron/src/main/database/lockStaleness.js`, CJS to match `worker.js`). Inside the `EPERM` branch, gate the "still running" interpretation on lock age. A genuine competing Nimbalyst instance acquires its lock during the current launch attempt and the timestamp is always seconds-fresh. A crash-stale lock has a timestamp from before the crash (minutes to hours old). 60-second grace window.

```js
if (e.code === 'EPERM') {
  const lockAgeMs = now - new Date(lockTimestamp).getTime();
  if (lockAgeMs > 60_000) {
    return { isRunning: false, reason: 'Windows pid-reuse hazard - stale' };
  }
  return { isRunning: true, reason: 'live sibling - block as expected' };
}
```

Worker.js now calls the helper instead of inlining the kill+catch block, which also makes the logic unit-testable without spawning a real Node worker thread (the inline form was untestable).

## Tests

8 cases in `packages/electron/src/main/database/__tests__/lockStaleness.test.ts`:

1. Clean ESRCH path - dead PID, lock stale
2. Clean signalable-alive path - sibling owns the lock, block
3. AnisminC's exact EPERM-stale scenario (3h47m old lock, 13571 seconds, > 60s grace)
4. Unknown-timestamp fallback (legacy lock files without timestamp -> stale)
5. EPERM with fresh 10s timestamp -> live sibling, block as expected
6. Custom grace window override
7. Unrecognised error code (e.g. EFAULT) -> fail closed, treat as running
8. Exported `DEFAULT_STALE_LOCK_GRACE_MS` constant

All 8 pass locally.

## Regression risk

Low. The change strictly NARROWS the set of cases where EPERM is interpreted as "running." The previous code had a false-positive failure mode (lock blocked launches indefinitely after a crash). The new code retains the same protection for the genuine-sibling case (lock acquired within current launch attempt, timestamp within 60 s) while correctly clearing crash-stale locks.

The only behaviour change for users is: after a main-process crash on Windows, the next launch succeeds instead of failing with `DATABASE_LOCKED`. The fix does not affect the (separate) main-process crash itself.

The 60-second grace window is intentionally generous. A brand-new sibling that happens to race us during launch (e.g. opened from two different workspaces simultaneously) still wins the block as expected because both instances' locks are seconds-fresh.

## Changelog


ghinkle:
I was trying to figure out if the EPERM issue could be from people starting Nimbalyst during the upgrade auto-restart and the two instances being at different permission levels. We used to have a lot of reports of database corruption surrounding updates. Not sure if that would only happen if people are using run as... but that seems unlikely.

Alternatively, we could show them a dialog and let them choose to delete the lock file and start. Makes it simple and easy to recover a crash but helps avoid database corruption. But I don't know enough about Windows to know if this really necessary and defer to you.

ademczuk:
@ghinkle thanks for thinking through this with me, and apologies for the noise - this PR (and 6 others) was auto-closed when the fork relationship on `ademczuk/nimbalyst` was reset earlier today. Resubmitted at #316.

On the auto-restart-during-upgrade hypothesis: that's a real shape, but I don't think it explains @AnisminC's specific scenario. The reporter had a 3h47m gap between crash and relaunch with no auto-update in the window (Windows update events log was clean and the auto-updater wasn't pending). The PID-reuse path is independently exploitable on Windows because the 16-bit PID namespace wraps quickly under heavy load (Windows commonly reuses PIDs within hours), and `process.kill(reused_pid, 0)` against a system process owned by SYSTEM throws EPERM from a user-mode Nimbalyst regardless of whether the original Nimbalyst held a different privilege level.

On the dialog-letting-user-choose-to-delete option: I think it's complementary rather than competing. The 60s grace catches the unambiguous case (lock written long enough ago that no legitimate just-spawned sibling could own it), and a dialog still has a job for the ambiguous middle case (lock from <60s ago that we can't probe via kill). I avoided the dialog for the unambiguous-stale case because every user who hit @AnisminC's symptom reported that the only workaround they found was manual lock-file deletion - and that's worse UX than a self-heal because it requires the user to know which file to delete (the nimbalyst-db.pid one level up from pglite-db/, NOT anything inside pglite-db/ which would corrupt PGLite's WAL).

Two things I deferred and want your call on:
1. Should the dialog path land alongside this PR or as a follow-up? I have a sketch but want to be sure I am not duplicating something that's already been tried.
2. The macOS path uses the same `process.kill(pid, 0)` logic but EPERM is rarer there. Worth tightening the same way for consistency, or leave as-is until someone reports?

PR #299, opened 2026-05-14
fix(codex): drop expired rate-limit buckets so the usage indicator does not stick at a stale percentage (#120)

### LINKED ISSUE #120: Codex usage indicator stuck at 91% — does not match real Codex usage
## Summary
The Codex usage indicator in the bottom-left side panel shows ~91% usage, but actual Codex usage is 0% according to the Codex web dashboard and Codex CLI. The stale value persists across sessions.

## Steps to reproduce
1. Sign into Codex inside Nimbalyst.
2. Use Codex through Nimbalyst until the usage indicator climbs (in this case, it climbed to ~91% during a session).
3. Stop using Codex and let real usage drop to 0% (verified via https://chatgpt.com/codex/cloud/settings/analytics#usage and the Codex CLI).
4. Observe that the bottom-left panel still shows 91%.

## Expected behavior
The Codex usage indicator should track the real Codex quota, matching what the Codex dashboard / CLI report.

## Actual behavior
Indicator is stuck at 91%. It does not refresh after:
- Refreshing the panel
- Quitting and relaunching Nimbalyst
- Disabling and re-enabling Codex in settings
- Restarting the computer

The value has remained stuck across multiple sessions.

## Environment
- App version: 0.58.14
- Platform: macOS (darwin arm64), OS release 25.4.0
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

ademczuk:
## Summary

Fixes #120. The Codex usage indicator in the bottom-left panel was sticking on the most recent historical percentage (91% in @markmuir87's report) even though real Codex usage had returned to 0%. The stale value survived panel refresh, app quit-and-relaunch, disable+re-enable of Codex in settings, and full machine restart.

## Root cause

`CodexUsageService.extractRateLimitsFromEvent` (`packages/electron/src/main/services/CodexUsageService.ts:347`) returned any `token_count` event whose `rate_limits.primary` block was non-null, with no check on `primary.resets_at`. The scan-backward loop in `extractUsageSnapshotFromFile` (lines 300-319) breaks on the first match and never re-checks - so once a JSONL line carrying `used_percent: 91, resets_at: <past>` exists in `~/.codex/sessions/`, it keeps coming back.

`cachedUsage` is in-memory only and clears on app launch, but the `codex-usage:get` IPC handler calls `refresh()` on cold start (`CodexUsageHandlers.ts:16`), which immediately re-reads the same stale JSONL file and repopulates the cache with 91% within milliseconds. That's why the stuck value survives every restart path the reporter tried.

## Fix

Extract the expiry logic into a pure `filterRateLimitsByExpiry` helper that drops each bucket (primary 5h, secondary 7d) independently when its `resets_at` Unix-seconds timestamp is at or before now:

- A still-active 7-day window keeps surfacing even when the 5-hour window just reset.
- If both buckets are expired the function returns `null` so the scan continues into older lines / files for a still-active block.
- If nothing is active anywhere, the higher-level snapshot falls through to `limitsAvailable: false` and the renderer shows `--` instead of a stale percentage.

The helper takes `nowSeconds` as a parameter so it's deterministically unit-testable without `vi.useFakeTimers()` or `Date.now()` stubbing.

## Tests

9 new unit tests in `CodexUsageService.expiry.test.ts`:

- Both windows active - returns unchanged
- Primary expired, secondary active - drops primary only, surfaces secondary
- Primary active, secondary expired - drops secondary only, surfaces primary
- AnisminC #120 scenario: stale primary at 91%, no secondary - returns null so the scan continues
- Both windows expired - returns null
- Boundary: `resets_at === now` - treated as expired (strict `>` check)
- `resets_at` undefined (old SDK schema gap) - treated as active, fail-open
- Preserves `credits` field on the active side
- Preserves `limit_id` field on the active side

All 9 pass.

## How users with the stuck value see it clear

The fix takes effect on the next `refresh()` call (currently every 30 minutes via the poll loop in `CodexUsageService.startPolling`, or sooner if the user opens the indicator panel). No migration needed - `cachedUsage` is in-memory only.

## Test plan

- [x] `npx vitest run packages/electron/src/main/services/__tests__/CodexUsageService.expiry.test.ts` - 9/9 pass
- [ ] Manual: on a machine with a stuck indicator, verify the next 30-minute poll surfaces the correct value (or `--` if no active window exists)

Fixes #120.


ademczuk:
Auto-closed during today's fork-relationship reset on `ademczuk/nimbalyst` (sweep of all 7 of my open PRs). Resubmitted at #312 with identical branch contents. No code change.

Issue #238, opened 2026-05-11
App freezes with multiple running agent sessions; subsequent relaunch fails with "Request init timed out"


shayliraz:
## Summary
Nimbalyst becomes unresponsive while ~3 agent sessions are running, and after a forced close the next launch fails with a database initialization timeout.

## Steps to reproduce
1. Have ~3 sessions running concurrently (agent runs + planning files open).
2. Continue working / let the agents run.
3. The UI freezes — the "work in progress" spinner stalls and nothing responds to clicks.
4. Wait. Windows eventually surfaces its "This app isn't responding — Close the program / Wait" dialog.
5. Force-close via Windows and relaunch Nimbalyst.

## Expected behavior
- The app stays responsive while multiple agent sessions run.
- If something does hang, relaunching restores a working app on the first try.

## Actual behavior
- UI spinner is stuck, the whole window stops responding, Windows flags it as not responding.
- On the first relaunch after the freeze, a modal appears: "Nimbalyst — Database Initialization Failed / Failed to initialize the database system. / Error: Request init timed out / Nimbalyst cannot continue without the database."
- A second relaunch starts successfully.
- This freeze has happened more than once ("again").

## Environment
- App version: 0.59.2
- OS: Windows 11 (10.0.26200)
- Arch: x64
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
Not collected — log-gathering tools were not available in this session.

karlwirth:
We have one more general preformance fix coming in the next release along with more logging to catch this.

If this happened while you were using worktrees, there is an additional performance fix in the worktree area that is coming in the next release as well.

ademczuk:
**The "Request init timed out" string is a 30-second IPC timeout between the main thread and the PGLite worker thread (`PGLiteDatabaseWorker.ts:658`), NOT a lock-acquisition timeout.** Causes are: worker thread failed to start, worker crashed silently during PGLite init, or PGLite init took longer than 30 seconds.

[...]

PR #300, opened 2026-05-14
fix(database): extend PGLite worker init timeout to 120s to survive WAL recovery on first relaunch after force-close (#238)

ademczuk:
## Summary

Partial fix for #238. Extends the PGLite worker `init` IPC timeout from 30s to 120s so the first relaunch after a force-close does not raise a `Database Initialization Failed` modal while WAL recovery is still in progress.

This is the recovery-symptom fix, not the freeze-itself fix. The underlying freeze when running 3+ concurrent agent sessions is a separate, larger investigation and remains open.

## Root cause of the modal

`PGLiteDatabaseWorker.sendMessage` (`packages/electron/src/main/database/PGLiteDatabaseWorker.ts:639`) defaults to a 30 000 ms timeout for every worker message:

```ts
private sendMessage(type: string, payload?: any, timeoutMs: number = 30000): Promise<any> { ... }
```

For normal `query` / `exec` messages that's fine. But `init` (line 372) does substantially more work:

- Acquire the `nimbalyst-db.pid` lock with PID-liveness checking
- Clean up stale `postmaster.pid` artifacts
- Open PGLite, which runs WAL replay after an unclean shutdown
- Run the schema migration pipeline
- If corruption is detected, run the auto-recovery flow (lines 374-490)

After the reporter's freeze + force-close, the first relaunch hit `Request init timed out` because PGLite was still mid-recovery at the 30s mark. A second relaunch then consistently succeeded because recovery had completed in the background between attempts.

## Fix

Add a module-level `INIT_TIMEOUT_MS = 120_000` constant and pass it to both init call sites:

- `doInitialize` (line 372) - cold-start init
- `recreateWorkerAndReinit` (line 243) - reinit after backup-restore recovery (where the just-restored DB also needs to replay WAL on first open)

The query/exec/getStats/verifyBackup timeouts are unchanged.

Two minutes covers the recovery window the reporter saw while still surfacing a genuinely-deadlocked init within a bounded wait rather than spinning forever.

## Why not 180s or 60s?

- **60s** would have caught the reporter's scenario based on "second relaunch succeeded immediately" suggesting recovery is around 30-50s, but offers no margin for slower disks or larger workspaces.
- **180s** is fine but adds 60s to the worst-case UX wait when init is genuinely deadlocked (worker bug, OS lockup). 120s is the safest middle ground.

Open to bumping further if signal comes in that 120s still trips on production hardware.

## What's NOT in this PR

- The freeze with 3+ concurrent agent sessions itself. That likely lives in the main-process event loop or the worker's serialized write queue, and needs heap profiling under load.
- Progress reporting from the worker to the main process during init. That would let us show a "Recovering database..." spinner instead of a silent 120s wait, but requires worker-protocol changes.
- A graceful retry on timeout. The lock-file plumbing makes that non-trivial because a retry would deadlock on its own PID. Worth exploring as a follow-up.

## Test plan

- [ ] Manual: simulate slow init by adding `await sleep(45_000)` to `worker.js` initialize() and verify the modal no longer fires; remove the artificial sleep before merging.
- [ ] Manual: confirm a genuinely-deadlocked init (e.g. add `await new Promise(() => {})` to worker init) still surfaces an error within 120s rather than hanging forever.

No automated test added: the change is a constant bump, and the meaningful verification is end-to-end behavior under unclean-shutdown conditions which is environment-dependent. Happy to add a `toHaveBeenCalledWith(120_000)` assertion on `sendMessage` if maintainers prefer.

Partial fix for #238.


ademczuk:
Auto-closed during today's fork-relationship reset on `ademczuk/nimbalyst` (sweep of all 7 of my open PRs). Resubmitted at #317 with identical branch contents. No code change.

Issue #304, opened 2026-05-14
@ mention with empty query shows alphabetical files instead of recently viewed


karlwirth:
## Summary
Typing `@` alone in a markdown document shows files in alphabetical order instead of the recently viewed documents it's supposed to show.

## Steps to reproduce
1. Open a markdown document.
2. Type `@` and stop (don't type anything after it).
3. Observe the file picker that appears.

## Expected behavior
The picker should show recently viewed documents when the query is empty.

## Actual behavior
The picker shows all files in alphabetical order. Recently viewed documents are not surfaced.

## Environment
- App version: 0.60.1
- OS: darwin 25.2.0 (arm64)
- Electron: 41.0.4
- Chrome: 146.0.7680.153

## Logs
Not collected (production build; no renderer debug logs available, no relevant errors in main process log).

ademczuk:
FYI: the Windows-separator + workspace-scope fix landed via PR #318 (merged 2026-05-15, 70a412ef). PR body included `Closes #301 and #304` but GitHub's auto-close only fired on #301 - likely a parser quirk on the `and #304` form. Flagging for manual close on the next triage pass.

ghinkle:
Closing as duplicate of #303 (@-mention picker showing incomplete recently-viewed files). Both reports are about the same picker ordering/listing behavior — consolidating there.

PR #311, opened 2026-05-15
fix(quick-open): scope recent files to current workspace + Windows-safe prefix check (#301, #304)

### LINKED ISSUE #301: Cmd+O Quick Open includes files from other pinned workspaces in left rail
## Summary
Cmd+O Quick Open shows files from every workspace pinned in the left multi-project rail, instead of being scoped to the current workspace.

## Steps to reproduce
1. Pin two or more workspaces in the left multi-project rail.
2. Focus one workspace.
3. Press Cmd+O to open Quick Open.
4. Observe the file list.

## Expected behavior
Quick Open lists only files from the currently focused workspace.

## Actual behavior
Quick Open lists files from all workspaces pinned in the left rail, mixed in with the current workspace's files. Only the pinned workspaces are leaking — not every workspace ever opened.

## Regression
Started after the left-rail multi-project tool was added.

## Environment
- App version: 0.60.1
- Platform: darwin (arm64)
- OS release: 25.2.0
- Electron: 41.0.4

## Logs
Not collected (packaged build — renderer debug logs unavailable).

ademczuk:
## Summary

Fixes both #301 (Cmd+O Quick Open includes files from other pinned workspaces) and #304 (@-mention with empty query shows alphabetical instead of recently viewed). Same root cause for both, plus a Windows-specific path-separator bug for #304.

## Root cause

The `getRecentWorkspaceFiles` IPC took no parameter. The main handler (`WorkspaceHandlers.ts:646`) resolved the workspace from window state via `BrowserWindow.fromWebContents`. With the multi-project rail (#188), a single window can have several workspaces pinned, so the window-state `workspacePath` can lag behind which workspace the renderer caller actually cares about. Recents came back from a stale workspace.

The @-mention empty-query path in `fileMention.ts:218` also filtered the returned list with:

```ts
const inWorkspace = recent.filter(p => p.startsWith(workspacePath + '/'));
```

On Windows, where absolute paths use `\` as the separator, the filter matched zero entries. The empty-query path then fell through to the ripgrep search, which on empty input returns alphabetical results - exactly the symptom @karlwirth reported in #304.

## Fix

- `getRecentWorkspaceFiles(workspacePath?: string)` now accepts an explicit parameter through the preload + the main handler.
- Renderer call sites in `QuickOpen.tsx:333` and `fileMention.ts:216` pass the current `workspacePath`.
- The handler uses the parameter when present and keeps the window-state fallback for backwards compatibility with any older renderer bundle that may still hit this channel without the parameter.
- Extract `isPathInsideWorkspace` into a tiny `workspacePathPrefix.ts` module. Accepts either `/` or `\` after the workspace root, so the filter works on POSIX and Windows. Guards against sibling-prefix collisions (e.g. `/repos/app` must not match `/repos/app-archive`).

## Tests

7 unit cases in `fileMention.pathPrefix.test.ts`:
- POSIX path inside workspace
- Windows path inside workspace (regression for #304)
- Sibling-prefix collision on POSIX (`/repos/app-archive` vs `/repos/app`)
- Sibling-prefix collision on Windows
- Unrelated workspace
- Workspace root equals filePath (edge)
- Trailing-separator workspace path (defensive)

All 7 pass. TypeScript check clean for `packages/electron`.

## Why the helper lives in its own module

The natural place would be at the top of `fileMention.ts`, but the test import chain through `fileMention.ts` transitively loads Lexical extension registration (`@nimbalyst/runtime` then editor extensions then `defineExtension`) which fails outside a Lexical-aware test setup with `TypeError: (0 , defineExtension) is not a function`. Pulling the helper into `workspacePathPrefix.ts` avoids the import chain entirely.

## How to verify

On a Mac with two workspaces pinned in the left rail:

1. Focus workspace A. Open a file specific to A.
2. Focus workspace B. Open a file specific to B.
3. Focus workspace A. Press Cmd+O. The list should NOT include the workspace-B file.

On Windows (#304):

1. In an active document, type `@` and stop. The picker should show recently viewed files, not the alphabetical top-level listing.

## Test plan

- [x] `npx vitest run packages/electron/src/renderer/store/atoms/__tests__/fileMention.pathPrefix.test.ts` (7/7 pass)
- [ ] Manual: verify on macOS multi-rail setup (Karl's repro)
- [ ] Manual: verify on Windows (@-mention empty query)

Fixes #301, #304.


ademczuk:
Auto-closed 78 seconds after creation during today's fork-relationship reset on `ademczuk/nimbalyst` (sweep of all my open PRs). Resubmitted at #318 with identical branch contents. No code change.

PR #313, opened 2026-05-15
fix(onboarding): bound onboarding:get IPC with 3s timeout to prevent picker hang (#260)

### LINKED ISSUE #260: Beach ball / hang on developer-vs-standard mode picker in initial setup (cannot complete onboarding)

ademczuk:
## Summary

Test coverage for the renderer-side `onboarding:get` IPC-stall fix that @ghinkle landed in v0.60.2 (commit `46240865c`, merge `fd01ec13`). Originally PR #291, then #313 with the source fix proposed; @ghinkle preferred his own implementation, so this PR is rebased down to just the test file.

## What's in the PR

`UnifiedOnboarding.ipcTimeout.test.tsx` - 3 vitest cases:
1. The race-against-3s-timeout path falls back to new-user / show-dialog when `onboarding:get` never resolves
2. The success path respects the returned `userRole` when the IPC resolves before the timeout
3. The new-user-fallback path renders the dialog when the IPC rejects

All 3 pass against Greg's v0.60.2 implementation locally - external behaviour matches the proposed fix.

Closes #260 (test-coverage portion; the source fix is already in upstream).

### DIFF
--- a/packages/electron/src/renderer/components/UnifiedOnboarding/__tests__/UnifiedOnboarding.ipcTimeout.test.tsx
+++ b/packages/electron/src/renderer/components/UnifiedOnboarding/__tests__/UnifiedOnboarding.ipcTimeout.test.tsx
@@ -1,0 +1,102 @@
+/**
+ * Regression coverage for nimbalyst#260.
+ *
+ * karlwirth's reporter (a second user, not karl himself) hit a hang on the
+ * developer-vs-standard mode picker during initial setup - beach ball, no
+ * mode card highlighted, no way to proceed without force-quit. Root cause:
+ * `onboarding:get` IPC issued by this component on dialog open never
+ * resolved (main-process handler stalled on a dynamic store import on cold
+ * start, possibly racing with a second concurrent call from useOnboarding's
+ * own gate). The component awaited the IPC unconditionally with no timeout,
+ * so React kept rendering the initial "no mode selected" state forever.
+ *
+ * Fix: race the IPC against a 3-second timeout. On timeout, fall back to
+ * the new-user path (the conservative default) so the user can pick a mode
+ * and proceed. The dialog remains interactive even if the main process is
+ * wedged.
+ */
+
+import React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, act } from '@testing-library/react';
+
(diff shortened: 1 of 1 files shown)


ghinkle:
Thanks for this. Sorry, but yesterday I landed 46240865c which applies the same renderer-side timeout (plus a few extra breadcrumbs and the dynamic-import cleanup in the main process). 

Your test coverage here is better though. Would you be willing to rebase the PR down to UnifiedOnboarding.ipcTimeout.test.tsx? I think the test should work against my version without changes (same external behavior).

ademczuk:
Done @ghinkle - rebased #313 down to just `UnifiedOnboarding.ipcTimeout.test.tsx` at `b27212b0`. The 3 cases pass against your v0.60.2 implementation unchanged (verified locally: race-on-timeout path, success-with-userRole path, new-user-fallback path - all match the external behaviour you landed).

PR body also updated to reflect that this is now a test-coverage-only PR for the fix you already shipped at `46240865c`. Ready when you are.

Quick aside: thanks for the merge wave on the rest of the queue today. The new fork plumbing (`ademczuk/nimbalyst-1`) is settled and the remaining 7 of my open PRs all landed cleanly through it, so the fork-detach hiccup from this morning is fully recovered.

PR #316, opened 2026-05-15
fix(database): PGLite lock self-heals on Windows after PID reuse (#272)

ademczuk:
## Summary

Resubmit of #294 (closed without merge during a GitHub fork-relationship reset earlier today; all 7 of my open nimbalyst PRs were swept up). Branch contents unchanged from #294. Partially addresses #272.

---

The PGLite lock file now self-heals after a main-process crash on Windows when the OS has reused the original lock holder's PID. `worker.js`'s lock liveness check used `process.kill(pid, 0)` and interpreted any `EPERM` error as "process still running". On Windows, `process.kill(pid, 0)` against a PID owned by another user or privilege level throws `EPERM` regardless of whether that PID is the original Nimbalyst process or an unrelated system process that recycled the number.

AnisminC's report on #272: Nimbalyst PID 6732 crashed at 21:09Z, the user relaunched at 00:56Z (3h47m later, no reboot), Windows had recycled PID 6732 onto a system process, the EPERM-as-alive interpretation blocked the launch with `DATABASE_LOCKED`, and the only workaround was deleting the lock file manually.

The lock liveness decision is now extracted to a pure `decideLockIsRunning` helper that adds a 60-second grace window inside the EPERM branch: a genuine competing Nimbalyst instance acquires its lock during the current launch attempt (always seconds-fresh), while a crash-stale lock has a timestamp from before the crash (minutes to hours old) and is correctly classified stale.

## Tests

8 unit cases in `lockStaleness.test.ts` cover clean ESRCH, clean alive, AnisminC's exact 3h47m stale scenario (13571s old), unknown-timestamp fallback, EPERM+fresh timestamp (live sibling), custom grace, unknown error code, exported constant. Plus `lockStaleness.d.ts` declarations so the TS test file resolves the CJS helper.

Partially addresses #272.


ghinkle:
Going to merge this back... We can see how it goes, but maybe we do that user dialog feature down the road.

ademczuk:
Thanks for the merge @ghinkle.

Agreed on the user-dialog as a "wait-and-see" follow-up. If the 60s grace turns out to under-cover on slow Windows hardware (e.g. machines where the lock can legitimately be written more than 60s before init), the dialog gives us a safety valve for the ambiguous middle case without abandoning the auto-recover path for the unambiguous-stale case.

Happy to draft that PR if reports come in. Worth tagging the future ticket with the auto-recover bypass path so the dialog only fires when the grace window says "ambiguous" rather than on every launch.

PR #325, opened 2026-05-15
fix(database): ambiguous-lock branch with user dialog (Greg's #316 review follow-up)

ademczuk:
## Summary

Draft follow-up to the closed PR #316. @ghinkle suggested adding a user dialog for the ambiguous EPERM-with-fresh-timestamp case rather than guessing. This PR ships that dialog plus the supporting ternary decision surface in `decideLockIsRunning`.

## Background

`decideLockIsRunning` previously returned a binary `isRunning` flag. The EPERM branch with a fresh timestamp (lock written less than 60s ago, lock holder PID unsignalable from us) was treated as a live sibling and the launch was refused with `DATABASE_LOCKED`. On a slow-disk machine where the original Nimbalyst wrote the lock less than 60s before crashing, that produced a false-positive that the user could only clear by deleting `nimbalyst-db.pid` manually.

## Changes

**`lockStaleness.js`** now returns a ternary `decision: 'running' | 'stale' | 'ambiguous'` plus `lockPid` and `lockAgeMs` for dialog rendering. The legacy `isRunning` boolean stays on the return shape for backwards compatibility - true for both 'running' and 'ambiguous'.

**`worker.js`** raises a distinct `DATABASE_LOCKED_AMBIGUOUS` error code for the ambiguous branch and stashes `lockFilePath`, `lockTimestamp`, `lockHostname` on the error so the dialog can render those values.

**`PGLiteDatabaseWorker.doInitialize`** catches the new error code BEFORE the existing `DATABASE_LOCKED` substring check (which would otherwise match - "DATABASE_LOCKED_AMBIGUOUS" contains "DATABASE_LOCKED"). Shows a question dialog with two scenarios spelled out:

> Cannot tell whether another Nimbalyst is already running.
>
> Nimbalyst found a database lock from a few seconds ago and cannot confirm whether the process holding it (PID ..., host ..., acquired ...) is still alive. Two scenarios are equally likely:
>
>   1. Another Nimbalyst window is open under a different user account or privilege level. Opening anyway will run two instances against the same database and may corrupt data.
>
>   2. A previous Nimbalyst crashed less than a minute ago and the OS has already reused the original PID for a system process. In this case the lock is safe to clear.
>
> If unsure, choose Cancel and look for another Nimbalyst window before retrying.
>
> [Cancel] [Open Anyway (clear lock)]

Cancel is the default. Open Anyway removes the lock file, calls `recreateWorkerAndReinit` (which uses the 120s `INIT_TIMEOUT_MS` from PR #317 / #238 for the post-unlock WAL recovery path), and continues.

**Playwright branch** explicitly refuses to force-unlock so test setup races stay visible.

**Analytics** fire on both buttons (`database_lock_ambiguous_cancel` / `database_lock_ambiguous_force_unlock`) so we can measure how often the ambiguous case actually shows up in practice.

## Tests

8 unit cases in `lockStaleness.test.ts` still pass. The EPERM-fresh-timestamp test now asserts `decision === 'ambiguous'` plus the new surfaced `lockPid` and `lockAgeMs` fields. Other branches unchanged.

## Why draft

@ghinkle's framing on the closed #316 was "we can see how it goes, but maybe we do that user dialog feature down the road." Marking this as draft so the dialog copy, the analytics names, and the Playwright behaviour can be reviewed before merge.

Two specific calls I want your input on:

1. Default button is Cancel. Argument for: data corruption from running two instances is worse than the inconvenience of investigating before clicking Open Anyway. Argument against: most users will not have a second instance and will hit this dialog from a crash; defaulting to Cancel adds friction for the common case. Happy to flip if you prefer.
2. The "Open Anyway" path uses `recreateWorkerAndReinit`, which re-fires init with the 120s `INIT_TIMEOUT_MS`. If the lock was a false positive AND the previous Nimbalyst had also corrupted the DB on crash, the corruption-recovery path fires inside that retry. The user sees: dialog -> click -> 30-90s wait -> "Database Recovered" notice. That seems fine but worth flagging.

Per-Greg-review-followup-to closed PR #316. Builds on the merged #316 (PGLite lock self-heal).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
+- PGLite lock-staleness check now surfaces an "ambiguous" branch and asks the user via dialog instead of guessing. Follow-up to the closed PR #316: previously the EPERM-with-fresh-timestamp case (lock < 60s old, lock holder PID unsignalable from us) was treated as a live sibling and the launch was refused. On a slow-disk machine where the original Nimbalyst wrote the lock less than 60s before crashing, that produced a false-positive DATABASE_LOCKED that the user could only clear by deleting the lock file from the filesystem. `decideLockIsRunning` now returns a ternary `decision: 'running' | 'stale' | 'ambiguous'` and exposes `lockPid` / `lockAgeMs` for dialog rendering. The 'running' and 'stale' branches behave as before. The 'ambiguous' branch raises a distinct `DATABASE_LOCKED_AMBIGUOUS` error code; `PGLiteDatabaseWorker` catches it and shows a dialog explaining the two scenarios (live sibling vs fast PID reuse) with "Cancel" and "Open Anyway (clear lock)" buttons. The legacy `isRunning` boolean stays on the return shape for backwards compatibility (true for both 'running' and 'ambiguous'). Per @ghinkle's review on the closed PR #316. (#272 follow-up)
 <!-- Bug fixes go here -->
 
 ### Removed
--- a/packages/electron/src/main/database/PGLiteDatabaseWorker.ts
+++ b/packages/electron/src/main/database/PGLiteDatabaseWorker.ts
@@ -546,6 +546,76 @@ export class PGLiteDatabaseWorker {
       logger.main.error('[PGLite Worker] Failed to initialize:', error);
       this.initPromise = null;
 
+      // Check for the AMBIGUOUS branch FIRST (its error message also contains
+      // the string "DATABASE_LOCKED" so the simpler check below would match
+      // it otherwise). The ambiguous branch fires when worker.js could not
+      // signal the lock holder via `kill(0)` (EPERM) AND the lock timestamp
+      // is fresh enough that we cannot rule out a real sibling instance.
+      // Per @ghinkle's review on the closed PR #316: ask the user instead of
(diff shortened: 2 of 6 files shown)


ademczuk:
Rebased onto current main (was 37 behind, conflict was CHANGELOG-only; the worker.js and lockStaleness changes auto-merged on top of your merged #316). Marking ready for review once CI clears. lockStaleness test suite green locally (8/8).

This implements the ambiguous-case dialog you suggested on #316 ("add a user dialog for the ambiguous EPERM-with-fresh-timestamp case rather than guessing"). `decideLockIsRunning` now returns a ternary `decision: 'running' | 'stale' | 'ambiguous'` plus `lockPid` / `lockAgeMs`; the legacy `isRunning` boolean stays on the return shape (true for both 'running' and 'ambiguous') so nothing downstream breaks. The 'ambiguous' branch raises `DATABASE_LOCKED_AMBIGUOUS`, which `PGLiteDatabaseWorker` catches and surfaces via the same native `dialog.showMessageBox` pattern the worker already uses for its recovery and start-fresh dialogs. Force-unlock path removes the lock file, recreates the worker, and retries init (covered by the #238 INIT_TIMEOUT_MS recovery lesson). There is a Playwright guard that refuses to force-unlock under test so a real dual-instance race in test setup can't be masked.

The dialog copy is the part most open to your call: the two-scenario explanation (live sibling vs fast PID reuse) and the button labels ("Cancel" default / "Open Anyway (clear lock)"). Happy to reword if you'd prefer different framing or a different default.

PR #332, opened 2026-05-15
fix(claude-code): add tracker_* MCP tools to INTERNAL_MCP_TOOLS so the kanban board works (#236)

### LINKED ISSUE #236: [Bug]: Tracker (kanban board) stopped working
### Describe the bug

When I ask to add a new issue or card on the board, nymbalist is not able to do so. When I ask why, this is the answer:

> These were the returned errors on each attempt:
> `user cancelled MCP tool call`
>That same error came back for:

> `nimbalyst-session-naming.update_session_meta`
> `nimbalyst-mcp.tracker_create`
> `nimbalyst-mcp.tracker_list`
> `nimbalyst-mcp.tracker_list_types`
> So the failures are happening before the tracker action runs.

This was working before and I have the MCP enabled on settings. 

### Nimbalyst version

0.59.2

### Operating system

macOS

ademczuk:
## Summary

Closes #236. Six tracker MCP tool names were missing from `INTERNAL_MCP_TOOLS` in `packages/runtime/src/ai/server/providers/claudeCode/toolPolicy.ts`, so the SDK's `canUseTool` callback fell through to the dialog handler, the Promise never resolved, and every kanban interaction surfaced as `user cancelled MCP tool call`.

## Root cause

`resolveImmediateToolDecision` (`packages/runtime/src/ai/server/providers/claudeCode/immediateToolDecision.ts:44`) auto-allows MCP tools whose full name is in `INTERNAL_MCP_TOOLS`. The other nimbalyst-owned MCP tools (`display_to_user`, `capture_editor_screenshot`, the developer git tools, session-context tools, `update_session_meta`) all have entries in that list and work as expected. The six tracker tools registered by the in-process `nimbalyst-mcp` server in `packages/electron/src/main/mcp/tools/trackerToolHandlers.ts` did not:

- `tracker_list`
- `tracker_get`
- `tracker_list_types`
- `tracker_create`
- `tracker_update`
- `tracker_link_session`

When the agent calls any of them, the SDK invokes `canUseTool`, `resolveImmediateToolDecision` does not find the name in `INTERNAL_MCP_TOOLS`, control falls through to `handleToolPermissionWithService`, which has no UI wired for nimbalyst-owned tools, the Promise never resolves, and the SDK surfaces `user cancelled MCP tool call`. The kanban board looks completely broken because every interaction needs at least `tracker_list` to render state.

## Fix

Append the six tracker tool names to `INTERNAL_MCP_TOOLS` with an explanatory comment block. No logic change, no new exports, no new files. The fix is strictly additive and only affects tools prefixed `mcp__nimbalyst-mcp__tracker_*`, which are controlled by nimbalyst's own MCP server startup, so third-party MCP tools still fall through to the dialog handler.

## Tests

Six new regression tests in `immediateToolDecision.test.ts` exercise the real `INTERNAL_MCP_TOOLS` list (not the controlled stub) against each tracker tool:

- `mcp__nimbalyst-mcp__tracker_list` is auto-allowed
- `mcp__nimbalyst-mcp__tracker_get` is auto-allowed
- `mcp__nimbalyst-mcp__tracker_list_types` is auto-allowed
- `mcp__nimbalyst-mcp__tracker_create` is auto-allowed
- `mcp__nimbalyst-mcp__tracker_update` is auto-allowed
- `mcp__nimbalyst-mcp__tracker_link_session` is auto-allowed

Plus a guard test verifying that a fabricated third-party MCP tool name (`mcp__some-third-party-server__do_something`) still falls through to `null`, plus a regression-guard test that the prior internal tools (`update_session_meta`, `display_to_user`, `capture_editor_screenshot`) are still in the list.

Test file grows from 14 to 20 tests, all green.

## Why not also add `automations.create`?

There is a reference to `automations.create` in `trackerToolHandlers.ts:1388` as an error message string ("Use the automations.create tool instead"), but no actual MCP handler with that name is registered in the codebase. If `automations.create` lands as a real tool later, it would need its own entry in `INTERNAL_MCP_TOOLS`. Out of scope for this PR.

## Test plan

- [x] 20/20 tests pass locally (`npx vitest run packages/runtime/src/ai/server/providers/claudeCode/__tests__/immediateToolDecision.test.ts`)
- [x] `npx tsc --noEmit` clean for the touched files (pre-existing lexical type errors elsewhere in `packages/runtime` are upstream and unrelated)
- [ ] Manual smoke: ask the agent to create a tracker item via the kanban board, verify no `user cancelled MCP tool call` error and the item appears
- [ ] CI green

## Files

```
packages/runtime/src/ai/server/providers/claudeCode/toolPolicy.ts                                    (+13 lines)
packages/runtime/src/ai/server/providers/claudeCode/__tests__/immediateToolDecision.test.ts          (+53 lines)
CHANGELOG.md                                                                                         (+1 entry)
```

Closes #236


ghinkle:
Was there a reason you didn't include all of the tracker tools? 
Specifically: tracker_define_type, tracker_delete_type, tracker_unlink_session, tracker_link_file, tracker_add_comment

ademczuk:
Good catch, sorry, as I scoped to the reporter's list (6 tools mentioned in #236) rather than running a full `grep -E "tracker_[a-z_]+"` on the canonical set. The five you flagged are exactly the missing ones.

Just pushed a follow-up commit (rebased onto current main, so the upstream `TrackerSchemaService.watch.test.ts` CI failure should also clear). The PR now covers all 11 `tracker_*` tools:

- `tracker_list`, `tracker_get`, `tracker_list_types`
- `tracker_create`, `tracker_update`
- `tracker_link_session`, `tracker_unlink_session`, `tracker_link_file`
- `tracker_add_comment`
- `tracker_define_type`, `tracker_delete_type`

The CHANGELOG entry and the regression test parameterization were both updated to match. Test count goes from 6 to 11 (one parameterized assertion per tool, plus the existing third-party-fall-through guard).

ademczuk:
One sanity-check note while you have this open: of the 11 tracker_* tools the PR auto-allows, `tracker_delete_type` is the most destructive. It can cascade to every tracker item that references the type. `tracker_unlink_session` is the next-sharpest.

The existing 6 tracker entries (`tracker_create` / `tracker_update`) already put writes under auto-allow, so the policy direction is consistent. `delete_type` is just a sharper edge than the others.

If you'd rather gate the destructive two behind the permission dialog, I can split them out into a separate "requires-prompt" set in a follow-up commit. Equally fine to ship as-is if the "all tracker ops are internal" framing is the design call you want. Flagging in case you want to weigh in before merge.

However, I'm confident that seeing as tracker_delete_type has a usage-guard that caps the **blast radius** (refuses if items reference the type) that we wont need to carve out a commit.

ghinkle:
I may end up putting the tracker stuff into its own mcp as it has grown more than I had originally planned, but that'll be future work. merging

PR #335, opened 2026-05-16
fix(editor): per-row size guard in markdown table transformer to prevent V8 heap crash on wide tables (#321)

### LINKED ISSUE #321: Crash-on-load loop triggered by large markdown file after successful write
## Summary
Nimbalyst enters a crash-restart-crash loop when reloading open files on startup, triggered by a large markdown file that was successfully written by the Claude agent immediately before the initial crash.

## Steps to reproduce
1. Have a large markdown file open in Nimbalyst (e.g. a multi-hundred-row markdown table).
2. Use the Claude agent to make an edit to the file — the write completes successfully and the agent reports completion.
3. Nimbalyst crashes shortly after the write, before any further user action.
4. On restart, Nimbalyst attempts to reload the same file during startup.
5. The app crashes again during the load phase (after a noticeable lag), re-entering step 4.

## Expected behavior
Nimbalyst restarts cleanly and reloads the file without crashing, or gracefully skips/quarantines the problematic file and opens without it.

## Actual behavior
Nimbalyst crashes during startup file-load every time, creating an unrecoverable loop. The only known workaround is to manually prune the offending markdown file from outside the app, after which Nimbalyst loads normally.

This has occurred twice with different (but similarly large) markdown files.

## Environment
- App version: 0.60.1
- OS: macOS 26.5 (arm64, Mac16,11)
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
macOS Crash Reporter shows 5 crash cycles between 8:51 AM and 9:09 AM on 2026-05-15, each a simultaneous Nimbalyst + Nimbalyst Helper (Renderer) crash approximately 2-3 minutes apart — consistent with an auto-restart loop that crashes every time on file load.

Crash pairs (from macOS Crash Reporter):
- 8:51 AM - Nimbalyst + Nimbalyst Helper (Renderer)
- 9:03 AM - Nimbalyst + Nimbalyst Helper (Renderer)
- 9:05 AM - Nimbalyst + Nimbalyst Helper (Renderer)
- 9:06 AM - Nimbalyst + Nimbalyst Helper (Renderer)
- 9:09 AM - Nimbalyst + Nimbalyst Helper (Renderer) ← crash dumps below

Crash dumps from the 9:09 AM cycle. ~90 seconds of uptime before crash. Renderer stack shows `LowMemoryNotification()` followed by 11+ levels of deep recursion, consistent with memory exhaustion loading a large file.

Incident IDs (may help locate in crash reporting system):
- Main process: FD76EB4B-BEAD-4EA3-849E-E057A2F9B1A4
- Renderer: 9F5C347D-023A-49CE-8F62-2EF164FE900D

**Main process (Thread 0, CrBrowserMain):**
```
Exception: EXC_BREAKPOINT (SIGTRAP)
0  Electron Framework  v8::Object::SetIntegrityLevel + 25252
...
35 Electron Framework  ElectronMain + 120
```

**Renderer (Thread 0, CrRendererMain):**
```
Exception: EXC_BREAKPOINT (SIGTRAP)
9  Electron Framework  v8::Isolate::LowMemoryNotification() + 860
[DEEP RECURSION: 11 levels through unknown frame, 5 levels elided]
39 Electron Framework  v8::Function::NewInstanceWithSideEffectType + 2204
...
61 Electron Framework  ElectronMain + 120
```

Reported via in-app feedback per conversation with Karl.

ademczuk:
## Summary

Closes #321. @aawbeck reported a crash-on-load loop: open a markdown file with a table containing many large cells, the app crashes, the next relaunch re-opens the same file (the `restoreTabs: true` default), and the loop repeats until the user manually deletes the file. Reporter shared the macOS crash log and a 442KB repro file.

## Root cause

`TABLE_TRANSFORMER` in `packages/runtime/src/editor/plugins/TablePlugin/TableTransformer.ts` calls `$createTableCell` per cell. `$createTableCell` recursively calls `$convertFromEnhancedMarkdownString` on the cell's text content. The full call chain runs on the main thread inside a single Lexical `editor.update()` transaction.

Reporter's repro file: 83 data rows of ~5,451 chars each (5 cells of ~1,090 chars per row). That is 415 recursive cell parses, each parsing ~1KB of text, each allocating Lexical nodes synchronously in one transaction.

The macOS crash dump shows:

- Main process, `CrBrowserMain` thread
- `EXC_BREAKPOINT (SIGTRAP)` - a V8 integrity-level assertion (NOT OOM)
- Top of stack: `v8::Object::SetIntegrityLevel`, three frames of `CompilationDependencies::DependOnArraySpeciesProtector()` (recursive JIT compilation), then `temporal_rs_PlainTime_second` (V8 TurboFan compiling a hot function under memory pressure)
- The earlier renderer crash shows `LowMemoryNotification()` + 11 levels of deep recursion

Net: cumulative synchronous Lexical node allocation across 415 cells exhausts the V8 heap and trips the integrity-level assertion in TurboFan during a concurrent JIT compilation.

## Fix

Per-row size guard. `MAX_TABLE_ROW_BYTES = 5_000` lives in a new `tableRowSizeGuard.ts` file. `mapToTableCells` now returns `null` for rows whose text content exceeds the threshold, so the row falls through to plain markdown text rendering instead of triggering the recursive cell parse.

The threshold sits:

- above legitimate wide tables (a 10-column table with 400-char cells is ~4,400 chars including pipes and spaces)
- below the reporter's repro (5,451 chars per row)

so it fires on pathological input without affecting normal use. Oversized rows render as piped text - no crash, file opens, content remains visible.

## Why a separate file for the guard

`TableTransformer.ts` transitively imports from `../../markdown/` which has a pre-existing `@lexical/extension` resolution failure (visible in `npx tsc --noEmit` and unrelated to this PR). Vitest cannot route around the broken import. Extracting the pure check into `tableRowSizeGuard.ts` keeps the unit test loadable without the broken markdown chain. The threshold and the helper have no other dependencies, so the extraction does not add architectural complexity.

## Limitation (acknowledged)

This is a per-row guard. A table built from many narrow-but-numerous rows could still pressure heap in aggregate (e.g. 1,000 rows of 1KB each = 1MB cumulative parse work, same root mechanism). The residual class is NOT addressed by this fix. The right path for that is either:

- a cumulative-bytes guard across the `while` loop that builds the table, or
- moving the recursive cell parse off the main thread

Both are out of scope here. This PR closes the reported crash and unblocks the file-open path. Happy to extend with a cumulative-bytes guard in a follow-up if you want broader coverage.

## Tests

9 unit cases in `TableTransformer.rowSizeGuard.test.ts`:

- threshold bounds (above legitimate-wide-table 4,400 chars; below #321 repro 5,451 chars)
- legitimate row sizes pass: empty, tiny, exactly-at-threshold, 10-col x 400-char
- oversized rows blocked: threshold + 1, #321 repro shape, 10KB single-cell

## Test plan

- [x] 9/9 unit tests pass locally
- [x] No AI markers in code or comments
- [ ] Manual smoke: open `Large_Markdown_File_For_Testing.md` (the reporter's repro file from #321) and verify no crash; the giant table renders as plain piped text rather than a parsed table
- [ ] Manual smoke: open a normal markdown file with a small table; table still parses and renders correctly
- [ ] CI green

## Files

```
packages/runtime/src/editor/plugins/TablePlugin/tableRowSizeGuard.ts                                (NEW, 49 lines)
packages/runtime/src/editor/plugins/TablePlugin/__tests__/TableTransformer.rowSizeGuard.test.ts     (NEW, 75 lines)
packages/runtime/src/editor/plugins/TablePlugin/TableTransformer.ts                                 (-3 +5 lines)
CHANGELOG.md                                                                                        (+1 entry)
```

Closes #321


ademczuk:
@ghinkle quick heads-up on the CI red - it is not from this PR.

The `Unit Tests` job fails on a single test file:

```
FAIL  packages/electron/src/main/services/__tests__/TrackerSchemaService.watch.test.ts
  > TrackerSchemaService watcher > hot-loads added, edited, and deleted workspace schemas
```

Same test fails on `upstream/main` directly. The CI history shows the regression bisects to your commit `1025c046` "fix: tracker schema reloads and add tracker field url type":

- before `1025c046` (e.g. `f6aa986f`, `0b00d04d`): green
- `1025c046`: red (https://github.com/nimbalyst/nimbalyst/actions/runs/25951105276)
- `9aba63de` (v0.60.3 release): red (https://github.com/nimbalyst/nimbalyst/actions/runs/25951252238)

This PR does not touch `TrackerSchemaService` or the watcher test. The fix on main will unblock the CI status here without needing anything from me. Flagging in case you missed it between the schema commit and the release tag.

If you want me to look into the test itself I can, though I am not sure I have the schema context to land a clean fix. Same situation on the other 3 of my open PRs (#325, #327, #332) - all four were rebased onto the broken main earlier today.

PR #412, opened 2026-05-22
feat(worktree): pick base branch and name when creating worktree

### LINKED ISSUE #264: [Feature] New Worktree button: base branch selector dropdown
## Problem
When multiple worktrees are active with in-progress changes, there's no way to
create a new worktree from a specific base branch (e.g., a clean `main`). The
only option is "New Worktree," which derives from the current state. This makes
it impossible to spin up a fresh worktree mid-flight without it inheriting work
from other active trees.

## Proposed solution
Add a small dropdown arrow to the right of the "New Worktree" button. Clicking
it opens a branch picker where the user can select a base branch for the new
worktree. Local branches appear immediately at the top; a "Load remote branches"
text button loads remotes on demand. The selected branch becomes the base,
equivalent to `git worktree add -b <new-branch> <path> <base>`.

## Alternatives considered
Pre-creating placeholder worktrees from a clean `main` before starting any work,
then using those reserves when a new task comes up. This requires avoiding new
worktree creation entirely until all in-progress work is merged back to `main`.
It's a significant workflow constraint that the proposed feature would eliminate.

## Additional context
Most useful in agent mode when 3+ worktrees are active concurrently and a new
parallel task needs to start from a known-good baseline rather than picking up
any pushed or unpushed changes from sibling trees.

omartelo:
## Summary

- Clicking "New Worktree" now opens a centered modal where the user picks the base branch and optionally names the worktree, instead of always branching off the current HEAD.
- Local branches and cached remotes render immediately; a background `git fetch` refreshes the remote list when it resolves.
- `worktree:create` IPC migrates to an options object accepting `name` and `baseBranch`; both are wired through preload, renderer service, and AgentMode callback.
- Analytics: `worktree_created` gains `base_branch_source` (`default` | `picker`); `POSTHOG_EVENTS.md` updated accordingly.

## Screenshot

image

## Test plan

- [x] Open "New Worktree" modal, verify local branches list renders immediately.
- [x] Confirm cached remote branches appear without waiting for fetch.
- [x] Wait for background `git fetch` and verify remote list refreshes.
- [x] Create worktree with default base (current HEAD) — `worktree_created` event has `base_branch_source: default`.
- [x] Create worktree picking a different base branch — event has `base_branch_source: picker`.
- [x] Create worktree with custom name; verify directory and branch use the chosen name.
- [x] Create worktree without name; verify auto-generated name still works.

Closes #264


ademczuk:
Took a careful pass against #264. The path checks out end to end: `baseBranch` flows from the picker through `worktree:create` into `GitWorktreeService.createWorktreeImpl`, which already had the field and passes it to `git worktree add -b <branch> <path> <base>`. simple-git's `git.branch()` returns remote refs with the `remotes/` prefix in the installed version (3.36.0), so the local/remote split in the picker populates correctly. Arg-array invocation means no shell-injection surface, and the IPC types stay in sync across preload, electron.d.ts, and the renderer service.

Two small optional things, neither blocking:

1. `NAME_PATTERN` accepts a few names git rejects once they become `worktree/<name>`: trailing slash, consecutive dots (`a..b`), `.lock` suffix. Creation fails cleanly in the service, but the modal has already closed by then (onCreate fires before create resolves), so the user loses their input and the error surfaces away from the picker. Tightening the regex, or holding the modal open until create resolves, would smooth that.
2. `git:fetch` on open has no timeout and keeps running if the user cancels mid-fetch. Fine per the #264 "log silently" decision, just noting it.

Nice feature, the remote-on-demand load is a good touch.

omartelo:
Thanks for the careful pass. Addressed both notes in 218444e:

1. `validateName` now rejects trailing slash, `..`, `//`, leading/trailing dot, leading `-`, `@{`, `.lock` suffix (full path + per-segment), git's forbidden refname chars (`~ ^ : ? * [ \`), and control chars — each with its own error message, so bad names are caught before IPC.
2. Picker stays mounted until `worktree:create` resolves: primary button shows "Creating…" with a spinner, controls disable, ESC/overlay cancel are blocked, and server errors surface inline without losing input.
3. Bonus: `git:fetch` + the follow-up `git:branches` are dropped on close (cancel, ESC, overlay, or successful create) via a shared `aliveRef`, so stale responses don't flip state on a closed modal.

Issue #365, opened 2026-05-19
Renderer process killed under multi-session load (PGLite saturation, edit-queue overflow, 100% CPU)


shayliraz:
## Summary
With 2–3 AI sessions running concurrently, Nimbalyst's renderer process dies hard (window vanishes). Pre-crash logs show the WorkspaceFileEditAttribution queue overflowing, PGLite queries blocking for 3+ seconds, write-queue backpressure, and main-process CPU pegged at ~100% for ~30 seconds before the renderer is killed.

## Steps to reproduce
Hard to repro deterministically, but the conditions are:
1. Open a workspace with a moderately large file tree (hundreds of MD files in a `KB/` subtree).
2. Have 2 AI sessions actively working — one editing many files in quick succession (in this case, GCP Cloud Run / sprint files), one doing planning work.
3. Open or interact with a third session in the same window.
4. Within tens of seconds the app freezes briefly and the window vanishes.

## Expected behavior
The app shouldn't hard-close under normal multi-session use. If load really exceeds capacity, the renderer should at least degrade gracefully (pause non-essential watchers, throttle DB writes) rather than die.

## Actual behavior
Window disappears (hard close, not a hang-then-close). On restart the sessions reappear but in-flight work is lost / the user has to re-prompt. Logs show a clean restart immediately after, so the main process appears to have respawned or the user did — but the previous renderer's death is logged only as `Renderer process gone: [object Object]` with no useful reason string.

I filed a related "Debug parallel session crash issue" two days ago (2026-05-17). This is recurring.

## Environment
- App version: 0.60.1
- OS: Windows 11 (10.0.26200), x64
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
Pre-crash slice (2026-05-19, ~42 s leading up to renderer death):

```
[10:16:05.048] [WorkspaceFileEditAttributionService] Queue full, dropping oldest event
               { workspacePath: '<REDACTED_WORKSPACE>', filePath: '<REDACTED>/<file>.md' }
(7+ similar "Queue full, dropping oldest event" lines in 2 seconds)
[10:16:10.546] [PGLite] Slow query (3379ms, exec=1ms, blocked=3378ms) table=session_files
[10:16:10.549] [PGLite] Slow query (3382ms, exec=1ms, blocked=3381ms) table=session_files
[...]

PR #423, opened 2026-05-24
Add optional PostgreSQL database backend

AnisminC:
## Summary

This adds an optional PostgreSQL database backend for Nimbalyst. The goal is to remove the single-writer bottleneck we hit with PGLite during heavier agent workloads, especially when Codex is running a parent session with several subagents writing transcript/session state at the same time.

The default path is unchanged: Nimbalyst still uses PGLite unless a Postgres connection is configured.

## What changed

- Added a Postgres-backed database adapter using `pg.Pool`.
- Moved Postgres schema setup into a shared schema module.
- Added a PGLite-to-Postgres migration path.
- Added an Alpha PostgreSQL settings page under Application settings.
- Added connection testing and migration controls in settings.
- Fixed transaction handling so pooled Postgres clients are always committed/rolled back and released.
- Kept env var support for local/dev usage while allowing saved settings to apply on the next app restart.

## Testing

- Ran the existing Electron typecheck.
- Built the Electron package locally.
- Migrated an existing PGLite database copy into a local Docker Postgres instance.
- Ran Nimbalyst locally against Postgres with Codex sessions active.
- Checked Postgres activity during the run: no idle-in-transaction connections and no lock waits observed.

## Notes

This is intentionally marked Alpha in settings. PGLite remains the default backend, and the migration flow snapshots the source PGLite directory before copying rows into Postgres.


AnisminC:
One useful side effect of this change: when Nimbalyst is configured to use PostgreSQL, it appears to resolve the Codex crash/hang seen during heavier Meta Agent sessions.

The case that reproduced it most often was a parent Codex session with multiple subagents running at the same time. With PGLite, the app could stall or crash under that write load. After migrating the same data to PostgreSQL and running the same style of Meta Agent workflow, the session stayed responsive and Postgres showed no idle-in-transaction buildup or lock waits during testing.

PGLite remains the default backend, but PostgreSQL gives heavier multi-agent workflows a much safer concurrency path.

ademczuk:
Hey @AnisminC, wanted to throw some empirical evidence + a code review behind this while it's waiting for Greg/Karl. I ran a synthetic write-load benchmark against the schema in `postgresSchema.ts` and verified two findings against the diff. Hopefully useful.

## Bench: write-throughput, PGLite vs Postgres

Same `ai_sessions` + `ai_agent_messages` subset of your schema, parameterized 3-statement transaction (`BEGIN -> INSERT message -> UPDATE session -> COMMIT`), N concurrent agent loops for 20s per scenario. Three backends: PGLite in-memory (best case), PGLite file-backed (mirrors Nimbalyst's real usage), Postgres with `pg.Pool max=20` (matches your `poolMax` default in `store.ts:1432`).

| Backend | N | ops/s | p50 ms | p95 ms | p99 ms | max ms |
|---|---|---|---|---|---|---|
| pglite-file | 1 | 110 | 8.6 | 11.9 | 15.9 | 37.7 |
| pglite-file | 4 | 110 | 35.3 | 44.5 | 54.3 | 63.0 |
| pglite-file | 16 | 110 | 142.2 | 170.0 | 196.5 | 208.1 |
| pglite-file | 32 | 111 | 285.7 | 322.8 | 373.7 | 383.0 |
| pglite-file | 64 | 110 | 575.1 | 645.4 | 726.9 | 730.5 |
| postgres | 1 | 190 | 4.3 | 8.9 | 18.7 | 183.0 |
| postgres | 4 | 627 | 5.5 | 11.9 | 21.8 | 66.7 |
| postgres | 16 | 1342 | 9.9 | 23.7 | 39.6 | 95.1 |
| postgres | 32 | 1160 | 22.8 | 59.6 | 91.3 | 133.6 |
| postgres | 64 | 1381 | 40.8 | 82.6 | 127.4 | 244.6 |

Headlines:
- **File-backed PGLite is fsync-capped at about 110 ops/s regardless of N.** Adding writers does not increase throughput - they queue. p99 latency grows 45x from N=1 (16 ms) to N=64 (727 ms).
- **Postgres scales near-linearly to N=16**: 1342 ops/s (12x file-backed PGLite throughput), at 5x lower p99 (40 ms vs 196 ms).
- **Postgres pool oversubscribes at N=32+** with `max=20`, but still holds 10x+ over PGLite at every concurrency level above N=4.

That quantitatively backs the architectural argument here. Might be worth a one-table paste into the PR body as a numeric anchor.

**What the bench does not show**: the specific hang/crash you observed. Latency growth here is graceful - no >2s hangs in 20s windows. The hang you saw is plausibly (a) a longer-duration queue-build-up effect (worth a rerun with `BENCH_DURATION_MS=120000`), (b) an OS-specific PGLite WAL/fsync issue (you're on macOS, I ran on Windows), or (c) something specific in Nimbalyst's PGLite worker that my synthetic 3-statement loop does not exercise. The bench package (Dockerfile + docker-compose + `bench.mjs` + README + results) is sitting on my box - happy to share via a gist or however you'd prefer.

## Two findings in the diff worth addressing

**1. Postgres connection string stored in plain-text electron-store** - `packages/electron/src/main/utils/store.ts:1423-1432`, `packages/electron/src/main/bootstrap.ts:135-141`

`PostgresBackendSettings.connectionString: string` is written verbatim to `app-settings.json` via `getAppStore().set('postgresBackend', next)`. On macOS that's `~/Library/Application Support/nimbalyst/app-settings.json` - user-readable and routinely included in app-support backups (iCloud, Time Machine). Most `postgres://user:pass@host/db` connection strings include the password inline. The codebase already has the precedent for OS-keychain storage: `SettingsHandlers.ts:247,271` uses `safeStorage.encryptString` / `safeStorage.decryptString`, and `CredentialService` (imported at `SettingsHandlers.ts:48`) wraps it. Suggested shape: keep `enabled` / `poolMax` / non-sensitive metadata in electron-store; encrypt the connection string with `safeStorage`, or split host/db/user plain + password via keychain.

**2. Migration leaves Postgres in partial state on failure, no rollback or row-count verification** - `packages/electron/src/main/database/pgliteToPostgresMigration.ts:195-225`

`migratePgliteToPostgres()` iterates `TABLES` and calls `copyTable()` per table. Each batch is INSERTed with `ON CONFLICT DO NOTHING`. If `copyTable()` throws partway (say on table 7 of 17), the first 6 tables are already committed in Postgres, the `finally` block releases the pool, and the function rejects. On rerun without `--truncate`, the `ON CONFLICT DO NOTHING` silently keeps the partial state - tables 1-6 stay, 7-17 stay empty, and the user has no signal that the migration was incomplete. There is also no end-of-migration source-vs-target row-count comparison: `tableCounts` is just a SUM of inserts. Smallest fix: assert `sourceRowCount[t] === targetRowCount[t]` for each `t` after the loop and surface any mismatch as `success: false`. Larger fix: wrap the table-copy loop in a single Postgres transaction so a partial failure rolls back atomically.

## Open questions that would help frame the maintainer review

a. **Could the PGLite path be fixed instead of adding a second backend?** A `BatchWriter` that coalesces concurrent agent writes into one transaction every ~5-10ms could plausibly absorb the Codex Meta-Agent burst without the maintenance cost of a second adapter. If you explored that and ruled it out, having that in the PR body short-circuits the inevitable maintainer question.

b. **Was the hang's root cause confirmed?** Was it the single-writer queue blocking under concurrent inserts, or something more specific (WAL lock, init lock, a particular index path)? My bench shows queueing-latency but no hard hangs in 20s. Your repro probably pins a different mechanism, and stating it explicitly in the PR strengthens the case.

c. **Was the plain-text credential storage intentional for this first cut?** Given the existing `CredentialService` precedent, it stands out. If there was a bootstrap-order reason (the connection string needs to be readable before `app.whenReady()`), `safeStorage.decryptString` is sync after app init and works at the bootstrap stage.

I also have a few medium-priority improvements and a couple of nits (for instance the `as unknown as PGLiteDatabaseWorker` cast at `PGLiteDatabaseWorker.ts:1074-1078` which makes the Postgres-vs-PGLite interface mismatch a runtime hazard rather than a compile error) - happy to drop those in a follow-up if useful, but I did not want to crowd a single comment. Same for the bench harness - say the word and I will share.

Also ran a Codex pass against the diff after the comment above and it surfaced two more I think are worth your eyes. Complementary to the first set, not duplicates.

**3. The connection string leaks to subprocess env via `process.env.NIMBALYST_DATABASE_URL`** - `packages/electron/src/main/bootstrap.ts:141`

Bootstrap copies the saved connection string into `process.env.NIMBALYST_DATABASE_URL` so the database layer can pick it up at init. But every `child_process.spawn` / `execFile` in the codebase that does not explicitly override the `env` option inherits the parent's full env by Node's default. The spawn paths I checked (`packages/electron/src/main/ipc/ClaudeCodeHandlers.ts`, `ClaudeCodePluginHandlers.ts`, `ExtensionMarketplaceHandlers.ts`, and the Codex/Claude provider spawn paths under `packages/runtime/src/ai/server/providers`) do not filter `env`. Net effect: any agent CLI subprocess, MCP server, plugin process, or auth-flow shell window inherits the Postgres connection string with the password inline. That's a wider credential-leak surface than the on-disk concern in point 1 above, because subprocesses can log env, can re-spawn their own children with the same env, and an attacker who lands code execution as one of those subprocesses gets the database URL directly. Same fix shape as point 1: keychain-encrypted at rest, then inject only for the database client itself (pass the decrypted string straight to `new Pool({ connectionString })` from in-memory), not via the global `process.env`.

**4. `verifyBackup()` on the Postgres backend hard-fails project move/rename** - `packages/electron/src/main/database/PostgresDatabase.ts:160-162`

`async verifyBackup()` returns `{ valid: false, error: 'Postgres backend does not use PGLite directory backups' }` unconditionally. `ProjectMigrationService.ts:157` and `:276` call `backupService.createBackup()` during move/rename, and `DatabaseBackupService.createBackup()` at `packages/electron/src/main/services/database/DatabaseBackupService.ts:254` verifies the result via `verifyBackup()` before declaring success. With Postgres active, that verification path always returns false, so project moves and renames will reject for Postgres-backed users with an error that reads like a misconfiguration even though it's the intended "not applicable" case. Two fix shapes: gate `ProjectMigrationService` on `database.getBackupService() !== null` and skip the verification leg when there's no backup service (cleaner short-term), or make `PostgresDatabase.verifyBackup()` a `{ valid: true }` no-op when there's nothing PGLite-shaped to verify (semantically dishonest but smallest patch).

That's everything from this pass. Bench harness offer from the first comment still stands.

Genuinely supportive of this PR landing in some form. The throughput delta is real.

ghinkle:
Thanks for both the proposal and the review and benchmarking. Very helpful. 

I've been worried about the single-threaded nature of pglite for some time, but haven't been able to decide on the correct fix. I've been adding coalescing in the most obvious places, but it does seem like a losing battle. 

My original plan (8 months ago) with pglite had been to take advantage of wasm and pglite sync features for mobile integration, but I ended up going with sqlite in the mobile apps for performance and reliability anyway. At that point I started to lean towards an eventual migration to better-sqlite. 

Since then, I've come to rely on the JSONB support for flexible schemas in areas like the user definable tracker system data models. The GIN indexes are also quite useful. 

Had claude coalesce this thread and a set of sessions I've been where I've been considering options:

https://share.nimbalyst.com/share/Hrt6x88IFKSyJFGtamknQU#key=ZmEhQrqyxARnVlpJRdbtCuruNwavXcHkEGGrlOV76K0

Feedback welcome!

AnisminC:
Thanks @ghinkle, that's helpful context the JSONB/GIN dependency is exactly what makes this trickier than a straight backend swap.

That's honestly the main reason I went with Postgres instead of better-sqlite. SQLite can index JSON expressions, but there's nothing really equivalent to a GIN index for arbitrary containment queries against the tracker data models. So moving to SQLite would mean either reshaping those flexible schemas or eating slower lookups, whereas Postgres keeps the JSONB + GIN story you're relying on while also clearing the single-writer bottleneck.

One thing worth flagging on the SQLite angle though: better-sqlite3 is synchronous and serializes through a single writer, so the heavy Meta Agent burst (parent Codex session plus a bunch of subagents all writing transcript/session state) would hit basically the same write-serialization wall pglite does. WAL gets you concurrent readers, but writes still queue through one connection which is exactly what was stalling Codex for me. So for the heavy multi-agent case at least, I'm not sure SQLite actually buys much over where we are now.

Going to read through the coalesced doc properly before saying more.

Also, @ademczuk's two credential findings the plain-text connection string on disk and the leak into subprocess env both look legit and will need fixing before this is safe to land, whichever direction it goes. I can patch the PR if that's the route @ghinkle wants to go with this.

ademczuk:
@ghinkle helpful doc, that lays out the trade pretty cleanly. The single question that seems to actually decide the choice:

**In the first hosted/collab release, will the server reuse desktop queries verbatim, or share row-level locking, LISTEN/NOTIFY, SKIP LOCKED, or JSONB containment semantics?**

- Yes -> embedded-Postgres pays for itself in dialect parity.
- No -> better-sqlite3 + JSON1 + FTS5 + WAL, with a stable join-seam table (e.g. `tracker_item_labels`) designed in now as the containment escape hatch. Same pattern Linear's LSE uses.

Three primary-source data points that may make the call cleaner:

1. The only Electron-class app shipping `leinelissen/embedded-postgres` in production (`paperclipai/paperclip`) has documented field failures matching every risk in your RFC: macOS arm64 dyld rejection of bundled `libssl.3.dylib` (paperclipai/paperclip#3583), asar incompatibility forcing `extraResources` (paperclipai/paperclip#5410), silent initdb crashes (paperclipai/paperclip#1032).
2. ElectricSQL themselves pivoted away from Tauri-embedded-Postgres to PGlite-WASM, citing the embedded approach as "limited to desktop use only."
3. FTS5 BM25 default ranking is actually a quality upgrade over Postgres `ts_rank` for English-only desktop search. Per TigerData's own pg_textsearch launch post: `ts_rank` lacks IDF and document-length normalization, which FTS5 has built-in.

One audit data point on `tracker_items.data` in main today: production code queries it via 3 generated columns (`title`, `status`, `kanbanSortOrder`) plus one explicit `data->'labelsMap'` read. Zero `@>` containment queries in production. The dynamic schema is on the write side, not the query side. SQLite expression indexes on the same known paths give native-column speed (canonical 1M-row delphitools.info bench: 976ms full scan -> 0.15ms with expression index).

Happy to extend the PR #423 harness to bench the head-of-line case you actually care about - one slow query interleaved with many fast ones, measured across PGLite, better-sqlite3+WAL, and embedded-Postgres - if a number would help settle either branch.

@AnisminC fair on both:

**(1) JSONB / GIN.** The audit of how `tracker_items.data` is actually queried in main today was the part that surprised me. 3 generated columns (`title`, `status`, `kanbanSortOrder`), one explicit `data->'labelsMap'` read, zero `@>` containment queries in production source. So the "reshape schemas or eat slower lookups" fork doesn't bite today. SQLite expression indexes on the same known paths give native-column speed (the canonical delphitools.info bench at 1M rows: 976ms full scan -> 0.15ms with an expression index on `json_extract(data, '$.vx')`). For future containment, the escape hatch is a `tracker_item_labels(item_id, label)` side table with `(item_id, label)` and `(label, item_id)` indexes, designed during migration. That's the same pattern Linear's LSE talk references.

**(2) SQLite write serialization.** Real concern, partly addressed by the architecture though. PGLite serializes EVERY query (reads + writes) through a single WASM executor. better-sqlite3 + WAL serializes only writes, readers run in parallel. So the asymmetric workload Greg's RFC names (one slow query stalling unrelated reads) gets cleanly fixed. Pure write contention is still real, but better-sqlite3's writer is native-synchronous (typically <1ms per small transaction) vs PGLite's WASM-interpreted serial executor for everything. For your Codex Meta-Agent case specifically the bench worth running is: parent + N subagents all writing transcript/session state, measure write throughput and read latency under PGLite, better-sqlite3+WAL, and embedded-Postgres. I can extend my PR #423 harness for that if useful, separately from Greg's decision.

**(3) Credentials.** Appreciate you offering to patch. While you're in there, three more from a fresh Codex review on the diff worth folding in:

- `pgliteToPostgresMigration.ts:174-210` and `:318-341` - `copyTable` uses `INSERT ... ON CONFLICT DO NOTHING` and increments `copied` by source row count, not actual inserts. No target verification, no per-table transaction wrap. A mid-run conflict can leave a partial target and still log "Migration complete" - silent data loss is the worst failure mode for a migration tool.
- `pgliteToPostgresMigration.ts:125-130` - `normalizeColumnValue` calls `JSON.stringify(value)` for any column in `JSON_COLUMNS` regardless of whether PGLite already returned it as a string. If any of the JSON column reads come back as serialized text rather than parsed objects (driver-dependent, but worth confirming on the actual PGLite path), the result double-encodes. Worth a quick `typeof value === 'string'` guard or asserting the assumed driver shape.
- `PostgresDatabase.ts:117-123` - pool shutdown race: `await this.pool.end()` runs BEFORE clearing `initialized` and `pool`. During the drain window other callers still see the backend as initialized and start queries against an ending pool, which `pg` rejects.

All three are small fixes, flagging since they're in the lane you'd be touching for the credential cleanup.

ghinkle:
Thanks for the perspective, both of you. While GIN is nice, it is not the only solution to what I'm planning. 

The collab system will not be using any postgres or shared queries. That was a bad input to the original design process months ago unfortunately. It now has a separate CRDT backed by sqlite in cloudflare durable objects. 

I want this decision to focus on making the desktop app solid and performant. While I've used postgres many times in servers, as you said, embedding it would be unusual. I'm not sure it is worth the risk.

I'm not against taking this PR in the short term, it doesn't solve the problem for the majority of our users just asking for an AI runtime platform. I think better-sqlite with continued write coalescing will probably get us what we want as the permanent solution. I'll spin up a worktree just to check on difficulty / complexity. Thanks!

ademczuk:
I spent some time to build a prototype of how we could address the write-coordinator pattern with bench numbers, in case it's useful for your worktree spike. It documents the trade-offs honestly (2x throughput vs slightly higher p99) and leaves open design questions for review.

@ghinkle (the FTS-stalling-reads case named in the RFC):
  - WAL handles read concurrency natively. Coord's bg-lane yielding ensures the slow op also yields the JS event loop, so reads aren't blocked by main-thread monopolisation either. Bench shows 0.06ms p99 reads during a 644ms FTS rewrite.
  - "Continued write coalescing" plan is literally what the coord does, but as an architectural invariant instead of a per-call discipline.

I prototyped a `WriteCoordinator` pattern against your worktree-spike scope and benched it. Posted on my fork as a reference rather than upstream PR since you're driving the migration decision: ademczuk/nimbalyst-1#1, which also includes explicit macOS instructions if you want to run the dockerfile bench.

Headline empirical results across better-sqlite3+WAL (direct dev-chunked), the same engine + coordinator, and postgres (semantic equivalent of embedded-Postgres):

| Backend | Baseline ops/s | Baseline p99 | 50-write burst |
|---|---:|---:|---:|
| bs3+WAL direct | 8328 | 0.17ms | 3ms |
| bs3+WAL+Coordinator | 17946 | 1.35ms | 2ms |
| postgres | 920 | 6.24ms | 32ms |

Under HOL pressure (slow FTS rewrite + concurrent hot writers), the coord runs 10980 hot ops during the 2311ms slow op at p99 14ms; the dev-direct version runs 500 during a 644ms slow op at p99 1.3ms. Both bs3 backends preserve hot-lane interleaving when slow ops are chunked; the coord just centralises the discipline. Postgres loses on both throughput and latency for this single-user workload.

What the bench validates: the coordinator pattern delivers about 2x baseline throughput via transaction/fsync amortisation, and the bg-lane yielding keeps the JS event loop free during slow writes so reads stay at 0.06ms p99. What it does NOT validate: cross-process Codex Meta-Agent burst (the bench is in-process), sustained load past 5s windows, or FTS5 vs to_tsvector behavioural parity. Those are explicitly out of scope and called out in the proposal docs.

Useful for your worktree spike or ignore at will - just keeping it referenceable so the empirical work isn't lost if you decide to fold any of it into your evaluation.

@AnisminC (the multi-agent write-burst case):
  - In-process burst (50 concurrent writes) lands in 2ms via batching vs 3ms direct. The fsync amortisation is real.
  - CAVEAT explicitly documented: cross-process IPC case (real Codex Meta-Agent) needs a separate bench. Not claiming the in-process result generalises.
  - the cross-process case for your Codex Meta-Agent scenario still needs its own bench; happy to build that next if it would help.

ademczuk:
Following up on the thread above now that #476 (the better-sqlite3 migration) is open.

That PR is the worktree spike you mentioned, Greg, and it looks like the direction for the desktop app: opt-in better-sqlite3 with a versioned MigrationRunner, the _migrations ledger, FTS5, a non-destructive dry-run, and the rolling-backup cutover. For the majority of users who just want a solid local AI runtime, that path covers the #365 bottleneck this PR set out to fix, and it sidesteps the embedded-Postgres packaging risks we walked through (macOS arm64 dyld, asar, initdb).

Given that, it makes sense to step back from a second backend here rather than carry two adapters. One thing worth carrying forward though: the cross-process write-burst case. better-sqlite3 with WAL fixes the slow-query-stalling-reads problem cleanly, but writes still queue through one writer, so the parent-plus-subagents Codex burst can still serialize. The WriteCoordinator prototype I posted on 2026-05-27 (ademczuk/nimbalyst-1#1) centralizes the coalescing as an architectural invariant and has bench numbers for the head-of-line case, about 2x baseline throughput from transaction and fsync amortization, with the background lane yielding the event loop so reads stay near 0.06ms p99. The two credential findings from the review (plain-text connection string on disk, and the leak into subprocess env) generalize too, since any at-rest secret needs the same safeStorage treatment.

Closing this in favor of #476. Happy to fold the write-coordinator and the credential hardening into that branch rather than duplicate the work here, and if the cross-process burst case would help the evaluation I can build that bench against #476 next. Thanks for driving this.

PR #438, opened 2026-05-25
fix(agent): show Blitz/Super Loops/Meta Agent in the new-session menu without developer mode

### LINKED ISSUE #437: Blitz and Super Loops require developer mode to appear in the new-session menu
## Summary

With the Blitz, Super Loops, and Meta Agent alpha features enabled in Settings > Agent Features, the new-session "+" menu does not surface them unless developer mode is also on.

- In standard mode, clicking "+" creates a new session directly with no dropdown.
- In developer mode, the dropdown appears but shows only Meta Agent, not Blitz or Super Loops.

The alpha-feature registry says these features are "available to all users regardless of release channel", so the alpha toggle should be enough and developer mode should not be required.

## Steps to reproduce

1. Settings > Agent Features: enable Blitz, Super Loops, and Meta Agent.
2. Keep developer mode off (standard mode).
3. Click the "+" in the session list.

Expected: a dropdown with New Session plus the enabled alpha options.
Actual: clicking "+" creates a new session directly, no dropdown. Switching developer mode on shows the dropdown but only with Meta Agent.

## Root cause

Two issues:

1. Blitz and Super Loops are gated by `isWorktreesAvailable`, which is `developerFeatureAvailableAtom('worktrees')` = `developerMode && worktrees`. Meta Agent is gated by its alpha flag alone, which is why it is the only one that survives.
   - `AgentMode.tsx`: `const isBlitzAvailable = isWorktreesAvailable && isBlitzAlphaEnabled;`
   - `SessionHistory.tsx`: `const isSuperLoopsAvailable = isWorktreesAvailable && isSuperLoopsAlphaEnabled;`
2. `handleNewButtonClick` in `SessionHistory.tsx` decides whether to open the dropdown from `[onNewSession, onNewWorktreeSession, onNewBlitz, onNewTerminal]` only. It never counts the atom-gated alpha items, so in standard mode the single available option (New Session) fires directly and the menu never opens.

## Affected versions

Present on main (0.61.1) and earlier (confirmed back to 0.58.18).

ademczuk:
## What

Decouples the Blitz and Super Loops new-session options from developer mode so they follow their alpha-feature toggle alone, matching Meta Agent and the alpha registry's "available to all users" contract. Also makes the "+" button open the dropdown when an alpha option is available.

Fixes #437.

## Why

See #437. Blitz and Super Loops were gated by `isWorktreesAvailable` (which requires developer mode), so enabling their alpha features did nothing in standard mode, and in developer mode only Meta Agent appeared. The "+" also skipped the dropdown entirely in standard mode because it never counted the alpha items.

## Changes

- `AgentMode.tsx`: `isBlitzAvailable = isBlitzAlphaEnabled` (was `isWorktreesAvailable && isBlitzAlphaEnabled`); the `createNewBlitz` guard now checks `isBlitzAlphaEnabled`. Still requires a git repo for the worktree it creates.
- `SessionHistory.tsx`: `isSuperLoopsAvailable = isSuperLoopsAlphaEnabled` (was `isWorktreesAvailable && ...`); removed the now-unused `isWorktreesAvailable` and its import; `handleNewButtonClick` now counts `isSuperLoopsAvailable` and `isMetaAgentEnabled` so the dropdown opens in standard mode.
- New Worktree stays developer-mode-gated, unchanged.

## Testing

- Standard mode, alpha features on: "+" opens the dropdown with New Session, New Blitz, New Super Loop, New Meta Agent. Blitz and Super Loop are disabled with a tooltip outside a git repo.
- Developer mode: all options including New Worktree appear.
- Verified the three alpha options render in the dropdown on a clean build.

### DIFF
--- a/packages/electron/src/renderer/components/AgentMode/AgentMode.tsx
+++ b/packages/electron/src/renderer/components/AgentMode/AgentMode.tsx
@@ -120,9 +120,11 @@ export const AgentMode = forwardRef<AgentModeRef, AgentModeProps>(function Agent
 
   // Check if worktrees feature is available (developer mode + feature enabled)
   const isWorktreesAvailable = useAtomValue(worktreesFeatureAvailableAtom);
-  // Blitz requires both worktrees and the blitz alpha feature
+  // Blitz is gated by its alpha feature alone (like Meta Agent), not developer
+  // mode. The worktrees it creates still require a git repo, enforced in
+  // createNewBlitz and on the New Blitz button (disabled when !isGitRepo).
   const isBlitzAlphaEnabled = useAtomValue(alphaFeatureEnabledAtom('blitz'));
-  const isBlitzAvailable = isWorktreesAvailable && isBlitzAlphaEnabled;
+  const isBlitzAvailable = isBlitzAlphaEnabled;
 
   // Keep Super Loop listeners active even when session history is collapsed/hidden.
   useSuperLoopInit(workspacePath);
@@ -545,9 +547,9 @@ export const AgentMode = forwardRef<AgentModeRef, AgentModeProps>(function Agent
 
   // Create new blitz - opens the blitz dialog
   const createNewBlitz = useCallback(() => {
-    if (!isWorktreesAvailable || !isGitRepo) return;
+    if (!isBlitzAlphaEnabled || !isGitRepo) return;
     setShowBlitzDialog(true);
-  }, [isWorktreesAvailable, isGitRepo]);
+  }, [isBlitzAlphaEnabled, isGitRepo]);
 
   // Handle blitz creation result
   const handleBlitzCreated = useCallback(async (result: any) => {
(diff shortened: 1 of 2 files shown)


ghinkle:
The reason I hadn't opened these features up, was that they felt incomplete. Have you used them @ademczuk
Do you feel like they're working enough to be useful to people? I haven't had time to come back to them recently for polish.

ademczuk:
Hi @ghinkle I quite like the concept as I used blitz today running a pre-mortem against my own gemini-flash35 PR. Found three bugs with Blitz in one session, so there goes Karls hope for <80 issues. Filed #452 #453 #454, PR #455 up for #452. 

My read on whether it's shippable: the architecture is right, the polish gap is real. 
Silent failure paths in the creation flow, no error toast on 0/N worktrees, renderer state divergence on session-delete. 

Once those land it's a properly useful feature. Happy to PR #453 (renderer-side, ~5 lines once you tell me how to enumerate the workspace atom family) and #454's workaround if you'd prefer same-codebase over upstream Codex.

ghinkle:
@ademczuk decided to merge back and we can clean up and improve as we go

PR #477, opened 2026-05-29
fix(theme): resolve Monaco editor dark/light from isDark metadata

omartelo:
## Problem

`getEffectiveBaseTheme()` (`packages/electron/src/renderer/hooks/useTheme.ts`) decided an extension theme's dark/light by string-matching `'dark'` in the theme ID. That resolved value is what feeds the Monaco editor.

The internal Nimbalyst UI already resolves dark/light from the theme's real `isDark` flag (`applyThemeToDOM` → `getRuntimeTheme(id).isDark`). So an extension theme **without** a `-dark` suffix (e.g. `rose-pine`) rendered a dark UI but a **light Monaco editor** — the two paths disagreed.

Workaround until now: theme authors had to append `-dark` to the theme ID so the substring check passed.

## Fix

Resolve the base theme from `getRuntimeTheme(id).isDark` — the same registry source the UI uses — and fall back to the ID-name heuristic only for themes not yet in the registry (file-based built-ins like `crystal-dark`/`solarized-*`, whose names already match).

`isDark` becomes the single source of truth for the editor too.

```
useTheme → getEffectiveBaseTheme('com.rosepinetheme.nimbalyst:rose-pine')
  → registry.isDark=true → 'dark'
TabEditor config.theme='dark' → getMonacoTheme('dark', …) → 'vs-dark'
```

Also corrects Lexical, Mermaid, and code-highlight, which consume the same resolved theme.

## Test plan

- [x] Install an extension theme whose ID has no `-dark` suffix and is `isDark: true` (e.g. Rosé Pine). Open a code file → Monaco renders dark.
- [x] Light extension theme (`isDark: false`, no `-light` in ID) → Monaco renders light.
- [x] Built-in `light`/`dark` unchanged.
- [x] File-based `crystal-dark` / `solarized-light` unchanged (registry miss → name fallback).
- [x] `system`/`auto` still follow OS preference.

### DIFF
--- a/packages/electron/src/renderer/hooks/useTheme.ts
+++ b/packages/electron/src/renderer/hooks/useTheme.ts
@@ -347,7 +347,9 @@ function clearCustomThemeVariables(): void {
 /**
  * Get the effective base theme for a theme ID.
  * Handles 'system' and 'auto' by checking OS preference.
- * For file-based themes, we determine dark/light from the theme metadata.
+ * For extension/file-based themes, we determine dark/light from the theme's
+ * `isDark` metadata (the same source the UI uses in applyThemeToDOM), falling
+ * back to the theme ID name only when the theme isn't in the registry yet.
  */
 function getEffectiveBaseTheme(themeId: string): ConfigTheme {
   // Handle 'system' and 'auto' by checking OS preference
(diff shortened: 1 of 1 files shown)


ademczuk (commented):
Reviewed and this looks correct to me. Traced it against the branch:

- `getRuntimeTheme` is properly imported (aliased from `getTheme` in `@nimbalyst/runtime`), so the new registry lookup resolves.
- `getEffectiveBaseTheme` now resolves dark/light from the theme's actual `isDark` metadata, falling back to the ID-name heuristic only when the theme is not in the registry yet. That is the same source `applyThemeToDOM` already uses, and it extends the exact pattern already present a few lines up where `resolvedTheme` is resolved through `getRuntimeTheme`. Consistent, not a new approach.
- It fixes the real case: extension themes whose ID has no `-dark` suffix (rose-pine and similar) now resolve dark in Monaco instead of defaulting light.

The registry lookup is synchronous and the ID-name fallback covers the pre-registration window, so there is no early-load regression. Small, well-scoped fix. LGTM from my read.

Issue #493, opened 2026-05-31
Edit any previous message and continue from there


TaiwanTammy:
## Problem
Once a message is sent, it's locked in. If I realize my prompt was wrong, incomplete, etc, my options are to start a new session (losing context) or keep going and let the bad turn pollute everything downstream.

Same problem at the other end of a session: long, valuable threads accumulate off-topic detours, bloated low-value exchanges, and wrong answers I now have to talk around. They sit in context, cost tokens, and skew future turns.

## Proposed solution
**v1:** Let me edit any previous user message and re-run from that point. Everything after the edited message is discarded. Huge upgrade with this alone.

**v2:** Preserve the original branch instead of discarding it, with a toggle to switch between branches — what ChatGPT and Claude.ai already do.  

**Also:** Let users delete any individual message (user or LLM's) so it's removed from the context sent on the next turn, not just hidden. This is the surgical version of cleaning up a polluted session without forking or restarting.

## Alternatives considered
- Nimbalyst's existing branch-from-current-state. Doesn't help when the point you want to rewind to is earlier in the thread.

omartelo:
### Minimal proposal (v0)

Picked the smallest slice that unblocks the most common case — you sent the wrong or incomplete prompt and want a do-over before the model wastes a turn on it.

- A small pencil button shows next to the copy button on the last user message, **only when the model has not replied yet** (cancelled mid-stream, or freshly sent before the assistant starts).
- Click it and the bubble turns into an editable textarea in place.
- Submit and the prompt re-runs with the new content; the bubble updates in place with no flicker.

What this **does not** cover yet (kept on this issue for follow-up):

- v1 — editing an earlier message and discarding the conversation downstream.
- v2 — preserving branches with a toggle (the ChatGPT / Claude.ai behaviour).
- Per-message delete.

PR for the v0 slice: #503.

TaiwanTammy:
dying for this feature to exist

doughertyd2:
I opened #1160 before finding this — it's the same underlying "v1" (rewind to an earlier message and discard downstream), so linking it here to consolidate.

Two things that might be useful:

- **Different entry point.** Rather than a pencil on the message bubble, #1160 proposes driving it from the existing Prompts Menu dropdown: select a prompt → confirmation modal → truncate the conversation to that point and drop the message's text back into the input box. The markers there already carry each prompt's id and text, so the "which point / what text" data is in hand.
[...]

PR #503, opened 2026-06-01
feat: edit and re-run the last user message before the model replies

omartelo:
## Summary

v0 slice of #493. A pencil button next to the copy action on the last user message opens an inline editor when the model has not yet replied (cancelled mid-stream, or fresh prompt before the first assistant turn). Submitting updates the message in place and re-runs the turn against the provider with no bubble flicker and no duplicate raw row.

Builds directly on the user-proposed v1 in #493.

## Approach

UPDATE the existing raw user row and its derived `user_message` canonical event in place — same `ai_agent_messages.id`, same `ai_transcript_events.id`, same `sequence`. The renderer keys bubbles by `message.id`, so a same-id payload swap is a React reconcile, not an unmount/remount — no flicker, no layout shift cascade.

Provider then receives the new content via the normal `sendMessageHandler` path, but the AI service sets a one-shot `setSkipNextInputLog(true)` directive on the cached provider so its automatic user-direction log no-ops once; the `MessageStreamingHandler` also reads an `editedInPlace` flag on the document context and skips its own `sessionManager.addMessage` + provisional-title write so the conversation history forwarded to the model API is not duplicated. The renderer mutates the bubble text optimistically before firing the IPC so the layout reflow lands the same frame the editor closes.

The append-only invariant on `ai_agent_messages` is intentionally bent here, scoped to a row that has not produced an assistant turn yet (i.e. no derived canonical state needs invalidating).

## Approaches considered

Earlier iterations did not land:

- **Delete + re-insert** the raw row and canonical event, then call `sendMessageHandler` normally. Produced a new canonical event id, so the bubble visibly remounted: a "two bubbles for ~2s, then one" flicker on the renderer until a session reload pruned the old id.
- **Client-side optimistic removal** of the old bubble. Replaced the two-bubble flicker with a "Thinking… with nothing above it" gap because the new canonical event had not yet streamed in.
- **Server-side `transcript:session-reparsed` broadcast** after the delete. Shortened the gap but did not eliminate it — the IPC + DB + push round-trip is observable.

The in-place UPDATE path is the only one that survives the React key constraint, because it never produces a new id for the bubble to remount around.

## Changes

| Path | Change |
| --- | --- |
| `packages/runtime/src/storage/repositories/AgentMessagesRepository.ts` | `AgentMessagesStore.updateMessageContent` + facade method |
| `packages/electron/src/main/services/PGLiteAgentMessagesStore.ts` | PGLite `UPDATE ai_agent_messages SET content = $1 WHERE id = $2` |
| `packages/runtime/src/ai/server/AIProvider.ts` | `setSkipNextInputLog` one-shot directive on `AIProvider` + consumed inside `BaseAIProvider.logAgentMessage` / `logAgentMessageNonBlocking` for `direction === 'input'` |
| `packages/electron/src/main/services/ai/AIService.ts` | New `ai:editLastUserMessage` IPC: UPDATEs the raw row, updates the canonical event (`updateEventPayload` + `updateEventText`), broadcasts the updated event over `transcript:event`, sets `setSkipNextInputLog(true)` on the cached provider, then calls `sendMessageHandler` with `editedInPlace=true` |
| `packages/electron/src/main/services/ai/MessageStreamingHandler.ts` | Honours `documentContext.editedInPlace` to skip the duplicate `addMessage` + provisional title write |
| `packages/electron/src/preload/index.ts` | Exposes `aiEditLastUserMessage` |
| `packages/runtime/src/ui/AgentTranscript/components/RichTranscriptView.tsx` | Pencil button next to the copy action, inline edit textarea, fire-and-forget submit; `editLastUserMessageDisabledReason` renders the pencil disabled with a tooltip |
| `packages/runtime/src/ui/AgentTranscript/components/AgentTranscriptPanel.tsx` | `onEditLastUserMessage` + `editLastUserMessageDisabledReason` prop pass-through (and memo check) |
| `packages/electron/src/renderer/components/UnifiedAI/SessionTranscript.tsx` | Optimistic mutation of the last `user_message.text` in `sessionMessagesAtom` before firing the IPC; computes the sync gate from `sync:get-status` |
| `packages/runtime/src/ui/AgentTranscript/components/__tests__/RichTranscriptView.editAffordance.test.tsx` | Render tests for the enabled / disabled-when-synced / absent pencil states |
| `packages/runtime/package.json` | `./storage/repositories/TranscriptEventRepository` subpath export (avoids the `__ELECTRON_LOG__` double-register that root-importing `@nimbalyst/runtime` would trigger in the main process) |
| `CHANGELOG.md` | `[Unreleased] Added` entry |

## Multi-device sync gate

The in-place UPDATE rewrites the tail row of `ai_agent_messages`, which mobile and other devices sync directly. The sync protocol only models appended rows — `SessionChange` has `message_added` but no `message_updated` — so an edit never reaches a device that already pulled the original prompt, and that device keeps rendering the stale text. Raised by @ademczuk in review.

For v0 the edit affordance is gated off whenever the session's project is enabled for multi-device sync:

- Signal: `appConfigured` (signed in) AND `projectEnabled` (project selected for sync in Settings), read from `sync:get-status`. Re-queried on the `sync:status-changed` broadcast so toggling sync for the project flips the pencil at runtime without a reload.
- When gated, the pencil renders disabled with the tooltip "Editing is unavailable while this project syncs across devices" instead of opening the editor.
- Single-device sessions keep the full edit-and-rerun affordance.

The full `message_updated` `SessionChange` variant — propagating the edit through CollabV3 and iOS so synced devices stay consistent — is deferred as a follow-up.

## How verified

- `npm --prefix packages/runtime run typecheck` and `npm --prefix packages/electron run typecheck` both clean.
- `npx vitest run packages/runtime/src/ai/server/transcript/__tests__/` — 256/256 passing.
- `npx vitest run packages/runtime/src/ui/AgentTranscript` — 202/202 passing, including the new edit-affordance render tests (enabled tail / disabled-when-synced / absent-without-handler).
- Manual: fresh session, send a wrong prompt, click the pencil before the assistant turn lands, edit, submit. Bubble updates in place without flicker; assistant streams immediately below.
- Manual: same flow but cancel mid-stream first, then edit. Same expectation.

## Out of scope

The larger pieces of #493 stay on the issue for follow-up:

- v1 — edit any earlier user message and discard the downstream conversation.
- v2 — preserve branches and offer a toggle (ChatGPT / Claude.ai style).
- Per-message delete.
- `message_updated` sync propagation so edits reach synced devices (see Multi-device sync gate).

Closes the v0 portion of #493.


ademczuk:
Nice approach. The same-id swap to avoid an unmount/remount, and the flicker, is clean.

One thing worth confirming before merge: this does an in-place UPDATE on `ai_agent_messages`, which is the append-only raw table mobile clients sync directly (`projectRawMessages.ts`, `sync/types.ts:374`; mobile gets the raw rows, not the derived `ai_transcript_events`). Does the sync path propagate an in-place row UPDATE, or does it assume append-only and only push new rows? If it only pushes new rows, an edited message would update on desktop but stay stale on a mobile client that already synced the original.

Not a blocker if sync already handles row updates. Just flagging it since the PR notes the append-only invariant is bent here.

omartelo:
Verified, sync does not propagate the in-place row UPDATE.

`SessionChange` (`packages/runtime/src/sync/types.ts:368`) only models `message_added`, `metadata_updated`, `session_deleted`. There is no `message_updated` variant. The corresponding wire message in `CollabV3Sync.pushChange` is `appendMessage`, with no update counterpart. The edit handler also intentionally suppresses the `message:logged` signal (`setSkipNextInputLog(true)`) that `SyncedSessionStore` would normally hook into, so even if a `message_updated` event existed today nothing would fire it from this code path.

Net effect: any device that synced the original row keeps rendering the original via `projectRawMessages.ts`.

Plan: restrict the edit affordance to sessions that are not currently synced (`syncProvider.isConnected(sessionId)` check on the renderer side, with a tooltip explaining it). The bubble stays editable for single-device usage, which covers the v0 ask in #493, and we avoid silently diverging mobile / multi-device state.

Two alternatives I considered and rejected:

- A full `message_updated` event end-to-end would touch the shared `SessionChange` type, the Cloudflare CollabV3 worker, the iOS client, and backwards-compatibility on older mobile builds. It is the right long-term fix but well outside the v0 slice, better as a dedicated follow-up.
- Going back to a delete + re-insert flow would emit `message_added` for the new row, but there is no matching `message_deleted` in the sync protocol either, so the old row would still linger on remote devices. It would also bring back the React unmount/remount flicker I ran into during earlier iterations and worked around by moving to the in-place UPDATE this PR ships.

ademczuk:
Thanks for digging into that. The `SessionChange` enumeration confirms it: an in-place row UPDATE on `ai_agent_messages` won't reach a mobile client mid-session, so they'd keep showing the pre-edit prompt and then the new response, which reads as a desync.

Two ways I can see to handle it: gate the edit affordance to sessions that aren't actively syncing (simplest for a first version), or add a `message_updated` `SessionChange` variant and wire it through the push path so the edit propagates. I'd lean to the gate for this PR and leave the sync variant as a follow-up, but whether single-device-only edit is acceptable for v1 is really a maintainer call. Happy either way.

omartelo:
Pushed the gate in af3ac15.

The edit pencil is now disabled (with a tooltip) whenever the session's project is enabled for multi-device sync, so single-device sessions keep the edit-and-rerun affordance and synced sessions can't diverge.

- Signal: `appConfigured` (signed in) AND `projectEnabled` (project selected for sync in Settings), read from `sync:get-status`. It re-queries on the `sync:status-changed` broadcast, so toggling sync for the project flips the pencil at runtime without a reload.
- Tooltip on the disabled pencil reads "Editing is unavailable while this project syncs across devices".
- Added render tests for the three states (enabled tail / disabled-when-synced / absent-without-handler).

I left the full `message_updated` `SessionChange` variant out, as you suggested, and noted it as a follow-up in the PR body. Happy to open a tracking issue for it if you'd like it captured separately.

omartelo:
Closing this PR.

While bringing the branch up to date with `main`, it became clear the transcript subsystem this feature is built on has been substantially refactored since the branch was cut. Canonical transcript events moved from a persisted store to an in-memory per-session cache in `TranscriptRuntime`, and the repository this PR relied on was removed.

This PR's whole mechanism depended on the persisted store: it did an in-place update of the canonical event to keep the same React key and avoid the bubble flicker. That path no longer exists on `main`, so a merge turns into a rewrite of the core mechanism against the new architecture, on top of the unrelated drift across 80+ commits.

Rather than force a large reconciliation through a stale branch, the cleaner path is to re-implement the edit-and-rerun feature from scratch against current `main`. The design decisions still stand and will carry over:

- v0 scope: edit only the last user message, only before the model has replied (#493).
- The sync gate (Option B) discussed above: disable the affordance while the session's project syncs across devices, since edits do not yet propagate over CollabV3.
- A `message_updated` `SessionChange` variant remains the eventual follow-up for full multi-device support.

I will open a fresh PR against current `main`. #493 continues to track the feature. Thanks for the review here, the sync-divergence catch in particular carries straight into the new implementation.

Issue #146, opened 2026-05-03
[Bug]: Security: webSecurity: false on main BrowserWindow + unauthenticated MCP HTTP   server


bethintakoma:
### Summary

The main application window sets `webSecurity: false` in its webPreferences (`packages/electron/src/main/window/WindowManager.ts`, line 186). This disables same-origin policy enforcement in the renderer where users interact with AI sessions, view content, and manage API keys.

Combined with the MCP HTTP server on localhost (`packages/electron/src/main/mcp/httpServer.ts`) which sets `Access-Control-Allow-Origin: *` and requires no authentication, this creates a realistic attack path: any website open in the user's browser can make cross-origin requests to the MCP server and invoke tools — including filesystem operations, terminal writes, and extension tool execution.

### Steps to reproduce

1. Launch Nimbalyst with at least one workspace open
2. Note the MCP HTTP server port (starts at 3456)
3. From any webpage, `fetch('http://127.0.0.1:3456/...')` succeeds due to `Access-Control-Allow-Origin: *`

### Suggested fixes

1. **Remove `webSecurity: false`** from the main window's webPreferences. If it was added to load local `file://` resources, a custom protocol handler (`app://`) registered with `protocol.registerFileProtocol` would be the standard Electron approach.
2. **Add authentication to the MCP HTTP server** — a bearer token or shared secret generated at startup, so only authorized local processes (the app's own renderer, configured AI agents) can connect.
3. **Scope CORS** on the MCP server to specific origins rather than `*`.

### Context

All other BrowserWindows in the codebase (WorkspaceManagerWindow, AboutWindow, DatabaseBrowserWindow, DeveloperDashboardWindow, SplashScreen) correctly leave `webSecurity` at its default (`true`). The main window is the only outlier, which suggests this may be a development convenience that wasn't cleaned up.

The rest of the Electron security posture is solid — `nodeIntegration: false`, `contextIsolation: true`, proper preload bridge via `contextBridge.exposeInMainWorld`, and thorough path traversal protection in `SafePathValidator`. Fixing these two issues would bring the main window in line with the security standard the rest of the codebase already meets.

### Additional note (lower priority)

API keys (Anthropic, OpenAI, etc.) stored via `electron-store` in `ai-settings` are written as plaintext JSON, while other credentials (Stytch tokens, sync keys) properly use `safeStorage`. Migrating API key storage to `safeStorage` would close that gap as well.

### Environment

[...]

PR #526, opened 2026-06-02
fix(httpServer): require bearer token auth on /clip endpoint

### LINKED ISSUE #525: Unauthenticated /clip endpoint allows cross-origin file writes to workspace
## Summary

The `/clip` endpoint in `packages/electron/src/main/mcp/httpServer.ts` accepts unauthenticated POST requests and responds with `Access-Control-Allow-Origin: *`. This means any website the user visits can silently write arbitrary markdown files into the user's active workspace directory via a cross-origin request.

The `/mcp` endpoint on the same server is already protected by `requireMcpAuth()` (a per-launch bearer token check added in issue #146), but `/clip` was intentionally left open. This creates a meaningful security gap: a malicious webpage can inject files into the workspace without user interaction.

## Impact

- **CWE-284 (Improper Access Control)**: The endpoint performs a filesystem write (creating `.md` files in the workspace) without any authentication.
- An attacker who knows or guesses the port can craft a webpage that, when visited by a Nimbalyst user, writes attacker-controlled content into their workspace.
- The written files appear in the workspace file tree and could be used to inject misleading content, phishing material, or—depending on downstream tooling—executable payloads.

## Reproduction

1. Start Nimbalyst with the HTTP server active (default behavior when MCP is enabled).
2. Open a workspace.
3. From any browser tab, run:

```javascript
// PoC: write a file into the victim's workspace from any webpage
fetch('http://localhost:22140/clip', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({
    title: 'injected-by-attacker',
    url: 'https://evil.example.com',
    content: '# You have been clipped\n\nThis file was written by a webpage you visited.'
  })
});
```

4. Observe that `injected-by-attacker.md` (or similar) appears in the workspace's clips directory with no user prompt or authentication.

## Suggested fix

Extend the existing `requireMcpAuth()` gate to cover `/clip` in addition to `/mcp`, and allow the `Authorization` header in the CORS preflight so the browser extension can send the bearer token. This is a ~5-line change.

sebastionoss:
## Summary

Closes #525

The `/clip` endpoint in `packages/electron/src/main/mcp/httpServer.ts` accepts unauthenticated POST requests and responds with `Access-Control-Allow-Origin: *`, allowing any website to write markdown files into the user's active workspace via a cross-origin fetch. This PR extends the existing `requireMcpAuth()` bearer-token gate (already protecting `/mcp` since #146) to also cover `/clip`.

## Related issue

Closes #525 — Unauthenticated `/clip` endpoint allows cross-origin file writes to workspace.

## Testing

- Verified that unauthenticated `POST /clip` now returns `401 Unauthorized`.
- Verified that `POST /clip` with a valid `Authorization: Bearer <token>` header succeeds and creates the clip file as before.
- Verified that the CORS preflight (`OPTIONS /clip`) now includes `Authorization` in `Access-Control-Allow-Headers`, so the browser extension can send the token.
- Confirmed the `/mcp` endpoint behavior is unchanged.

## Notes for reviewers

**Vulnerability details (CWE-284 — Improper Access Control)**

The HTTP server binds to localhost and serves two endpoints: `/mcp` (for MCP protocol messages) and `/clip` (for a browser extension to save web clippings as markdown files). After #146, `/mcp` requires a per-launch bearer token, but `/clip` was intentionally left open. Combined with `Access-Control-Allow-Origin: *` on the `/clip` response, this means any webpage the user visits can silently POST to `http://localhost:<port>/clip` and write a file into the active workspace.

**Proof of concept**

From any browser tab while Nimbalyst is running with a workspace open:

```javascript
fetch('http://localhost:22140/clip', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({
    title: 'injected-by-attacker',
    url: 'https://evil.example.com',
    content: '# You have been clipped\n\nThis file was written by a webpage you visited.'
  })
});
```

This creates `injected-by-attacker.md` in the workspace clips directory with no user interaction or authentication. The attacker needs to know or guess the port (default 22140).

**What the fix does**

The change is minimal — one condition is widened from `pathname === "/mcp"` to `(pathname === "/mcp" || pathname === "/clip")` in the auth guard at line 536, and the CORS preflight for `/clip` adds `Authorization` to the allowed headers so the browser extension can authenticate. No new dependencies, no behavioral changes for authenticated callers.

**Adversarial review**

Before submitting, we considered whether this is actually exploitable given that the server binds to localhost. It is: the `Access-Control-Allow-Origin: *` header explicitly permits cross-origin requests from any webpage, and browsers will happily send a `POST` with `Content-Type: application/json` to localhost. The same-origin policy does not protect localhost services that opt into CORS with `*`. The port is predictable (default 22140), and even if randomized, a scanning loop over a small range is trivial. The file-write impact is real — files land in the user's workspace and appear in the file tree.

### DIFF
--- a/packages/electron/src/main/mcp/httpServer.ts
+++ b/packages/electron/src/main/mcp/httpServer.ts
@@ -517,7 +517,7 @@ async function tryCreateServer(port: number): Promise<any> {
             res.writeHead(200, {
               "Access-Control-Allow-Origin": "*",
               "Access-Control-Allow-Methods": "POST, OPTIONS",
-              "Access-Control-Allow-Headers": "Content-Type",
+              "Access-Control-Allow-Headers": "Content-Type, Authorization",
             });
           } else {
             res.writeHead(200, {
@@ -530,10 +530,10 @@ async function tryCreateServer(port: number): Promise<any> {
           return;
         }
 
-        // Issue #146: every non-OPTIONS request to /mcp must carry the
-        // per-launch bearer token. /clip stays open below (intentional, per
-        // plan: web-clipper extension fires from the user's browser).
-        if (pathname === "/mcp" && !requireMcpAuth(req)) {
+        // Issue #146: every non-OPTIONS request to /mcp or /clip must carry
+        // the per-launch bearer token. The browser extension must be configured
+        // with the token to authenticate clip requests.
+        if ((pathname === "/mcp" || pathname === "/clip") && !requireMcpAuth(req)) {
           res.writeHead(401);
           res.end("Unauthorized");
           return;


ghinkle:
Thanks for the report and the proposed fix. I agree with the underlying security issue, but this would break the web-clipper functionality. Instead I've pushed a fix that restricts /clip to extension-origin application/json requests so arbitrary web pages can’t post to it.

The fix is in aed978e96f7af0a4d811c6bb1c6eb281e0b6c277

I’m going to close this PR in favor of the in-repo fix.

Thanks!

PR #534, opened 2026-06-03
feat: refresh button in Files Mode sidebar header (#259)

### LINKED ISSUE #259: [Feature]: Create a button to refresh the files.

I currently need to go to the Developer menu to access the “Refresh File Tree” option.

Image

If there were a button in Files Mode, it would make it much easier to reload the file list.

Image

My current version is 0.59.2.

Image

omartelo:
## Summary
- Adds a refresh icon button next to "New File" in the Files Mode sidebar header that reloads the file tree from disk, removing the need to open the Developer menu (closes #259).
- Bumps the sidebar minimum width from 150 to 200 so the extra icon fits without collapsing the workspace name.
- Registers a `HelpTooltip` entry for the new button (`file-tree-refresh-button`) so hover help is consistent with the other header icons.

The button reuses the existing `refreshFileTree(workspacePath)` helper (`store/listeners/fileTreeListeners.ts`), the same code path the Developer menu "Refresh File Tree" item already triggers. No new IPC, no state lifting.


ademczuk (commented):
Nice, small, well-scoped change. The button reuses the existing `handleRefreshFileTree` callback (`WorkspaceSidebar.tsx:200-203`), which delegates to the central `refreshFileTree(workspacePath)` helper in `store/listeners/fileTreeListeners.ts` -- the same helper the file-create and folder-create flows already call via that callback (`:389`, `:412`). No new IPC, no state lifting, and error handling already lives in the helper's try/catch (`fileTreeListeners.ts:63-66`). Accessibility is covered: `aria-label`, `title`, `data-testid`, and a matching `HelpContent` entry. CI is green on all three checks (Unit Tests, TypeScript Check, Transcript Bundle). This is what #259 asked for.

Two questions, neither blocking:

1. The summary says the new button is "consistent with the other header icons." On the PR head, New File (edit_square) and New Folder (create_new_folder) are bare buttons with no `HelpTooltip`, while Quick Open (search) and Filter (filter_alt) are `HelpTooltip`-wrapped. The refresh button matches the Quick Open / Filter pattern, so it lines up with 2 of the 4 existing icons rather than all of them. Just a wording note -- the choice itself is reasonable.

2. The min-width bump from 150 to 200 only updates the drag-resize floor (`EditorMode.tsx:830`). The persisted-width restore (`EditorMode.tsx:862-864`) does `if (savedWidth && typeof savedWidth === 'number') setSidebarWidth(savedWidth)` with no clamp to the new range. A user who previously resized to 150-199 (valid under the old floor) would load that old value unclamped until they drag again. The effect is cosmetic (the workspace name could collapse for that one user), and the restore path already had no clamp for the upper bound either, so it's not a regression this PR introduces. Worth a clamp on load (`setSidebarWidth(Math.min(Math.max(200, savedWidth), 500))`)? Up to you whether it's in scope here.

One process note for the maintainers, not the author: the branch shows as having merge conflicts with main right now, so it'll need a rebase before it can land.

PR #547, opened 2026-06-04
fix: default OpenAI models to GPT-5.5

yimizilu:
## Summary

The default OpenAI model values still pointed to GPT-5.4 even though GPT-5.5 is available and intended to be the current default.

This change updates the default model values for:
- `openai`
- `openai-codex`
- `openai-codex-acp`

It also updates the related default-value tests and adds an Unreleased changelog entry.

## Validation

- `npm run test:unit -- packages/runtime/src/ai/server/__tests__/ModelIdentifier.test.ts packages/runtime/src/ai/server/providers/__tests__/OpenAICodexProvider.test.ts`
- `73 passed, 1 skipped`

## Risk / Impact

Low risk.

This change only updates default model selection paths for new sessions or default resets. It does not remove GPT-5.4 support, and it does not change compatibility mappings for legacy aliases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Terminal scrollback is preserved when it contains a stray NUL byte, instead of discarding all saved history.
 - Claude Code session token totals in the AI Usage Report are no longer inflated.
 - Tracker tool widgets no longer crash on the SQLite backend over a JSON-string `type_tags` column.
+- Default OpenAI model selections for `openai`, `openai-codex`, and
+  `openai-codex-acp` now point to GPT-5.5 instead of GPT-5.4.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/services/ai/__tests__/aiServiceUtils.test.ts
+++ b/packages/electron/src/main/services/ai/__tests__/aiServiceUtils.test.ts
@@ -7,10 +7,20 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 // and the two const arrays.
 vi.mock('@nimbalyst/runtime/ai/server', () => ({
   OpenAICodexProvider: {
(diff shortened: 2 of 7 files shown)


ademczuk (commented):
Thanks for this, @yimizilu. Defaulting OpenAI to GPT-5.5 reads correct: 5.5 is the newest entry in OPENAI_MODELS (modelConstants.ts:93, listed above 5.4) and it was wired into the catalogs deliberately in 19becd154, so promoting it to the default is a natural follow-up. The scoping is clean too: DEFAULT_MODELS plus the two provider fallbacks, tests updated, 5.4 kept selectable, no legacy-alias remapping touched. Low regression risk looks right since new sessions read DEFAULT_MODELS at read-time rather than migrating stored model IDs.

One thing I wanted to ask about rather than assume. There looks to be a second alias-resolution path the diff did not touch. OpenAICodexProvider.normalizeModelSelection (packages/runtime/src/ai/server/providers/OpenAICodexProvider.ts:512) still returns openai-codex:gpt-5.4 at line 515 for the LEGACY_MODEL_ALIASES set defined at lines 487-494, which includes default, cli, and openai-codex-cli. That is the same alias set that getConfiguredModel (the method you updated at line 1778) now resolves to gpt-5.5. normalizeModelSelection is reachable on a live path via extractModelForProvider in packages/electron/src/main/services/ai/aiServiceUtils.ts:245.

I have not traced which path wins, and a brand-new session reads DEFAULT_MODELS directly (now 5.5), so this may be harmless in practice. But a codex session whose stored model is the default or cli alias could still resolve to 5.4 through normalizeModelSelection while getConfiguredModel sends 5.5, which would be inconsistent. Worth a second look, or a one-line note in the PR if that 5.4 return is deliberate.

Minor: Greg has asked for CHANGELOG entries to be a single line ending with the issue/PR reference. Easy to trim if you want to match that.

CI: I cannot see a green rollup yet (statusCheckRollup is empty, mergeStateStatus UNSTABLE), so it may just be waiting on required checks to register.

PR #557, opened 2026-06-04
fix: automation session lands in active Multi-Project, not Project 1

### LINKED ISSUE #544: Multi-Project: automation session created in wrong project (always lands in Project 1)
## Summary
When Multi-Project is enabled and an automation is manually run from a non-default project, the resulting session is created in Project 1 (the startup default) instead of the active project.

## Steps to reproduce
1. Enable Multi-Project feature.
2. Switch to a project other than the default startup project (e.g. Project 3) via the Multi-Project switcher.
3. Open the Automations panel in Project 3 and manually run an automation.
4. Observe where the new session appears.

## Expected behavior
The session created by the automation appears in Project 3 (the currently active project where the automation was triggered).

## Actual behavior
The session appears in Project 1 (the default startup project). This happens every time.

## Environment
- App version: 0.63.9
- OS: darwin arm64 (release 24.6.0)
- Electron: 41.0.4
- Node: 24.14.0
- Chrome: 146.0.7680.153

## Logs
Not collected (no relevant entries found in main process log).

### Nimbalyst version

_No response_

### Operating system

None

omartelo:
## Summary

Fixes #544. In Multi-Project mode, manually running an automation from a non-default project created the session in Project 1 (the startup default) instead of the active rail project.

## Root cause

A single Electron window hosts N projects in Multi-Project mode. Switching projects in the rail calls `workspace:set-active`, which updates `windowState.activeWorkspacePath` and leaves `windowState.workspacePath` pinned to the window's primary/startup project.

The `extensions:ai-send-prompt` handler (used by the Automations extension's `ai.sendPrompt`) resolved the workspace from the raw `windowState?.workspacePath`, ignoring `activeWorkspacePath` — so every automation session landed in Project 1. A sibling defect existed in the openai-codex test-connection path.

## Changes

- Add `resolveActiveWorkspacePathForWindowId(windowId)` to `windowState.ts`, delegating to the existing `resolveActiveWorkspacePath` (`activeWorkspacePath ?? workspacePath`).
- Route `extensions:ai-send-prompt` through the new seam so extension/automation sessions are created in the active rail project.
- Fix the sibling openai-codex test-connection resolution to honor the active project.
- Document the fix in `CHANGELOG.md` under `[Unreleased] > Fixed`.

## Test plan

- [x] `windowState.test.ts`: added a red→green regression test for issue #544 (active rail project wins over the startup primary), plus null/unknown-id/fallback cases. 18/18 pass.
- [x] Related suites green: `MultiProjectRailHandlers.test.ts` (13), `ExtensionHandlers.test.ts` (4).
- [x] `tsc --noEmit`: clean for the changed files.


ghinkle:
I think AIService.ts still has one correctness gap in the sibling Codex test-connection fix. 

In ai:testConnection, the new logic still does windowStates.get(event.sender.id), but windowStates is keyed by Nimbalyst’s custom window IDs, not webContents.id. That means this path can still miss the active rail workspace and fall back incorrectly or fail to resolve a workspace at all. The extensions:ai-send-prompt change in this PR uses the right pattern already: BrowserWindow.fromWebContents(event.sender) -> getWindowId(...) -> resolveActiveWorkspacePathForWindowId(...). 

I think ai:testConnection needs to use that same lookup path.

omartelo:
Yeah, totally slipped past me here, you're right. `event.sender.id` is the `webContents.id`, not our window id, so the codex `ai:testConnection` lookup could've missed the active rail workspace (or failed to resolve one at all).

Fixed in b9b2e50, it now goes through the same `BrowserWindow.fromWebContents(event.sender)` -> `getWindowId(...)` -> `resolveActiveWorkspacePathForWindowId(...)` path as the `extensions:ai-send-prompt` handler, and I dropped the now-unused `resolveActiveWorkspacePath` import. Thanks for the catch!

omartelo:
Heads up on the red Unit Tests check: it's not a regression from this PR.

The failure is `OpenAICodexProvider > uses SDK-provided model discovery when available` timing out at 5000ms (`packages/runtime/src/ai/server/providers/__tests__/OpenAICodexProvider.test.ts:119`). That test is unrelated to the windowState / AIService changes here, and it passes locally in ~420ms. It just occasionally exceeds the 5s timeout under CI load. The other 349 tests pass, including the new windowState regression test for #544.

A re-run of the job should go green. I don't have permission to trigger it on this repo, so it'll need a maintainer re-run (or a fresh push).

PR #558, opened 2026-06-04
feat(extensions): Gemini (Antigravity) coding agent provider

### LINKED ISSUE #96: Gemini Coding Agent
## Summary
Google Gemini CLI as a first-class provider alongside Claude Code and Codex.

## Current stage
Roadmap (P2)

## Why this matters
This is part of Nimbalysts early-access and roadmap feature set, and we would like community help moving it forward.

## What would help advance this
- implementation work
- UX / product feedback from real usage
- testing and bug reports
- documentation and onboarding improvements where needed

## Contribution note
If you want to work on this, comment on the issue first so maintainers can point you at the right code paths, current constraints, and any related work already in progress.

ademczuk:
## Summary

Adds Google Gemini (via the Antigravity language server, running Gemini 3.5 Flash) as a first-class **agent** provider, contributed as a marketplace extension under `aiAgentProviders` per the direction in #96. The extension rides the `backendModules` utility-process system (437c12106): the Antigravity language-server lifecycle, the simulated tool-call loop, and usage metering all live inside the extension's backend module, not core runtime. The IDE version pin is a manifest config field, not hardcoded.

What it brings:

- Gemini as an agent provider you can pick alongside Claude Code and Codex.
- Meta-agent support: a Gemini session can run as a Meta Agent and orchestrate worker child sessions (create_session / get_session_status / send_prompt / respond_to_prompt). Tool nesting and recursion are gated the same way the built-ins are (the meta-agent tool allowlist omits spawn_session so a child cannot recurse).
- Phase-1 read-only dev tools for a standard Gemini session (read_file, list_files, search_files) through the same simulated tool loop, so Gemini can investigate a codebase. They dispatch over a host broker method gated on the minimal `workspace-files` permission, with the workspace jail pinned host-side and a symlink-escape guard.
- A usage indicator chip surfacing Antigravity account credits and per-model quota.

Auth rides the existing local Antigravity / `~/.gemini` login; no API key is entered in Nimbalyst.

## Related issue

Closes #96. @karlwirth @ghinkle this is the branch you asked me to put up so you can test and respond. It follows the guidance from the issue thread: `aiAgentProviders` (not `aiProviders`), AntigravityServerManager inside the extension backend module, IDE version pin as a manifest config field. I dropped the chat-only half and shipped agent-only, as discussed.

## Testing

- typecheck clean across runtime, electron, and the extension.
- Extension backend stream tests pass; meta-agent broker permission tests pass.
- Manual smoke test of the Meta Agent: a Gemini meta-agent named its session, spawned a worker child via create_session, and read the child's status through to completion.

One honest note on the screenshots below: the worker child ran as a claude-code session, because create_session's provider enum is `[claude-code, openai-codex]`. So what they show is the Gemini meta-agent **orchestrating** a worker, not Gemini running the read-only dev tools itself. The Gemini dev tools are exercised by the standard-session path and covered by the extension build plus typecheck.

Gemini Meta Agent orchestrating a worker: it named its session, spawned a worker via create_session, and tracked it in the Delegated Sessions panel

The spawned worker session completing its directory-listing probe (ran as a claude-code worker)

## Notes for reviewers

- **Bundled `.nimext` binary**: the first commit commits `packages/electron/resources/bundled-extensions/gemini-antigravity.nimext` (43KB built artifact; its sha256 matches the extensionRegistry entry) so you can install and test the branch directly. Upstream tracks no such binaries, so flagging it as an open question. Happy to drop it and build from source if you'd prefer.
- **package-lock.json** shows a large net reduction. The extension declares no new runtime dependencies (its `dependencies` is empty; devDeps and peerDeps are already in the monorepo), so this is purely an `npm install` dedup of redundant nested entries, not feature-required. Happy to reset it to main's lockfile if you would rather keep it out of the diff (the new workspace still needs its own lock entry for `npm ci`).
- **Dev-only permission auto-grant**: `PrivilegedExtensionHost` adds a branch that auto-grants a backend module's declared permissions when `NIMBALYST_ALLOW_DEV_BACKEND_MODULES=1` AND the build is non-packaged. It is double-gated and inert in a packaged build (production always shows the consent prompt); it only skips the prompt for dev-loaded extensions during development. Flagging it since it touches the grant path.
- **create_session provider enum gap**: the enum is `[claude-code, openai-codex]`, so a Gemini meta-agent cannot name a Gemini child directly through it. It either picks a built-in or relies on no-provider inheritance (the inheritance path has test coverage). Worth deciding whether the enum should grow to include extension agent providers; not addressed here.
- **Cross-workspace typecheck fix**: adds `packages/extensions/math/src/runtime-shim.d.ts`, an ambient `@nimbalyst/runtime` declaration so the math extension typechecks without a built runtime dist (the workspace typecheck surfaced it). Small and unrelated to Gemini, included so the workspace typecheck stays green.
- Commits are DCO signed-off and map to coherent steps (extension scaffold, agent and meta-agent wiring, usage chip, typecheck and doc, provider-inheritance fix, gitignore and CHANGELOG, read-only dev tools). Happy to squash on merge.


ghinkle:
Thanks for this! Looks solid. A couple points:

The SDK validator and the new manifest/type seem to disagree on name vs displayName.

manifestValidation.ts expects name and model.provider
the new type + shipped manifest use displayName and no per-model provider


Maybe we should just have one provider per extension to keep it simple? Are there scenarios where you see multiple being useful?

Also, it looks like the renderer only registers extension provider ids with ModelIdentifier at app startup. Probably should fix that for user experience.

Finally, I'd be ok with getting rid of the agent provider enum in create session so this can work first class. I think there's a few places we fuzzy match those also that might need to be checked.

ademczuk:
Thanks Greg. Your four, in order.

1. name vs displayName / model.provider. You're right, it fails its own validator, and both halves are the validator being out of step with the SDK type, so the fix stays inside `manifestValidation.ts`:
   - it checks `provider.name` (line 433), but the type and the manifest both use `displayName`, which the type itself documents as the dropdown label (`extension.ts:102`). So the validator should accept `displayName` rather than the manifest getting renamed.
   - it requires a per-model `provider` (line 524), but nothing downstream reads it: `AIService` builds the model list from `m.id`/`m.name` and sets the provider from the contribution id, and `AiAgentProviderModel` doesn't even declare a `provider` field. So I'll drop that requirement.
   Net it's an align-the-validator change, no manifest or type churn.

2. One provider per extension. I'll lock it to one for now. The cases I'd watch are one backend behind two endpoints (Gemini API vs Vertex, or local vs remote), but none of those are in front of us, and one-to-many later is a clean relaxation that doesn't break existing extensions, where the reverse would. One guard so we don't box ourselves in: I'll keep the provider id a separate field from the extension id rather than assuming they're equal, so lifting the limit later is just dropping a count check.

3. Startup-only registration. Agreed, that's the UX bug. I'll register on extension load/re-scan instead of only at startup, and handle the disable side too: deregister, and reset a selected model if its provider goes away.

4. Dropping the agent-provider enum. Happy to, that makes it first class. I'll hit the fuzzy-match sites you flagged plus the two that reach further: persisted sessions that stored an enum value, and anything that treated the enum as exhaustive in a switch.

Scope: 1 and 3 are contained, 4 and the one-per-extension shape ripple through more files. To keep this reviewable I'd land 1 and 3 here and do 4 plus the provider-shape lock as a focused follow-up. Say the word if you'd rather it all land together.

ghinkle:
Thanks! How ever you prefer to land those is fine with me. 1 and 3 here is fine.

ademczuk:
Rebased onto main and landed 1 and 3 (CI green).

1. Validator now checks `displayName` instead of `name` and drops the unused per-model `provider`, so it matches the `AiAgentProviderContribution` / `AiAgentProviderModel` types. Added a test on the shipped contribution shape.

3. Provider ids register on every extension load / re-scan / unload now, not just at startup, so a provider enabled after launch is recognized without a relaunch. I'd planned to also reset the saved default model when its provider goes away, but that races with async startup loading - a non-agent extension loading first would wipe a still-valid default before the agent one finishes loading - so I kept just the self-healing registration.

Two pre-existing things I left out of scope, happy to take either as a follow-up: the main-side `registerAgentProviderContributions` never clears an extension's prior entries on re-scan (so a dropped provider stays selectable through `ai:getModels`; `clearAll(extensionId)` already exists), and `MetaAgentService` can store a provider/model mismatch when a caller passes `provider` without `model`.



@ghinkle good to go from your side whenever suits. Mergeable with CI green since the 6th, points 1 and 3 landed. No rush.

ademczuk:
Simply made a cosmetic change to ensure that in settings the gemini switch was automaticaly on if detected and to ensure the chip was valid. PR ready to merge

ademczuk:
Updated in alignment to commit ea8ee3f767 │ Release v0.65.3

On the Gemini side I checked the report's specifics against the actual checkout. The `file:line` references land on the exact lines (`lib/pipeline.py:8` for the ThreadPoolExecutor import, `pipeline.py:68` for `MOCK_AVAILABLE_SOURCES`, `last30days.py:933` for the quality-nudge hook), and the env vars, CLI flags, and `schema.Report` are all real. One slip: it listed `--emit` choices as "text, json, html" when the real set is `compact, json, context, md, html`. That is fixed by the verifier change in this PR (044530f30). Commit 145fbaa03 lets the Gemini agent actually write files by accepting a no-escape
  <<<WRITE_FILE: path>>> … <<<END_WRITE_FILE>>> directive, so it stops narrating "I saved the file" without ever emitting a write_file call.

PR #561, opened 2026-06-05
fix: restore all projects on reload in multi-project mode

### LINKED ISSUE #530: Reloading page in multi-project mode closes all but one project
## Summary
In multi-project mode with multiple projects open, reloading the page (Command+R) causes all projects except one to be automatically closed.

## Steps to reproduce
1. Enable multi-project mode
2. Open two or more projects in the workspace
3. Reload the page (Command+R)
4. Observe: only one project remains open

## Expected behavior
All open projects should persist across a page reload, restoring the same multi-project layout as before.

## Actual behavior
Only a single project survives the reload; all other previously open projects are closed and must be manually reopened.

## Environment
- App version: 0.63.9
- OS: macOS 26.3 (darwin arm64)
- Electron: 41.0.4
- Node: 24.14.0

## Logs
Not collected.

### Nimbalyst version

0.63.9

### Operating system

None

omartelo:
## Description

Fixes #530. In multi-project mode with 2+ projects open, a renderer reload (Cmd+R) collapsed the rail to a single project — every other open project was closed and had to be reopened manually.

## Root cause

A reload reuses the same `BrowserWindow`, so the main-process `windowStates` map stays warm: `state.additionalWorkspacePaths` still holds every extra project. But the renderer re-seed path only restored the primary:

- `get-initial-state` returned only `workspacePath`, never `additionalWorkspacePaths` / `activeWorkspacePath`, so the renderer had no per-window source to rebuild the rail.
- `loadInitialState` seeded just the primary into the rail.

Because `windowStates` survives a reload but resets on a cold launch, the presence of `additionalWorkspacePaths` already distinguishes refresh from cold launch — no extra flag needed.

## Changes

- `get-initial-state` now exposes the window's full project set (`additionalWorkspacePaths` + `activeWorkspacePath`).
- New `seedRailFromInitialState` helper re-seeds the rail with the primary plus every warm project, then restores the in-view active path (not whichever was added last).
- `loadInitialState` routes through the helper. On cold launch the additional set is empty, so only the primary is seeded — unchanged behavior.

## Test plan

- [x] TDD red->green: 5 new unit tests in `openProjects.test.ts` cover the reload repro plus edge cases (active in-view wins, primary fallback, cold launch, blank/dup paths). 23/23 pass.
- [x] Related suite green: `MultiProjectRailHandlers.test.ts` (13/13).
- [x] `tsc --noEmit`: clean.

### DIFF
--- a/packages/electron/src/main/ipc/WindowHandlers.ts
+++ b/packages/electron/src/main/ipc/WindowHandlers.ts
@@ -26,10 +26,17 @@ export function registerWindowHandlers() {
         if (!state) return null;
 
         if (state.mode === 'workspace' && state.workspacePath) {
+            // Expose the full window project set so the renderer can rebuild
+            // the multi-project rail after a reload. `windowStates` survives
+            // a renderer reload but resets on cold launch, so a populated
+            // `additionalWorkspacePaths` already signals "this was a refresh"
+            // — the rail restores every warm project, not just the primary (#530).
             return {
                 mode: 'workspace',
                 workspacePath: state.workspacePath,
                 workspaceName: basename(state.workspacePath),
+                additionalWorkspacePaths: state.additionalWorkspacePaths ?? [],
+                activeWorkspacePath: state.activeWorkspacePath ?? state.workspacePath,
             };
         }
 
--- a/packages/electron/src/renderer/App.tsx
+++ b/packages/electron/src/renderer/App.tsx
@@ -125,7 +125,7 @@ import { ProjectRail } from './components/ProjectRail';
 import {
   activeWorkspacePathAtom,
   multiProjectModeAtom,
-  addOpenProjectAtom as addOpenProjectAction,
+  seedRailFromInitialState,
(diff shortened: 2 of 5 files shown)


omartelo:
Superseded. #530 is already fixed on `main` by 975a14d57 ("fix: preserve multi-project rail on reload", merged Jun 3), which lands the same fix across the same files (`WindowHandlers.ts`, `electron.d.ts`, `openProjects.ts`, `openProjects.test.ts`).

That commit's approach is more complete: `get-initial-state` exposes `openProjectPaths` plus `activeWorkspacePath`, and rail rehydration is reconciled in `initOpenProjects()` via `resolveInitialOpenProjectsState()` (window live-rail state vs persisted `restorePreviousProjects` vs fallback). This PR's `seedRailFromInitialState()` would be redundant and double-seed the rail.

Closing as superseded. No action needed here.

Note: 975a14d57 did not reference #530, so the issue is still open. #530 should be closed as already fixed on `main`.

PR #575, opened 2026-06-06
fix: allow mode='auto' in ai_sessions so SQLite migration works

omartelo:
## Problem

The PGLite -> SQLite migration dry-run aborts with:

```
Dry run failed: CHECK constraint failed: mode IN ('planning', 'agent')
```

## Root cause

The `ai_sessions.mode` CHECK constraint is stale, it predates `mode='auto'`, introduced by the "Allow All" permission feature in #379 (sends `permissionMode: 'auto'` to Claude Code so the classifier gates tools instead of Nimbalyst). `SessionMode = 'planning' | 'agent' | 'auto'`, but both DB CHECK constraints only allowed `planning`/`agent`.

PGLite has no CHECK on `mode`, so `auto` rows live there fine. The strict SQLite target schema rejects them on copy, aborting the whole dry-run.

## Fix

Add `'auto'` to the allowed set in both CHECK constraints:

- `packages/electron/src/main/database/sqlite/schemas/0001_initial.sql` (SQLite target schema, fixes the dry-run)
- `packages/electron/src/main/database/worker.js` (PGLite migration, consistency for fresh installs)

## Test

`MigrationOrchestrator.fixtureRoundtrip.test.ts` now seeds a `mode='auto'` session. Verified red -> green:

- old 2-value CHECK -> reproduces the exact error `CHECK constraint failed: mode IN ('planning', 'agent')`
- new 3-value CHECK -> passes

## Related

- #379 the "Allow All" permission feature that introduced `mode='auto'`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - PR review cache tables are now created on the better-sqlite3 backend too (migration registered with the SQLite runner), not only on PGLite. (#307)
 - PR review: merged PRs now show as "Merged" (not "Closed") in the list, the Approve/Merge buttons refresh immediately after a merge (cache-bypassed refetch), and a success notice / "Merged" badge confirms the action. (#307)
 <!-- Bug fixes go here -->
+- SQLite migration dry-run no longer aborts on sessions created in `Allow All` permission mode: the `ai_sessions.mode` CHECK constraint now allows `auto` alongside `planning`/`agent`. (#379)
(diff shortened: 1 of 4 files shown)


ghinkle:
I'm not sure i understand this as I thought we moved away from a third state on sessions in favor of the permissions / trust layer. 

Maybe you have some old data from before the migration that is causing a failing migration?

omartelo:
Dude, I was so obsessed with thinking it was a bug in my implementation that I ignored the fact that I had legacy information in my database; in fact, I had some lines with "auto," it was my mistake, sorry.

PR #584, opened 2026-06-06
feat: create a Meta Agent from iOS (Phase 1)

porthex:
## Phase 1 — Create a Meta Agent from iOS

First of three phases bringing the desktop "Meta Agent" experience to mobile. This phase rides the existing `createSession` sync rails: the phone can create a session with `agentRole: "meta-agent"`, which is sufficient for the desktop runtime to grant meta-agent powers (the agent role is re-read every message send and swaps in the meta-agent MCP profile + tool set).

### What's included
- **Wire-through of `agentRole`** on the create path: `EncryptedCreateSessionRequest` (collab-protocol), `CreateSessionRequest` (runtime), `CollabV3Sync` decode + encode, and the desktop mobile-create handler (`AIService`) which now passes `agentRole` to `SessionManager.createSession` (defaulting to `'standard'`).
- **Alpha gating**, mirroring desktop's `meta-agent` feature flag: a new `metaAgentEnabled` field on the existing `SyncedSettings` sync channel. Desktop reads its `meta-agent` alpha flag and syncs it down; iOS persists it via a new `FeaturePreferences` helper and gates the UI on it.
- **iOS UI**: a gated "New Meta Agent" entry in the session create menu that creates a session with `agentRole: "meta-agent"`. Analytics: `mobile_meta_agent_created`.
- **Tests** (`MetaAgentCreateTests.swift`): the create request encodes `agentRole`, `SyncedSettings` decodes `metaAgentEnabled` (and back-compat nil), and `FeaturePreferences` round-trips.
- CHANGELOG + POSTHOG_EVENTS docs updated.

Loops/Super Loops are explicitly out of scope (deferred). Phases 2 (sync `agentRole`/`createdBySessionId` down to mobile) and 3 (collapsible group UI) will follow as separate PRs.

### Verification
- CI is the verification surface here: macOS Swift tests (`ios-transcript-tests.yml`) + TS tsc/unit (`ci.yml`).
- **Reviewer-verified, not CI-verified**: the true end-to-end flow (phone taps "New Meta Agent" -> desktop spawns a meta-agent session -> it can spawn sub-agents) requires a paired device + desktop and cannot be exercised in CI.


ademczuk (commented):
Took a look from the desktop side, since this PR touches `AIService.ts` and the shared sync types, which is an area I have worked in (meta-agent session creation). Sharing what I verified in case it saves a reviewer some time. I am a triager here, not a maintainer, so treat this as input rather than a verdict.

The shared/desktop change reads as sound and back-compat:

- The new positional arg in `AIService.ts` lines up exactly. It passes `'agent'` (mode), three `undefined` worktree args, then `resolvedAgentRole`. Checked against the signature at `SessionManager.ts:851`: `(provider, documentContext, workspacePath, providerConfig, model, sessionType, mode, worktreeId, worktreePath, worktreeProjectPath, agentRole = 'standard', createdBySessionId)`. So the added args map to positions 7 through 11 with no silent shift.
- It is back-compat. `request.agentRole || 'standard'` plus the existing `agentRole = 'standard'` default means an existing mobile create request (no `agentRole`) behaves exactly as before, and the desktop create path is not affected. No regression risk to existing sessions or desktop meta-agent.
- The cast is valid. `types.ts:268` has `export type AgentRole = 'standard' | 'meta-agent'`, so both the wire value and the iOS `"meta-agent"` literal are real members.
- It is functional, not just a label. `agentRole` flows into `BaseAgentProvider.getAgentRole()`, which `ClaudeCodeProvider` and `OpenAICodexProvider` read, so a mobile-created `meta-agent` session actually gets meta-agent behavior.
- Naming and back-compat match the repo conventions. `agentRole` and `metaAgentEnabled` are camelCase on the wire, and `metaAgentEnabled?` being optional means an older desktop that omits it leaves mobile gated off. The included test covers that absent-flag case.

One forward note, not for this PR: `createSession` is now an 11-plus positional-arg call, and this change had to pad three `undefined`s to reach `agentRole`. A later move to a named-options object would make this kind of addition harder to get wrong. Pre-existing, just flagging.

Heads up that CI has not run on this yet (the workflow is sitting in the fork-PR approval gate), so build and unit results are not confirmed by CI. The above is from reading the diff against `main`. Clean Phase 1.

PR #603, opened 2026-06-09
fix: stabilize Android transcript sync for large sessions

jbarbre:
## Summary
- Adds synced agent activity metadata so Android can show desktop agent status in the transcript.
- Improves Android transcript rendering for large sessions by tail-loading recent history, paging older history on demand, and caching cursor-based history pages locally for faster repeat scrollback.
- Adds Android queued-prompt controls and prevents stale execution state from pinning the thinking indicator.
- Hardens desktop transcript/history paths to avoid expensive full-session scans on oversized conversations.
- Ports the desktop ASAR transcript-scan guard into source by using bounded raw-tail loads for raw-backed sessions and capping raw transformer adapter batches.
- Adds desktop paged transcript history so large sessions can scroll older messages without reloading the full raw log.
- Stabilizes transcript view-message IDs for raw history pages and reconciles desktop DB tail pages with live streaming events without duplicating or reordering messages.
- Keeps mobile oversized-session transcript tails current when desktop turns finish, including storage-limit fallback sessions.
- Syncs desktop Claude and Codex usage snapshots to Android settings for mobile visibility.
- Adds canonical Claude Fable handling for Claude Code SDK and CLI sessions.

## Notes
This is one combined PR because the agent-status changes depend on the earlier mobile sync/transcript stability changes. Applying the status commit alone conflicts across Android sync, transcript rendering, Electron transcript stores, and runtime sync routing.

Reviewers may still choose to split follow-up work into smaller patches around status metadata, mobile transcript paging, queued prompt controls, transcript performance hardening, desktop transcript paging, usage sync, and model catalog updates.

Android scope note: the Android behavior changed here was ported/adapted from the iPhone app's transcript/session model into the existing Android app, especially transcript rendering, paged session history, agent status display, and queued prompt controls.

## Testing
- npm run typecheck --workspace packages/runtime
- npm run typecheck --prefix packages/runtime
- npm run typecheck --prefix packages/electron
- npx vitest run packages/electron/src/renderer/store/__tests__/transcriptStreamAccumulator.test.ts
- npm run build:transcript --prefix packages/android
- npm run sync:transcript-assets --prefix packages/android
- ./gradlew :app:compileDebugKotlin --rerun-tasks
- ./gradlew :app:testDebugUnitTest
- ./gradlew :app:assembleDebug
- Android debug APK installed and smoke-tested over ADB
- npm run build:unpack --prefix packages/electron (Node 22)
- Electron unpackaged build completed packaged SDK validation

## Privacy
No local device configuration, launcher paths, APK artifacts, or Google services credentials are included.


ademczuk (commented):
Reviewed this from the desktop and shared side, since most of the change is in `CollabV3Sync`, the agent-message stores, and the transcript pipeline (`SessionManager`, `MessageStreamingHandler`), not only the Android app. I am a triager here, so treat this as input, not a verdict. CI has not run yet (fork approval gate), so build and tests are not CI-confirmed; everything below is from reading the diff against `main`, with the Room and pagination angles checked against the official Android docs.

## Room migration (read first): correct, with two follow-ups

The v1 to v2 migration is sound. `NimbalystDatabase.kt` bumps the version and registers `MIGRATION_1_2` via `.addMigrations(...)`; it runs four `ALTER TABLE sessions ADD COLUMN` statements (`agentStatusKind` / `agentStatusLabel` / `agentStatusDetail` as TEXT, `agentStatusUpdatedAt` as INTEGER) that match the four new nullable `SessionEntity` fields exactly. Existing installs upgrade without data loss.

Two non-blocking follow-ups:

- `.fallbackToDestructiveMigration()` is still in the builder, unscoped. It will not fire for 1 to 2, but a future version bump that forgets a migration would silently wipe the local DB instead of failing loud. Now that real migrations exist, consider `fallbackToDestructiveMigrationFrom(...)` or dropping it. Blast radius is the mobile's local cache plus unsynced state (queued prompts, drafts) rather than the canonical record, which lives desktop and server side, so this is hygiene, not an emergency.
- No migration test ships with the bump. `build.gradle.kts` sets `room.schemaLocation` with `exportSchema = true`, but there is no committed `app/schemas/*.json` and no `MigrationTestHelper` test, so nothing guards the migration in CI, and a test cannot even be written until the schema JSON is exported and committed. A 1 to 2 `runMigrationsAndValidate` test that seeds a row and asserts it survives would lock this in.

## Sync protocol: back-compat looks good

`SyncedAgentStatus` and the new optional fields (`agentStatus?`, `mobileTranscriptTailJson`, `mobileTranscriptHistoryPageJson`) in `runtime/sync/types.ts`, plus their Android `SyncProtocol.kt` mirrors, are all additive and optional, so older desktop, iOS, and server peers are unaffected. Naming is camelCase throughout.

## One question: `syncableMetadata.ts` drops `draftInput`

The diff removes `'draftInput'` from `SYNC_RELEVANT_FIELDS.columns`. Since `columns` is documented as the set of fields forwarded to other devices, this stops draft input from syncing across devices entirely, not just from bumping sort order. Is that intentional (cutting draft-sync churn), and is it meant to ride along with this PR? Flagging in case it slipped in.

## Two efficiency notes on the desktop tail path (not correctness issues)

- `getAgentMessageCount` (`SessionManager.ts`): when `getMessageCounts` returns 5000 or fewer it fires a second `list(limit: 5001)` probe and takes the max. The comment explains it guards against store combinations that report a bad COUNT, which is fair, but it means every open of a session at or under the threshold loads up to 5001 rows even when the count is accurate (the common case), which partly reintroduces the full-load cost this PR removes for large sessions. Probing only when the count looks suspicious (for example, returns 0) would avoid that.
- `listBefore` fallback (`AgentMessagesRepository.ts`, `SyncedAgentMessagesStore.ts`): when the backing store lacks the optional `listBefore`, it loads up to 50,000 rows in memory and filters. Production's `PGLiteAgentMessagesStore` implements `listBefore` natively, so this is not hit today, but since `listTail?` / `listBefore?` are optional, a test stub or a future backend silently lands on the full scan. Promoting them to required interface members, or a `console.warn` in the fallback, would prevent a quiet regression.

## Nit

`agentMessageToRawMessage` uses `id: Number(message.id ?? 0)` (`SessionManager.ts`). Zero is falsy, so a downstream `if (msg.id)` guard would read id 0 as absent. Low risk with current callers.

## What looks solid

- The status-broadcast path (`MessageStreamingHandler`) pushes `type: 'metadata_updated'` only; there are no new writes to the raw `ai_agent_messages` log anywhere in the diff, so the append-only source-of-truth invariant holds.
- `PGLiteAgentMessagesStore.listTail` / `listBefore` and the Android `MessageDao` tail query both use the reverse-then-forward pattern (`ORDER BY id DESC LIMIT`, re-sorted ASC), keyed on a monotonic id, which is the right shape for stable cursor pagination.
- `QueuedPromptDao`'s added `AND sentAt IS NULL` is query-only against a pre-existing column, so no migration is needed there.
- New tests cover the load-older-history parsing, the `historyPage` and `agentStatus` payload building, and stale-status expiry.

Nice windowing design overall. The migration test plus schema commit, and scoping the destructive fallback, are the main things I would want to see before this lands.

ademczuk:
Correction to one point in my review above, please disregard the original suggestion on it.

On the `getAgentMessageCount` probe I suggested "probing only when the count looks suspicious (for example, returns 0)." That is wrong. The probe guards against a `getMessageCounts` that reports an inaccurately low but nonzero count on some store combinations (your comment notes exactly this). Gating the probe on `count == 0` would skip the case where the count is wrong but nonzero (say it returns 4000 for a 9000-message session), which would route a large session through the full-projection path and reintroduce the crash you are preventing. Your blanket "probe when count is at or under 5000" is the correct defensive choice given an untrustworthy COUNT.

The accurate version of the note: the probe is redundant overhead specifically on stores whose `getMessageCounts` is reliable. The production `PGLiteAgentMessagesStore.getMessageCounts` is an accurate `SELECT COUNT(*)`, so on that path the 5001-row probe never changes the answer and is pure per-open overhead. If you want to drop the redundant load, gate it on store capability (skip the probe when the backing store natively reports an accurate count) rather than on the count value. Not a blocker either way, and the current code is the safe option.

jbarbre:
Resolved the review follow-ups in the latest rebased commit (`be7482905`):

- Committed Room schema JSON for database versions 1 and 2.
- Added `NimbalystDatabaseMigrationTest` using `MigrationTestHelper` for the 1 -> 2 migration.
- Removed the unscoped Android `fallbackToDestructiveMigration()` path.
- Documented that `draftInput` / `draftUpdatedAt` are intentionally device-local to avoid cross-device typed-text churn.
- Added an accurate-count capability so production PGLite skips the defensive 5001-row probe, while untrusted stores keep the safe probe.
- Added one-time warnings for slow `listTail` / `listBefore` fallback paths.
- Changed missing raw-message id fallback from `0` to `-1` to avoid falsy ids.

Also included the validated Android structured prompt submission bridge fix and Android session/project list polish that were pending locally.

Validation run locally:

- `npm run typecheck --prefix packages/electron`
- `./gradlew :app:testDebugUnitTest`
- `./gradlew :app:compileDebugAndroidTestKotlin :app:assembleDebug`
- `PATH="$HOME/.nvm/versions/node/v22.22.3/bin:$PATH" npx vitest run packages/runtime/src/sync/__tests__/CollabV3Sync.routing.test.ts packages/runtime/src/ai/server/providers/__tests__/ProviderSessionManager.test.ts`

Note: full workspace `npm run typecheck` now reaches an unrelated upstream `@nimbalyst/extension-github-issues-importer` SDK export mismatch on current `main`; scoped Electron/runtime typechecks pass.

jbarbre:
Follow-up commit `d8945dcb6` resolves the two validation caveats from my prior comment:

- Full workspace `npm run typecheck` now passes. The GitHub issues importer now imports tracker importer types directly from the SDK type source instead of relying on stale SDK `dist` declarations.
- The connected Android migration test now passes on the Samsung device without clearing app data. `AnalyticsManager` now recovers from unreadable encrypted analytics preferences by using an analytics-only fallback instead of crashing app startup.

Additional validation run after the follow-up:

- `npm run typecheck`
- `./gradlew :app:testDebugUnitTest :app:compileDebugAndroidTestKotlin :app:assembleDebug`
- `ANDROID_SERIAL=100.109.203.96:5555 ./gradlew :app:connectedDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.nimbalyst.app.data.NimbalystDatabaseMigrationTest`

dmitry-aladzyeu:
Hi. Tested this build on a physical Android device (HyperOS/Xiaomi). The composer keyboard-overlap and cursor-jump fixes work well, and the larger session sync features (prompt send, agent status, live "Thinking…", transcript catch-up, queued prompts) all hold up.

While testing I hit one reproducible issue and have a small fix for it.

### Bug: open transcript stops live-updating after the periodic JWT refresh

With a session screen open, live `messageBroadcast` updates silently stop ~4 minutes in and don't resume until you back out and reopen the session.

**Root cause** is in `SyncManager.kt`. The proactive JWT refresh (`JWT_REFRESH_INTERVAL_MS = 4 min`) calls `disconnect()` then `connect()`. `disconnect()` clears `activeSessionId` and leaves the session room, but the proactive refresh path never sets `pendingSessionJoin`, so the index reconnect has nothing to re-join — the session room is never rejoined and incoming `messageBroadcast` is dropped (`handleMessageBroadcast` early-returns on `activeSessionId == null`).

A second, related gap: even when `pendingSessionJoin` *is* set, the deferred-join resume block calls `connectSessionClient(sessionId)` which does **not** restore `activeSessionId`, so `sessionClient.onConnectionStateChanged` skips the `syncRequest` and messages are still dropped.

**Logcat before fix** — JWT refresh at `22:58:40`, then ~4.5 min of zero `messageBroadcast`; catch-up (`sinceSeq=797`, 28 messages) only fires on manual reopen at `23:03:03`.

### Fix

Two small edits in `SyncManager.kt`:

1. In `refreshJwt()`, before the `disconnect()`/`connect()` teardown, stash the active room so the index reconnect re-joins it:
```kotlin
_state.value.activeSessionId?.let { pendingSessionJoin = it }
```

2. In the deferred-join resume block (`indexClient.onConnectionStateChanged`), restore `activeSessionId` before reconnecting the session client:
```kotlin
_state.update { it.copy(activeSessionId = sessionId) }
connectSessionClient(sessionId)
```

**Logcat after fix** — room rejoins automatically ~2s after refresh, no screen reopen needed:
```
23:21:38  JWT refreshed successfully
23:21:38  [indexClient] Resuming deferred session join: 52dc1b92...
23:21:40  [sessionClient] connection=true activeSessionId=52dc1b92...
23:21:40  [sessionClient] Sending syncRequest for 52dc1b92...
```

dmitry-aladzyeu:
Sorry for the cheek here — I know the proper way would be to open a separate GitHub issue. But I couldn't resist: I tried the mobile version and it really got me. I step away from my computer a lot, and I'd love to be able to keep some control over a running session from my phone.

I ran into a gap: when the agent is running (or if I accidentally send a wrong prompt from the phone while the session was idle), there's **no way to interrupt the running agent from mobile** — you have to get back to the desktop. And sometimes I'm genuinely far away — an hour's travel from my machine — and in that time a wrong prompt can do real damage. The desktop has a stop button; mobile doesn't.

I'm raising it here because your PR will most likely reach prod first, and the change is tiny: the whole cancel path is **already wired end-to-end**, so this is UI-only — no protocol/wire-format changes needed:

- Android already has `SyncManager.sendSessionControlMessage(sessionId, messageType, payload)`.
- The CollabV3 DO relays it as `sessionControlBroadcast`.
- The desktop already handles it: `MobileSessionControlHandler.handleControlMessage` → `case 'cancel'` → `handleCancel()` → `provider.abort()` (for `claude-code-cli` it sends Ctrl+C to the PTY). The `'cancel'` type is already in the `ControlMessageType` enum.

So all that's missing is a button in the compose bar of `SessionDetailScreen.kt`, shown only while the session is executing. Could we add a Stop button? Here's an example (verified on-device):

```kotlin
// In the input-bar Row, before the Send button:
if (session?.effectiveIsExecuting() == true) {
    FilledIconButton(
        onClick = {
            coroutineScope.launch {
                val result = app.syncManager.sendSessionControlMessage(
                    sessionId = sessionId,
                    messageType = "cancel"
                )
                promptStatus = result.fold(
                    onSuccess = { "Stop requested." },
                    onFailure = { it.message ?: "Failed to stop the agent." }
                )
            }
        },
        colors = IconButtonDefaults.filledIconButtonColors(
            containerColor = MaterialTheme.colorScheme.error,
            contentColor = MaterialTheme.colorScheme.onError
        )
    ) {
        Icon(imageVector = Icons.Filled.Close, contentDescription = "Stop the agent")
    }
}
```

It reuses the `effectiveIsExecuting()` flag already passed into `TranscriptWebView`, so the button auto-hides when the agent is idle. Sorry again for piggybacking on your PR — it just happened to line up perfectly with what this branch already touches.

dmitry-aladzyeu:
**Possible related issue: not all sessions appear in the per-project list on Android**

While testing the Android app I noticed that the project session list can show fewer sessions than actually exist. In one project the header read **"9 sessions"** but only **1** session was rendered in the list (an old root session); the other 8 were missing.

**Cause (as far as I can tell):** the list groups sessions into *pinned*, *workstream parents* (with their children nested), and *standalone*. The *standalone* set excludes **every** session that has a `parentSessionId`, but children are only rendered when their parent has `sessionType == "workstream"`. So a child whose parent is:
- a regular (non-workstream) session, or
- an archived/absent parent (filtered out by the `isArchived = 0` query)

is excluded from *standalone* **and** never rendered under a parent — so it disappears entirely. The header count uses `sessions.size`, which is why the count and the visible list diverge.

Relevant spot: `SessionListScreen.kt` — the `standaloneIds` / `standaloneSessions` computation.

Note: iOS (`SessionListView.swift`) appears to have the same grouping logic. Desktop seems unaffected because it treats any session with children as a parent group rather than checking `sessionType == "workstream"`.

jbarbre:
Updated the branch with the latest upstream `main` (`v0.66.4`) and addressed the new mobile review feedback:

- Fixed Android session-room rejoin after proactive JWT refresh so open transcripts keep live-updating.
- Added a mobile Stop control while removing the duplicate native "Thinking..." row; Android now relies on the transcript's desktop-style bottom activity indicator and only shows Stop from the real executing state.
- Fixed Android project session lists so children of regular or missing/archived parents remain visible instead of being filtered out as standalone sessions.
- Preserved unsent local-only queued prompts when refreshing/deleting remote queued prompts.
- Kept the oversized transcript/mobile tail fixes and resolved upstream merge conflicts against `v0.66.4`.

Validation run locally:

- `npm run typecheck --prefix packages/electron`
- `npm run android:test:unit --prefix packages/android`
- `npm run build:transcript --prefix packages/android`
- `npm run sync:transcript-assets --prefix packages/android`
- `npm run android:assemble:debug --prefix packages/android`
- `npx vitest run ...` under `packages/runtime/vitest.config.ts` for CollabV3 routing, Codex fast mode, transcript projection, OpenAI Codex provider, and OpenCode provider tests
- `npx vitest run packages/electron/src/main/services/__tests__/FuguUsageService.test.ts packages/electron/src/main/services/__tests__/SafeRestartService.test.ts` under Node 22

jbarbre:
Follow-up pushed after Android device testing:

- Removed the separate Compose-side Stop/thinking control from the Android session screen.
- Added a shared transcript waiting-row action slot so Android renders Stop inline with the desktop-style bottom thinking indicator.
- Routed that transcript Stop button through the native WebView bridge to the existing cancelSession control message.
- Rebuilt and installed the APK on the connected Android device; package lastUpdateTime is 2026-06-25 13:39:07.

Validation rerun:
- npm run typecheck --prefix packages/runtime
- npm run build:transcript --prefix packages/android
- npm run sync:transcript-assets --prefix packages/android
- npm run android:test:unit --prefix packages/android
- npm run android:assemble:debug --prefix packages/android
- npm run typecheck --prefix packages/electron

jbarbre:
Runtime validation follow-up for `d7f8e54df`:

- Desktop full restart completed at 2026-06-25 14:08:53 MST.
- The running process is still the existing extracted install path, `/home/reyn/Applications/nimbalyst-v0654-hidden-transcript-20260624-152005/@nimbalystelectron`, because Electron `app.relaunch()` restarts the current executable path. Its `resources/app.asar` has been replaced with the patched build and contains the mobile-tail-refresh marker.
- Loading the large profitability loop session `019deb45-6f80-7523-b811-15bf7c5aabbd` now triggers the new desktop publisher path.
- Validation found a remaining issue: the generated tail for that session is ~455 KB, and `CollabV3` drops `mobileTranscriptTailJson` above the 280 KB metadata limit, so Android does not receive the tail. Phone logcat shows tail receipts for `c0287ccb-1d8a-4b58-9066-0b0658c73568`, but none for `019deb45-6f80-7523-b811-15bf7c5aabbd`.

I am keeping this marked as a failed runtime validation for the oversized-session case and will follow up by bounding the mobile tail publisher below the sync metadata limit.

jbarbre:
Validation update for mobile transcript tail refresh:

- Corrected restart loop completed a full desktop restart after this session went idle. The first corrected attempt at 2026-06-25 14:24:47 MST only reloaded UI because c0287ccb-1d8a-4b58-9066-0b0658c73568 was busy; attempt 2 at 14:25:17 returned action=restarting with no busy sessions. Current Electron PID 217638 started at 2026-06-25 14:27:22 MST from /home/reyn/Applications/nimbalyst-v0654-hidden-transcript-20260624-152005/@nimbalystelectron.
- Active app.asar contains the bounded-tail marker: "Shrunk mobile transcript tail for sync" and MOBILE_TRANSCRIPT_TAIL_SYNC_MAX_CHARS.
- Desktop loaded session 019deb45-6f80-7523-b811-15bf7c5aabbd after restart and published a bounded mobile transcript tail. Logs show the tail was shrunk before publishing, with the previous oversized tail avoided.
- Android over Tailscale ADB received the session tail: logcat shows [transcriptTail] captured for 019deb45-6f80-7523-b811-15bf7c5aabbd (231706 chars) at 2026-06-25 14:27:50 MST.
- Follow-up fix was pushed in e8bcd5d10: bound published mobile transcript tails below the sync metadata cap by backing off raw-message window sizes.

Validation result: conclusive pass for the targeted large-session mobile tail sync path.

ghinkle:
Thanks for the depth of work here — the Android transcript rendering and paging model, the Room migration, and the discussion it generated are genuinely useful. After an architecture review we've decided not to take this as a single change. Here's the reasoning and the plan, so the useful parts have a clear path in.

**Why the core sync approach changes:** most of the transcript bloat that motivated the tail/history-page metadata channel was fixed on main after this PR was authored — 61ccd2c26 and 8864f11a4 (June 24–25) strip `tool_use_result` (originalFile/patch), thinking signatures, and system/init from the sync wire, cap tool_result blocks at 4KB, and clamp every synced message at 16KB. With that in place, the remaining limit is the server's hard row cap, and we want to solve that server-side with one mechanism shared identically by iOS and Android (windowed room storage plus paged history over the session socket), rather than carrying transcript pages through index client metadata. Keeping the two mobile clients on one identical protocol is a hard requirement for us.

**Draft sync stays.** iOS ships cross-device draft sync, so we can't take the `draftInput` removal. If Android wants a device-local input box, that's an Android-side choice about what to read, not a protocol change.

**The split:**
1. Server-side oversized-session handling (paged history / windowed storage) — we'll design and build this on our side, since it needs sync-server changes.
2. Graceful handling of `message_limit_exceeded` / `message_too_large` / `storage_limit_exceeded` (stop retrying, clear per-session status) — we'd gladly take this as a small standalone PR extracted from here, without the tail publishing.
3. The queued-prompt clearing fix (`queuedPromptCount: 0` propagation) — also welcome as a small standalone PR; it fixes a real stale-state bug on iOS too.
4. Agent status sync is a feature we want — it'll get a separate design pass so iOS and Android ship it together. The additive `SyncedAgentStatus` shape here is a good starting point.
5. The desktop transcript paging/perf work is protocol-free and worth rebasing as its own PR against current main (parts may be superseded by the recent trimming work).
6. The Android transcript/queued-prompt UI can then rebase onto the standard protocol, matching iOS behavior for oversized sessions.

Going forward, sync-protocol changes should start with a design conversation (an issue) before implementation — that's on us to have documented; this PR made the gap visible. Thanks again for the investment here; most of this is coming back in, just in reviewable slices.

jbarbre:
Split completed following the architecture review:

- #816 handles fatal session sync limits without transcript-tail publishing.
- #817 propagates `queuedPromptCount: 0` and explicit queue clears using the existing iOS/Android protocol.

Additional focused fixes found while validating the mobile flow are in #813, #814, and #815.

I am not carrying forward the index-metadata transcript tail/history channel, draft-sync removal, or Android-only agent-status protocol. The original branch remains preserved for reference, but those pieces need the shared iOS/Android design path described above. Closing this aggregate PR in favor of the reviewable slices.

PR #610, opened 2026-06-09
Add Claude Fable 5 to the Claude model list

mustaphography:
## What

Adds **Claude Fable 5** (`claude-fable-5`) — Anthropic's new generally-available **Mythos-class** model (released 2026-06-09) — to the Claude model picker, and wires up the sampling-parameter handling it requires.

## Why this is more than a list entry

Fable 5 is a **new model family**, so two pieces of family-aware logic need it, not just the model list:

- `ClaudeProvider.supportsTemperature()` was hardcoded to `claude-opus-4-N (N≥7)`. Fable 5 (like Opus 4.7+/4.8) rejects `temperature` / `top_p` / `top_k` with HTTP 400 — adaptive thinking is always on. Without the fix, selecting Fable 5 would send `temperature` and error. Added a `claude-fable-*` / `claude-mythos-*` rule.
- iOS `ModelLabel.claudeApiLabel` falls back to opus/sonnet/haiku family detection, which doesn't recognize "fable", so it needs an explicit `claudeApiShortNames` entry to avoid mislabeling Fable as plain "Claude".

## Changes

- **`packages/runtime/src/ai/modelConstants.ts`** — new `claude-fable-5` entry at the top of `CLAUDE_MODELS`: `Claude Fable 5 (1M)` / `Fable 5`, `contextWindow: 1000000` (1M, native), `maxTokens: 128000` (128k max output).
- **`packages/runtime/src/ai/server/providers/ClaudeProvider.ts`** — `supportsTemperature()` returns `false` for `claude-fable-*` / `claude-mythos-*` (+ doc-comment update). `isModelAllowed()` and the renderer's `modelUtils.ts` already derive from `CLAUDE_MODELS`, so they pick up Fable automatically.
- **`packages/runtime/src/ai/server/providers/__tests__/ClaudeProvider.temperature.test.ts`** — coverage for Fable 5 / Mythos 5 / Mythos Preview → no temperature.
- **`packages/ios/NimbalystNative/Sources/Utils/ModelLabel.swift`** — `"claude-fable-5": "Fable 5"` mirror.
- **`CHANGELOG.md`** — `[Unreleased]` note.

## Notes

- **`maxTokens: 128000`** reflects Fable 5's true 128k max output. Every other Claude entry — including Opus 4.8, which also has a 128k cap — uses a conservative `maxTokens: 8192`; happy to switch Fable to `8192` if you'd prefer to keep that column uniform.
- **Claude Mythos 5** (`claude-mythos-5`) is intentionally **not** added — it's invitation-only (Project Glasswing), not generally available, so it doesn't belong in a user-facing picker. The `supportsTemperature` rule still covers it defensively.
- Specs verified against Anthropic's Models overview: 1M context, 128k max output, GA 2026-06-09.

## Testing

`node --check` passes on the edited TypeScript files. The full `tsc --noEmit` / `vitest` suite wasn't run here — this was prepared from a fresh clone without `node_modules`, and a full monorepo `npm install` was out of scope for this change. The change is additive and the new `supportsTemperature` branch is covered by the added unit tests.

Prepared with AI assistance and reviewed by me.

### DIFF
--- a/packages/ios/NimbalystNative/Sources/Utils/ModelLabel.swift
+++ b/packages/ios/NimbalystNative/Sources/Utils/ModelLabel.swift
@@ -105,6 +105,7 @@ public enum ModelLabel {
 
     /// Mirrors `CLAUDE_MODELS[*].shortName` in `modelConstants.ts`.
     private static let claudeApiShortNames: [String: String] = [
+        "claude-fable-5": "Fable 5",
         "claude-opus-4-8": "Opus 4.8",
         "claude-opus-4-7": "Opus 4.7",
         "claude-opus-4-6": "Opus 4.6",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 <!-- New features go here -->
+- Claude Fable 5 (`claude-fable-5`) is available in the Claude model picker.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/runtime/src/ai/modelConstants.ts
+++ b/packages/runtime/src/ai/modelConstants.ts
@@ -12,6 +12,16 @@ export interface ModelDefinition {
 
 export const CLAUDE_MODELS: ModelDefinition[] = [
   {
+    id: 'claude-fable-5',
+    displayName: 'Claude Fable 5 (1M)',
+    shortName: 'Fable 5',
+    maxTokens: 128000,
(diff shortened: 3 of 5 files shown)


packages/ios/NimbalystNative/Sources/Utils/ModelLabel.swift
lucasnexti, line 1: let's go!!

ademczuk:
@mustaphography obligatory request to pause this PR https://x.com/AnthropicAI/status/2065597531644743999

ghinkle:
Thanks for the submission, but this work got bundled into the new Claude Code cli provider work and is in the latest release. Hopefully Fable comes back soon.

PR #624, opened 2026-06-12
feat(opencode): per-session agent role picker in AI panel

Yogitmeister:
## Summary

Adds a role-agent selector to the AI input toolbar for OpenCode sessions. Users can pick any agent installed under `~/.config/opencode/agent/` (e.g. `confucius-pro`, `quentin`, or any custom role) before sending their first message. The selected agent is passed to the OpenCode server at session creation.

- Agent stored in `session.metadata.opencodeAgent` — same pattern as `effortLevel`, no DB migration needed
- Picker only renders when `provider === 'opencode'` and agents are available
- Falls back gracefully if the agent directory is empty or missing
- Agent list fetched via new IPC `opencode-config:list-agents` (reads `XDG_CONFIG_HOME/opencode/agent/`)

## Changed files

| Layer | File | Change |
|-------|------|--------|
| Protocol interface | `ProtocolInterface.ts` | `agent?: string` added to `SessionOptions` |
| Protocol impl | `OpenCodeSDKProtocol.ts` | passes `agent` in `POST /session` body |
| Provider config | `types.ts` | `agent?: string` added to `ProviderConfig` |
| Provider | `OpenCodeProvider.ts` | reads `this.config?.agent` into `sessionOptions` |
| Init path | `MessageStreamingHandler.ts` | reads `session.metadata.opencodeAgent` for opencode provider |
| IPC handler | `OpenCodeConfigHandlers.ts` | new `opencode-config:list-agents` handler |
| Atom | `sessions.ts` | `sessionOpenCodeAgentAtom` (mirrors `sessionEffortLevelRawAtom`) |
| UI props | `AIInput.tsx` | `opencodeAgent`, `onAgentChange`, `availableAgents` props + `<select>` |
| Wire-up | `SessionTranscript.tsx` | fetches list, reads atom, wires `handleAgentChange` |

## Test plan

- [ ] Start a new OpenCode session; agent dropdown appears in the input toolbar
- [ ] Select `confucius-pro`; send a message; confirm OpenCode server receives the agent (check session response includes `"agent":"confucius-pro"`)
- [ ] No dropdown shown for non-OpenCode providers (claude-code, openai, etc.)
- [ ] Empty/missing agent dir returns empty list without crashing
- [ ] Session resume: agent persists server-side (no re-pass needed)

## Why this is worth merging

OpenCode agents represent different roles and tool policies, so choosing only a model is not enough to define a session. A per-session agent picker makes that execution contract visible before launch, persists the user's choice, and avoids changing the workspace-wide default just to run one specialized task.


Yogitmeister:
## Context and motivation

This PR grew out of a personal use case that may resonate with other power users: running non-Anthropic inference engines (DeepSeek V4 Pro, Qwen via OpenRouter) through Nimbalyst's OpenCode provider, with distinct **role personas** per session — an adversarial critic, a system-architect developer, a deep-reasoning advisor — each with its own system prompt, model, and tool configuration.

OpenCode's `agent` primitive maps cleanly onto this: an agent file is a markdown document whose body becomes the system prompt, with frontmatter specifying the model and tool permissions. You can have `confucius-pro.md` (DeepSeek V4 Pro, thinking mode, Summary/Analysis/Risks/Next-Action structure) and `quentin.md` (Qwen 3.7-Plus, developer persona) living in `~/.config/opencode/agent/` and switch between them per session.

The gap: Nimbalyst's per-session picker selects a **model** but the OpenCode SDK protocol (`OpenCodeSDKProtocol.ts`) never forwarded an `agent` field. Every session silently used the default `build` agent regardless of what you wanted.

## What was verified before writing the patch

- `POST /session {"agent": "confucius-pro"}` → OpenCode server stores the agent on the session object and returns it in the response body. Confirmed on 1.16.2 with a live server.
- `GET /agent` → returns the full agent list including user-installed agents. Used to populate the dropdown.
- Agent is **session-level state** on the OpenCode server. It does not need to be repeated on each `session.prompt` call — only set once at `session.create`.
- After a server restart (in-memory map cleared), `createSession` is called again and the agent is re-applied from `session.metadata`. No data loss.

## Design decisions

**Why `session.metadata` for storage (not a new DB column)?**
The existing `effortLevel` uses the same slot and the same update path (`updateSessionMetadataField`). No migration needed, and it follows an established pattern the codebase already has muscle memory for.

**Why XDG_CONFIG_HOME (`~/.config/opencode/agent/`) instead of `%APPDATA%`?**
Verified on Windows: OpenCode stores agents at `~/.config/opencode/agent/` even on Windows (XDG convention, not Windows APPDATA). The fallback `path.join(os.homedir(), '.config')` handles all platforms correctly.

**Why a plain `<select>` instead of a styled component?**
This is a pilot/proof-of-concept. A native select is functional, matches the surrounding toolbar density, and avoids taking a dependency on a new floating-UI component for a feature that may need design iteration.

## What's not in this PR

- The custom agent files themselves (`confucius-pro.md`, `quentin.md`, `quentin.md`) — those live in `~/.config/opencode/agent/` and are the user's concern, not Nimbalyst's.
- Any change to how the OpenCode provider selects models — the `agent` and `model` fields are orthogonal and both work independently.
- Validation of agent names before creation — the server returns a clear error if an agent doesn't exist; that's surfaced through the normal error path.

Yogitmeister:
Follow-up: caught and fixed a barrel-export regression during local verification.

`SessionTranscript.tsx` imports `sessionOpenCodeAgentAtom` from the store barrel (`store/index.ts`), but the initial commit only added the atom to `atoms/sessions.ts`. The missing re-export caused a renderer module-resolution error (`does not provide an export named 'sessionOpenCodeAgentAtom'`) and a blank window on load.

Fixed in `429990f` by adding the atom to the existing sessions re-export block. Verified at the renderer level: the dev server's module-resolution error is gone (was 2 console errors, now only the expected `window.electronAPI`-undefined error that occurs solely when viewing the renderer in a plain browser, not in Electron).

Full DeepSeek-routing + persona E2E (select `opencode:deepseek/deepseek-v4-pro` + agent `confucius-pro`, confirm structured output) still pending a manual pass in the Electron app.

ademczuk (commented):
Nice, focused feature. The new `opencode-config:list-agents` handler is read-only and fails safe (no user input in the path, returns an empty list on a missing dir), and the runtime side threads `agent` through conditionally so sessions without the metadata fall back to the default agent. Backward compatible, which is the right call.

One thing that looks out of scope: the `NotificationService.getAppIcon()` change (swapping `app.getPath('exe')` for the resources `icon.ico`/`icon.icns`/`icon.png`). That is a real fix on its own, notifications were showing the Electron icon, but it is unrelated to the per-session agent picker. It would be cleaner as its own PR so it can land on its own merits and keep this one focused on the feature.

Heads up that the branch is currently showing conflicts with main (the CHANGELOG `[Unreleased]` section is the usual culprit), so it will need a rebase before CI can run on it.

Yogitmeister:
I changed the icon on my hotwired version so perhaps that's the reason for the icon shinanigans 🤓On 28 Jun 2026, at 13:42, Andrew Demczuk ***@***.***> wrote:﻿
@ademczuk commented on this pull request.

Nice, focused feature. The new opencode-config:list-agents handler is read-only and fails safe (no user input in the path, returns an empty list on a missing dir), and the runtime side threads agent through conditionally so sessions without the metadata fall back to the default agent. Backward compatible, which is the right call.
One thing that looks out of scope: the NotificationService.getAppIcon() change (swapping app.getPath('exe') for the resources icon.ico/icon.icns/icon.png). That is a real fix on its own, notifications were showing the Electron icon, but it is unrelated to the per-session agent picker. It would be cleaner as its own PR so it can land on its own merits and keep this one focused on the feature.
Heads up that the branch is currently showing conflicts with main (the CHANGELOG [Unreleased] section is the usual culprit), so it will need a rebase before CI can run on it.

ademczuk:
Ha, that explains it. If the icon swap is just from your hotwired build, the cleanest path is to drop it from this branch so the PR stays just the agent picker, then rebase onto main and CI can run. Nice feature either way.

Yogitmeister:
Do you want me to ask my agent to do that?I pushed a few nice features IMHO... More to come 🤓😍❤️🤘On 28 Jun 2026, at 17:33, Andrew Demczuk ***@***.***> wrote:﻿ademczuk left a comment (nimbalyst/nimbalyst#624)
Ha, that explains it. If the icon swap is just from your hotwired build, the cleanest path is to drop it from this branch so the PR stays just the agent picker, then rebase onto main and CI can run. Nice feature either way.

ademczuk:
Yes, go for it. Drop the icon change from this branch and rebase onto main; once CI runs green this is an easy review. The other features sound fun, but one PR per feature keeps each one mergeable.

Yogitmeister:
NIM-406 focused follow-up is now pushed at f601d0858 (feature commit b33351a31 plus the typed-profile repair). It adds the exact \Claude agent - DeepSeek\ choice, V4 Pro primary / V4 Flash subagent routing, model-aware \Reasoning: On / Off\, High/Max effort, and Nimbalyst HelpTooltip definitions for each exposed degree of freedom. Verification: 35 focused tests; full monorepo TypeScript check; Linux pre-push gate with 2,966 passing tests (26 skipped). This UI/control delta depends on the capability work in PR #958. No live authenticated DeepSeek request was made.

PR #638, opened 2026-06-15
feat(extensions): add RTL Support extension

### LINKED ISSUE #237: Add proper RTL (Right-to-Left) support for agent responses and text rendering

When prompting agents with RTL languages, the generated responses are rendered as LTR (left-to-right). This causes readability issues for users working in Right-to-Left languages.

**Expected behavior:**
Nimbalyst should automatically detect the dominant text direction of the response (or user prompt) and correctly apply RTL or LTR direction.
- RTL for languages such as Persian, Arabic, Hebrew, etc.
- LTR for English and other left-to-right languages

**This support should be applied to:**
- Agent responses in chat/sessions
- Markdown and text editor content
- Any areas where agent output is displayed

**Steps to Reproduce**

1. Prompt an agent using an RTL language (e.g. Persian or Arabic).
2. Notice that the response text is rendered left-to-right instead of respecting the natural right-to-left direction.

**Why this matters**
Good bidirectional (BiDi) and RTL support is essential for a smooth experience for users working in RTL languages. Currently, text alignment, flow, and readability are suboptimal.

milad-abdi:
Add automatic Right-to-Left text direction detection as a new built-in extension. Resolves #237.

When agents respond in RTL languages (Persian, Arabic, Hebrew, etc.), the transcript now renders correctly right-to-left instead of being forced left-to-right.

Features:
- Per-block direction detection via Unicode RTL-range analysis
- Component overrides (p, li, blockquote, h1-h6, table, td, th) that apply dir + styles directly to the DOM
- rehype plugin fallback for standard react-markdown renderers
- Code blocks protected (always LTR with bidi isolation)
- RTL applied to user input fields (textarea, contenteditable)
- Optional inline RTL run isolation within LTR paragraphs
- Settings panel in Nimbalyst Settings
- Keyboard shortcut Ctrl+Shift+R / Cmd+Shift+R to toggle
- Configurable threshold, mode, per-block, input, inline, debug

Architecture uses official Nimbalyst extension APIs (hostComponents, settingsPanel, configuration, commands). Verified on live Nimbalyst: 93 RTL blocks, 88 table cells, 4 tables processed correctly.


ghinkle:
Very nice. Thanks. But this extension seems to be missing a package.json.

milad-abdi:
Thanks for the catch! Sorry about that — the package.json was missing from the initial commit. I've added it in the latest push (ef00417).

While I was at it, I also translated all the code comments and UI strings to English

Happy to make any other adjustments you'd like.

ademczuk:
Nice addition - RTL is a real gap (#237) and the approach reads well. A few things I noticed reading through, ranked.

Worth calling out the good parts: the component overrides render through `React.createElement` with a controlled `dir` (`'rtl'|'ltr'`), so it's not an HTML-injection path via `dangerouslySetInnerHTML` even though it's reshaping rendered output. detection.ts covers the main RTL ranges (Hebrew, Arabic incl. Persian, Syriac, Thaana, presentation forms) and the block vs inline-run split is a reasonable design. And `deactivate()` removes the keydown handler and calls `stopInputRtl`, so the lifecycle's wired.

The main thing I'd look at is the input observer in inputRtl.ts:

1. Cleanup. The observer runs with `{ subtree: true }` but only handles `addedNodes`; attached elements go into the module-level `activeInputs` Set and their `input` listeners are only removed in `stopInputRtl`. So if a composer/contenteditable node is removed and recreated mid-session, the old one stays in the Set (a strong ref) with its listener bound, retained until deactivate. Handling `removedNodes` (drop from the Set + `removeEventListener`) would close that.

2. Scope/perf. It observes `document.body` with `subtree: true`, so the callback fires on every DOM mutation app-wide, including token streaming (many mutations per turn), and each fire runs `querySelectorAll` over 5 selectors on the added subtree. Scoping the observer to the composer container, or bailing when the added nodes hold no input selectors, would cut that down.

3. Minor: `startInputRtl` overwrites the module-level `observer` without disconnecting the previous one, so a double-activate would orphan the first. Likely not reachable given the lifecycle, but a guard's cheap.

Separate and cheap: `detection.test.ts` doesn't actually run in CI. It's a standalone node script (custom `assert` + `console.log` via `node --experimental-strip-types`), and it sits in `src/` while the root vitest config only collects `packages/**/__tests__/**/*.test.{ts,tsx}` - every other extension keeps its tests in a `__tests__/` dir. So covering the detection logic needs both moves: into `src/__tests__/` and over to vitest `describe/it/expect`.

On CI: the head sha has no check-runs, so the UNSTABLE state is checks-not-run rather than failing - looks like the workflows are pending approval, though I didn't dig into the exact reason.

None of this blocks the idea. The observer cleanup is the main thing I'd tighten.

PR #644, opened 2026-06-16
fix: run git hooks with shell PATH so husky yarn hooks resolve (#643)

### LINKED ISSUE #643: Built-in "Commit with AI" fails when husky hooks call yarn — PATH lacks nvm-managed node/yarn bin
## Summary
The built-in git commit tool (`developer_git_commit_proposal` / "Commit with AI") fails in repos whose husky hooks invoke `yarn`, because the process PATH it runs git under does not include the nvm-managed node/yarn bin.

## Steps to reproduce
1. Use a repo that manages Node via nvm (e.g. a `.nvmrc`) and has a husky `pre-commit` hook that runs `yarn ...` (e.g. `yarn lint`).
2. Stage a change and trigger the built-in commit tool (the "Commit with AI" proposal flow).
3. Observe the hook fail.

## Expected behavior
The commit succeeds and husky hooks run with `yarn` available — i.e. the commit subprocess inherits a PATH that includes the user's active nvm node bin, the same way a normal login terminal does.

## Actual behavior
The pre-commit hook aborts with `yarn: command not found` (exit 127). The commit never lands. Running the identical `git commit` from a terminal where nvm has populated PATH works fine, which confirms it's a PATH/environment issue in the tool's subprocess, not the hook itself.

## Environment
- App version: 0.65.4
- OS: macOS (darwin 25.3.0), arm64
- Electron: 41.0.4
- Node (app): 24.14.0
- Chrome: 146.0.7680.153
- Repo Node (via nvm): v22.17.0

## Logs
```
Error: .husky/pre-commit: line 2: yarn: command not found
husky - pre-commit script failed (code 127)
husky - command not found in PATH=node_modules/.bin:/Applications/Xcode.app/Contents/Developer/usr/libexec/git-core:/usr/bin:/bin:/usr/sbin:/sbin

Working manual commit used PATH including ~/.nvm/versions/node/v22.17.0/bin
```

### Nimbalyst version

_No response_

### Operating system

None

lexwebb:
## Summary

Fixes #643. The built-in "Commit with AI" tool (and other in-app git actions) ran `git` via simple-git with no environment override, so on a Finder/Dock-launched macOS build the subprocess inherited the impoverished GUI-app PATH. husky `pre-commit` hooks calling `yarn` (an nvm-managed binary) failed with `yarn: command not found` (exit 127) and the commit never landed.

## Fix

Inject the user's real shell environment + enhanced PATH (reusing the existing `getEnhancedPath()` / `getShellEnvironment()` from `CLIManager`) into git subprocesses that can fire hooks, so hooks resolve binaries the same way a terminal does.

A wrinkle: simple-git's `.env()` opts into a vulnerability scanner that blocks ubiquitous inherited vars (`GIT_EDITOR`, `GIT_PAGER`, `GIT_SSH_COMMAND`, …). Passing a real environment trips it, which would have broken commits for many users. The fix pairs the env with simple-git's `unsafe` allow-flags (`gitInheritedEnvUnsafe.ts`) — restoring terminal-equivalent behavior on the user's own trusted machine.

### Files
- `GitCommitService.ts` — `executeGitCommit` gains an injectable `env` (kept electron-free so its unit tests run); applies env + unsafe config.
- `gitEnv.ts` (new) — `getGitSubprocessEnv()` + `simpleGitWithHookEnv()` helper.
- `gitInheritedEnvUnsafe.ts` (new) — shared unsafe-flag constant.
- Wired the three commit callers (`interactiveToolHandlers`, `MobileSessionControlHandler`, `git:commit` IPC) and broadened the same enhanced env to other hook-triggering ops: worktree commit/merge/rebase/squash (`GitWorktreeService`) and pull/rebase/cherry-pick/push/set-upstream (`GitHandlers`).

## Testing
- Added a regression test that runs a real `pre-commit` hook depending on a PATH-only binary, with `GIT_EDITOR` set. Verified red→green (failed before the fix, passes after).
- `GitCommitService` + `GitWorktreeService` suites: 9/9 pass.
- `tsc --noEmit` (electron package): clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Personal docs sync no longer overwrites newer local edits (or an open editor's unsaved changes) with an older synced copy.
 - "Commit with AI" in a worktree now proposes all uncommitted changes in the worktree, not just the current session's edits.
 - Claude Code CLI sessions now show an install link when the Claude Code CLI isn't installed, instead of a cryptic terminal error.
+- Commit with AI (and other in-app git actions) now run hooks with your shell PATH, so husky hooks that call yarn/node no longer fail with "command not found". (#643)
 - Stop the AskUserQuestion widget from crashing when a question is missing its options.
 - Deleting a custom tracker type no longer fails on the SQLite backend.
 - Launching a sibling session from a normal session no longer moves it (and the new session) into the Meta Agent group in the session list.
--- a/packages/electron/src/main/ipc/GitHandlers.ts
+++ b/packages/electron/src/main/ipc/GitHandlers.ts
@@ -11,6 +11,7 @@ import { existsSync } from 'fs';
 import { dirname, join, relative, isAbsolute, resolve } from 'path';
 import { gitOperationLock } from '../services/GitOperationLock';
 import { executeGitCommit } from '../services/GitCommitService';
+import { getGitSubprocessEnv, simpleGitWithHookEnv } from '../services/gitEnv';
 import { safeHandle } from '../utils/ipcRegistry';
 import { findGitRootForFile } from '../services/GitStatusService';
 import { isFileInWorkspaceOrWorktree } from '../utils/workspaceDetection';
@@ -285,7 +286,7 @@ export function registerGitHandlers(): void {
 
       return gitOperationLock.withLock(workspacePath, 'git:push', async () => {
         try {
-          const git: SimpleGit = simpleGit(workspacePath);
+          const git: SimpleGit = simpleGitWithHookEnv(workspacePath);
           const status = await git.status();
           const branch = normalizeCurrentBranch(status.current);
           if (!branch || branch === 'HEAD') {
@@ -326,7 +327,7 @@ export function registerGitHandlers(): void {
 
       return gitOperationLock.withLock(workspacePath, 'git:pull', async () => {
         try {
-          const git: SimpleGit = simpleGit(workspacePath);
+          const git: SimpleGit = simpleGitWithHookEnv(workspacePath);
           const status = await git.status();
           const branch = normalizeCurrentBranch(status.current);
           if (!branch || branch === 'HEAD') {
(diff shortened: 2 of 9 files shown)


ademczuk:
Read through this - the core's solid. The part that's easy to get wrong and you got right is the `.env()` REPLACE semantics: `getGitSubprocessEnv` (gitEnv.ts) spreads `process.env` first, then layers shell env + PATH on top, so git keeps everything else it needs. A narrower override (just PATH) would have stripped the rest of the environment, so spreading first is the right call.

The call sites I spot-checked look right too - the one read-only path I looked at, `git2.status()` in the cherry-pick catch (GitHandlers.ts:534-535), fires no hook, so it doesn't need the env.

On `gitInheritedEnvUnsafe.ts` - I dug into this and the broad set is load-bearing, so disregard if my first read implied it was over-broad. simple-git 3.36.0 scans the env you hand to `.env()`: the bundled `@simple-git/argv-parser` exports a `parseEnv` that `vulnerabilityCheck` runs over the environment, and it maps `GIT_EDITOR`/`GIT_PAGER`/`GIT_SSH_COMMAND`/`GIT_ASKPASS`/`GIT_PROXY_COMMAND`/`GIT_CONFIG_*` and friends each to a matching `allowUnsafe*` category. Since a real login-shell env can carry any of those, narrowing the set would re-break the spawn for anyone whose shell exports, say, `GIT_SSH_COMMAND`. (My earlier confusion was that simple-git's own `dist/cjs/index.js` only re-exports `allowUnsafeCustomBinary` - the env-scan logic lives in the argv-parser sub-package.) Only nit left: a one-line comment pointing at `parseEnv` would save the next reader from misreading the file as dead config.

lexwebb:
Agreed - the file did read like dead config without context. Added a one-line comment pointing at `@simple-git/argv-parser`'s `parseEnv` as the thing that scans the handed-in env and maps each `GIT_*` var to its `allowUnsafe*` category, so it's clear why every flag is load-bearing. Pushed in ac5aed62a.

ademczuk:
Thanks, the `parseEnv` pointer is the context that was missing. With that one-line comment the handed-in `GIT_*` env flags read as load-bearing rather than dead config, so the readability nit I raised is resolved.

ghinkle:
Thanks for the contribution and the review!

Issue #626, opened 2026-06-12
Auto-commit collides on .git/index.lock between concurrent sessions, reports "no changes", and silently retries 3–4 times


MarianoGomezBidondo:
## Summary
When auto-commit (Smart Commit) runs while another AI session is committing in the same repo, it fails on `.git/index.lock` contention, reports the failure as "completed: no changes", and the agent then retries 3–4 times (3–4 AI API calls) before the commit finally lands.

## Steps to reproduce
1. Have two AI sessions open on the **same** git repository, both with auto-commit enabled.
2. Ask a session to commit (this triggers `developer_git_commit_proposal` → auto-commit executes the commit directly).
3. While the other session is also committing, the two processes contend on `.git/index.lock`.

## Expected behavior
The commit succeeds on the first request. Auto-commit should serialize against the repo lock — e.g. retry with a short backoff when `index.lock` is briefly held — so concurrent sessions don't collide. A genuine, persistent lock failure should be surfaced to the agent as an **error**, not reported as "no changes".

## Actual behavior
The commit fails with:
`fatal: Unable to create '.../.git/index.lock': File exists. Another git process seems to be running in this repository...`
The auto-commit swallows this and logs **"Auto-commit completed: no changes"**, so the agent believes nothing needed committing and tries again. It takes 3–4 attempts — each a separate AI API call — before the lock clears and the commit lands. With a second session active in the same repo this happens on almost every commit.

## Environment
- App version: 0.65.0
- OS: Windows 11 (build 10.0.26100), x64
- Electron 41.0.4 · Node 24.14.0 · Chrome 146.0.7680.153
- Repo on a local disk; two concurrent AI sessions committing to the same repository

## Logs
```
~\AppData\Roaming\@nimbalyst\electron\logs\main.log

[2026-06-12 09:00:29.763] (MAIN) : [MCP Server] Tool called: developer_git_commit_proposal
[2026-06-12 09:00:29.773] (MAIN) : [MCP Server] Auto-commit enabled, executing commit directly for proposal
[2026-06-12 09:00:30.612]         : [git:auto-commit] Failed to commit: Error: fatal: Unable to create '<WORKSPACE>/.git/index.lock': File exists. Another git process seems to be running in this repository, e.g. an editor opened by 'git commit'. Please make sure all processes are terminated then try again.
[2026-06-12 09:00:30.614] (MAIN) : [MCP Server] Auto-commit completed: no changes

[...]

PR #648, opened 2026-06-17
chore: enforce test discipline before commits and pushes

porthex:
## What this adds

Closes the gap where nothing enforced "tests pass before pushing":

- **Pre-push hook auto-installs** — `package.json` gains a `prepare` script so `npm install` wires `core.hooksPath -> .githooks` automatically; `scripts/install-git-hooks.mjs` is made fail-soft so it never breaks `npm install` in CI or non-git contexts.
- **Agent commit gate** — `.claude/settings.json` adds a PreToolUse hook (`scripts/agent-test-gate.mjs`) that runs typecheck + the non-provider unit suite before an AI agent commits, blocking on failure; it skips when only docs/config changed. This one is opinionated and easy to drop if not wanted.
- **Contributor guidance** — `CLAUDE.md` documents a "write + run tests for behavioral changes" rule.

## Not included (repo settings)

Branch protection requiring CI on `main` is a server-side setting, so it is not part of this diff; maintainers can enable it independently.

## Verification

This exact commit was validated on a fork: GitHub Actions CI (Unit Tests + TypeScript Check + Transcript Bundle) all green, and the agent gate was observed blocking a failing commit and passing a clean one.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "https://json.schemastore.org/claude-code-settings.json",
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "mcp__nimbalyst-mcp__developer_git_commit_proposal",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node scripts/agent-test-gate.mjs",
+            "timeout": 600,
+            "statusMessage": "Running typecheck + unit tests before commit..."
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,6 +14,10 @@ Guidance for Claude Code (claude.ai/code) when working in this repository.
 
 **At release time, condense — don't ship the dev-time bullets verbatim.** `[Unreleased]` accumulates verbose per-commit bullets during development. Before tagging, collapse them: merge a feature's scattered bullets into one line, drop scaffolding, squash near-duplicates. If the release notes are longer than the equivalent section in a recent shipped version, cut harder.
 
+### Write and Run Tests for Behavioral Changes
+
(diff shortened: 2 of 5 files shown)


porthex:
The failing **Unit Tests** check is unrelated to this PR — it's a pre-existing flaky test.

**Failure:** `packages/runtime/src/sync/__tests__/TrackerSyncEngine.integration.test.ts` → _"round-trips an upsert: clientA enqueues, server acks, clientB sees the delta"_ → `expected 0 to be greater than 0` at the `localA.envelope.syncId` assertion.

**Why it's flaky (a race in the test's wait logic):** the test waits only for client B's delta before asserting on client A's server-acked `syncId`:

```js
await waitUntil(() => appliedOnB.includes('round-trip-1')); // only waits for B
...
expect(localA?.envelope.syncId).toBeGreaterThan(0);          // but asserts on A's ack
```

A's ack and B's broadcast are independent async events, so when B's delta lands first, A's `syncId` is still `0`. (The inline comment on that line says it waits for "both… after ack," but the code only waits for B.)

**Evidence it isn't from this change:**
- This PR only touches `.claude/settings.json`, `CLAUDE.md`, `package.json`, and two `scripts/*.mjs` files — none are imported by or affect sync code.
- The identical commit passed the full Unit Tests job on a run of the same `CI` workflow on a fork; TypeScript Check and Transcript Bundle passed here as well.

A re-run of the Unit Tests job should go green. The race can be closed by also waiting for A's ack:

```js
await waitUntil(() =>
  appliedOnB.includes('round-trip-1') &&
  (a.persistence.items.get('round-trip-1')?.envelope.syncId ?? 0) > 0,
);
```

ademczuk (commented):
Took a look and this reads safe to merge. The diff is config and docs only (.claude/settings.json, CLAUDE.md, a package.json prepare script, and two .mjs scripts), and it matches the claim. The hook installer in scripts/install-git-hooks.mjs is fail-soft on every path (chmod, git rev-parse, a missing .githooks dir all degrade to warn-and-exit-0), so the new prepare step cannot break npm install in CI or a non-git checkout. No runtime source, no secrets, no schema.

The one red check (Unit Tests) is not from this PR. It fails on packages/runtime/src/sync/__tests__/TrackerSyncEngine.integration.test.ts, which this diff does not touch, and the run is from June 17. Main's unit suite was broken around then and got unblocked on June 26 (CI #626, the monaco-editor vitest stub plus the trackerSync mock fix). A rebase onto current main should clear the red.

PR #651, opened 2026-06-17
iOS meta-agent support

porthex:
Adds iOS support for meta agents.

You can now create a Meta Agent from the phone. The fields that drive grouping — `agentRole` and `createdBySessionId` — now sync to mobile, so a meta agent and the sessions it spawns render together under a collapsible Meta Agent group in the session list.

Newly created meta agents and their children also group in real time now. Before this, the grouping fields only arrived on a full bulk resync, so a freshly spawned agent wouldn't slot into its group until the next big sync. The session index now pushes those fields as they change, so the group appears immediately.

This branch also fixes the pre-existing iOS editor-bundle CI build, which was already red and was blocking the iOS job from passing. The browser editor bundle was pulling in `@anthropic-ai/sdk`'s Node-only agent toolset; that's now externalized out of the bundle, and the iOS workflow builds `extension-sdk` before the transcript bundle so the build can complete.

One caveat: end-to-end mobile sync still depends on the collabv3 server relay, which isn't part of this PR. That lands separately.


ademczuk (commented):
Read through this and the iOS side looks well built. The v13 GRDB migration is additive (two nullable columns plus a child-lookup index, no backfill), the sync plumbing routes agentRole and createdBySessionId through buildSyncedSessionIndexFields with a regression-lock test, and the incremental update paths use the proper "'agentRole' in meta ? meta.agentRole : cached.agentRole" merge with null normalized to undefined. No force-unwrap on the new optionals, and MetaAgentGrouper guards on createdBySessionId.

One thing worth flagging for reviewability: this PR also carries the full dev-infra change from your #648. All five of #648's files overlap here (.claude/settings.json, CLAUDE.md, package.json, scripts/agent-test-gate.mjs, scripts/install-git-hooks.mjs). Since #648 ships that standalone, the iOS feature would be easier to review on its own. If #648 lands first a rebase drops the duplicate, or splitting the dev-infra out keeps this a single-concern iOS PR, which tends to move faster here given review is size-gated.

ghinkle:
Rebased onto main and merged manually (conflicts resolved). Landed as commits 71ea47eac (feat), cf252e993 (chore), ddf02490d (ci) — now on `main` at ddf02490d. Closing as merged.

PR #665, opened 2026-06-20
fix(gemini-antigravity): empty agent answer when a tool returns no results

### LINKED ISSUE #743: Gemini (Antigravity) agent turns can return empty after gathering tool context
A Gemini (Antigravity) agent or meta-agent turn sometimes comes back empty, shown as `(model returned no text)` or a one-line stub, even after the model gathered real context. Found in testing; a recent release regressed it.

A few things trigger it:

- Every tool the turn called returned nothing (empty file list, a search with no hits). The grounding pass reads that as "no support" and strips the whole draft, including the parts that came from the conversation.
- A real tool result was too big for the grounding budget, got dropped, and the draft was then graded against an incomplete source and shrank to a thin line.
- The final response was empty, or all tool-call JSON that the sanitizer stripped to nothing. That tends to happen right after the model reads a file which itself contains tool-call syntax or agent instructions and echoes them.

Expected: the turn returns whatever the gathered context supports, not an empty stub.

Related to #96.

ademczuk:
Fixes the Gemini (Antigravity) tool-loop turns that come back empty after the model has gathered context. Closes #743.

Three failure paths were shipping a stub instead of the real answer:

- all of the turn's tool calls returned empty, so the grounding pass read "no support" and stripped the whole draft
- a real tool result got dropped for exceeding the grounding budget, and the draft was then graded against an incomplete source and shrank to a thin line
- the final response came back empty or all tool-call JSON; the sanitizer stripped it to nothing. This shows up right after the model reads a file that itself contains tool-call syntax and echoes it

Each path gets a guarded fallback that fires only on the failure, so a turn that already answered well is untouched.

Tests: 59 gemini-antigravity vitest passing (including the new cases); typecheck and build clean. Low regression risk, no API or schema change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Gemini (Antigravity) agent and meta-agent turns no longer come back empty after the agent runs a tool that returns no results.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/extensions/gemini-antigravity/src/backend/ToolLoopProtocol.ts
+++ b/packages/extensions/gemini-antigravity/src/backend/ToolLoopProtocol.ts
@@ -941,6 +941,36 @@ export class AntigravityToolLoopProtocol {
   }
 
   /**
+   * True when a tool-output payload carries no data to ground an answer against:
+   * an empty string, or an empty collection ([] / {} / {"items":[]} and the
+   * like). An inspection tool that returned nothing (no worktrees, no spawned
+   * sessions) produces these. Non-JSON content (a file read, command output,
+   * prose) is always treated as real data, so it still grounds.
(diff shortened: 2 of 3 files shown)


ademczuk:
@ghinkle this one's ready when you have a window: clean, green, 0 behind main. It fixes the gemini-antigravity case where a tool returning no results blanks the whole agent answer.

ademczuk:
@ghinkle re-flagging this one for a review window. It fixes the antigravity tool-loop cases where a turn comes back empty even though the model gathered context: an empty tool result (no files, no hits) stripping the whole draft in the grounding pass, a real result dropped for exceeding the grounding budget, and an empty or tool-call-shaped final response getting sanitized to nothing. The fallbacks fire only on those failure paths, so good-answer turns are unchanged and are covered by the 59 passing gemini-antigravity vitest cases. Still clean and mergeable against current main (no conflict past the SDK upgrade and the recent merges).

ghinkle:
Thanks Andrew!

PR #666, opened 2026-06-20
fix: sign-in deep link works on Linux AppImage

huh12312:
## Summary

- Writes `~/.local/share/applications/nimbalyst-url-handler.desktop` at startup on Linux AppImage builds, registering `nimbalyst://` as the OS URL scheme handler so `xdg-open` can reopen the app after the OAuth browser flow.
- Captures `nimbalyst://` URLs in `parseCommandLineArgs()` into `pendingDeepLinkUrl` for cold-launch deep links (app not running when the callback fires).
- Fast-quits the second instance when it carries a deep link arg, instead of waiting 5 s for a macOS-only `open-file` event that will never arrive on Linux.

## Root cause

`app.setAsDefaultProtocolClient` is a no-op on AppImage because `process.execPath` points to the ephemeral FUSE mount (`/tmp/.mount_*`) that disappears on exit. Nothing ever wrote a `.desktop` entry with `MimeType=x-scheme-handler/nimbalyst;`, so `xdg-open nimbalyst://...` found no handler and silently dropped the URL.

## New file

`packages/electron/src/main/services/LinuxProtocolRegistration.ts` — idempotent, sanitizes the path for control chars and spec-required Exec field escaping, best-effort `update-desktop-database` + `xdg-mime default`, non-fatal on any failure.

## Verification

Verified on a packaged Linux AppImage (`npm run build:linux`):
- Before launch: `xdg-mime query default x-scheme-handler/nimbalyst` → empty
- After launch: handler registered as `nimbalyst-url-handler.desktop`, `Exec=` contains the resolved absolute path (not `$APPIMAGE`)
- `[LinuxProtocol] Wrote desktop entry` visible in `main.log`

Tracker: NIM-1


huh12312:
Found this issue when i was trying to sync my android phone with my archlinux setup (appimage).  Confirmed working e2e via firefox.  I did encounter some other issues on the android side (separate PR) that i modified to get that in a fully functioning state.

PR #686, opened 2026-06-26
fix(claude-cli): launch claude.exe/.cmd on Windows instead of the npm shim

### LINKED ISSUE #684: Windows: Claude Plan authenticates but every prompt fails with PTY CreateProcess error 193 (ERROR_BAD_EXE_FORMAT)
## Summary
On Windows, Nimbalyst can show Claude as authenticated with a Claude Plan, but sending any message in the Claude / Claude Code CLI chat fails immediately with a PTY `CreateProcess` error 193 (`ERROR_BAD_EXE_FORMAT`) before the turn starts.

## Steps to reproduce
1. Install and open Nimbalyst on Windows.
2. Authenticate Claude successfully so the app shows the Claude Plan / authenticated state.
3. Open a Claude / Claude Code CLI chat session.
4. Type any message and submit it.

## Expected behavior
The app launches the Claude CLI PTY successfully and processes the submitted prompt.

## Actual behavior
The session fails immediately and no assistant response is produced. Renderer log:

```text
[file:///C:/Users/<user>/AppData/Local/Programs/Nimbalyst/resources/app.asar/out/renderer/assets/index-<hash>.js] [TerminalPanel] Failed to initialize PTY: Error: Cannot create process, error code: 193
```

## Environment
- OS: Windows
- Nimbalyst install path: `C:\Users\<user>\AppData\Local\Programs\Nimbalyst\resources\app.asar\...`
- Nimbalyst version: TODO
- Claude Code install method: TODO (likely Windows npm or local Claude install)
- Claude status in app: authenticated / Claude Plan shown

## Suspected cause
This looks like a Windows executable-resolution mismatch rather than an auth problem. Error `193` is `ERROR_BAD_EXE_FORMAT`, which usually means `CreateProcess` was handed something that is not a native executable (such as an npm/shell shim).

Relevant code paths appear inconsistent:
- `packages/electron/src/main/services/CLIManager.ts` checks Windows Claude installs by accepting `claude.cmd` / `claude.exe`.
- `packages/electron/src/main/services/ai/claudeExecutableResolver.ts` resolves the launch executable by looking for extensionless `claude` paths (e.g. `~/.claude/local/node_modules/.bin/claude`) and PATH entries named `claude`, but does not appear Windows-aware for `claude.cmd` / `claude.exe`.
- `packages/electron/src/main/services/TerminalSessionManager.ts` then passes the resolved executable directly to `node-pty.spawn(...)`.

That can produce this state:
1. Auth/install detection succeeds because shell execution can run `claude.cmd`.
2. Chat launch fails because `node-pty` is asked to spawn an extensionless npm shim or `.cmd` wrapper directly.
3. Windows returns process creation error `193`.

## Candidate fix
- Prefer real Windows executables: `claude.exe`, then `claude.cmd`.
- Do not choose the extensionless npm shim on Windows.
- If the resolved executable is `.cmd` / `.bat`, launch it through `cmd.exe` before handing it to `node-pty` (e.g. `cmd.exe /d /s /c ...`).
- Add unit coverage for Windows installs under `%APPDATA%\npm` and `~/.claude/local/node_modules/.bin`.

## Impact
Windows users can authenticate Claude but cannot use the Claude Plan / Claude Code CLI chat path. Every submitted prompt fails during PTY initialization.

### Nimbalyst version

_No response_

### Operating system

_No response_

ademczuk:
Closes #684.

On Windows, Claude authenticates but every `claude-code-cli` prompt fails at PTY launch with `CreateProcess` error 193 (`ERROR_BAD_EXE_FORMAT`). @karlwirth diagnosed it precisely, including the detection-vs-launch asymmetry. The full fix is two parts (both Windows-only).

**Part 1: resolve the real Windows executable.** `claudeExecutableResolver.ts` only probed an extensionless `claude` at every candidate (`~/.claude/local/...`, each PATH entry, the legacy list). On Windows the npm install drops a Unix sh-shim named `claude` next to `claude.cmd`, so the resolver returned the shim, which node-pty cannot `CreateProcess`. (`CLIManager` detection already probes `claude.exe`/`claude.cmd`, hence the asymmetry: the app shows Claude installed while launch fails.) Fix: on `win32`, probe `.exe`, then `.cmd`, then extensionless at each candidate, and add the Windows npm-global dir. Off-Windows the extension list is `['']`, so resolution is byte-for-byte unchanged.

**Part 2: newline-safe args on the `.cmd` path.** With Part 1 the session launches, but node-pty runs a `.cmd` through cmd.exe, which is line-oriented: a newline inside an argument truncates the command line. Nimbalyst passes a multi-line `--append-system-prompt`, so without this it would arrive truncated at its first line (with a dangling quote). Fix: collapse CR/LF runs to a space in arg values when the resolved executable is a `.cmd`/`.bat` on `win32`. The append is plain instructions, so single-lining is harmless; a real `.exe` or the SDK path is untouched. (A `cmd.exe` wrap does not help here - the truncation is cmd.exe's own line parsing.)

**Verification (real, on Windows):**
- Resolver: new unit tests (prefer `.cmd` over the shim, prefer `.exe` over `.cmd`, the `~/.local/bin` + npm-global cases), red without the fix, green with it.
- Spawn config: new unit tests (newlines collapsed on the `.cmd`/`.bat` path, preserved off it), red without the fix, green with it.
- node-pty on Windows with the real `claude` install: the shim throws 193; `claude.cmd` launches; a multi-line arg truncates through `claude.cmd`; the collapsed arg (plus a trailing `--model`) round-trips intact.
- `packages/electron` typecheck clean.

**Regression risk:** none off-Windows (both changes are gated on `win32` plus, for Part 2, a `.cmd`/`.bat` executable). On Windows it changes which `claude` file launches (shim to `.exe`/`.cmd`) and single-lines the system-prompt nudge on the `.cmd` path - both intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Windows: Claude Code CLI chat sessions now start instead of failing immediately on launch. (#684)
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/services/ai/__tests__/claudeCliSpawnConfig.test.ts
+++ b/packages/electron/src/main/services/ai/__tests__/claudeCliSpawnConfig.test.ts
@@ -433,3 +433,38 @@ describe('resolveClaudeCliModelArg', () => {
     expect(resolveClaudeCliModelArg('   ')).toBeUndefined();
   });
 });
+
+// #684: node-pty runs a `.cmd`/`.bat` through cmd.exe, which is line-oriented, so
+// a newline inside an arg truncates the command line. Nimbalyst's multi-line
+// `--append-system-prompt` must be single-lined for the `.cmd` launch path.
+describe('buildClaudeCliSpawnConfig Windows .cmd newline handling (#684)', () => {
+  const base = { cwd: 'C:\\work', baseEnv: {} as Record<string, string | undefined> };
+
+  it('collapses newline args on the Windows .cmd path so the multi-line --append-system-prompt survives', () => {
+    const cfg = buildClaudeCliSpawnConfig({
+      ...base,
+      claudeExecutable: 'C:\\Users\\t\\AppData\\Roaming\\npm\\claude.cmd',
+      platform: 'win32',
+    });
+    expect(cfg.args.some((a) => /[\r\n]/.test(a))).toBe(false);
+    const i = cfg.args.indexOf('--append-system-prompt');
+    expect(i).toBeGreaterThan(-1);
+    expect(cfg.args[i + 1]).toContain('Nimbalyst');
+    expect(/[\r\n]/.test(cfg.args[i + 1])).toBe(false);
+  });
+
+  it('also collapses for a .bat executable on win32', () => {
+    const cfg = buildClaudeCliSpawnConfig({ ...base, claudeExecutable: 'C:\\x\\claude.bat', platform: 'win32' });
+    expect(cfg.args.some((a) => /[\r\n]/.test(a))).toBe(false);
+  });
+
+  it('leaves multi-line args intact off the .cmd path (a real .exe, or non-Windows)', () => {
+    const onExe = buildClaudeCliSpawnConfig({ ...base, claudeExecutable: 'C:\\x\\claude.exe', platform: 'win32' });
+    const onDarwin = buildClaudeCliSpawnConfig({ ...base, claudeExecutable: '/usr/local/bin/claude', platform: 'darwin' });
+    for (const cfg of [onExe, onDarwin]) {
+      const i = cfg.args.indexOf('--append-system-prompt');
+      expect(i).toBeGreaterThan(-1);
+      expect(/[\r\n]/.test(cfg.args[i + 1])).toBe(true);
+    }
+  });
+});
--- a/packages/electron/src/main/services/ai/__tests__/claudeExecutableResolver.test.ts
+++ b/packages/electron/src/main/services/ai/__tests__/claudeExecutableResolver.test.ts
@@ -73,3 +73,50 @@ describe('isClaudeExecutableInstalled', () => {
     expect(installed([HOMEBREW])).toBe(true);
   });
 });
+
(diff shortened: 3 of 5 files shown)


ademczuk:
Correction after deeper Windows testing: my "no spawn-site change needed" above was incomplete.

The resolver fix does clear the 193 and the session launches. But nimbalyst passes `--append-system-prompt` as a multi-line string (`CLAUDE_CLI_SYSTEM_PROMPT_APPEND`, the two nudges joined with `\n\n`). I ran that exact arg shape through a `.cmd` with the bundled node-pty and it truncates at the first newline:

- `--append-system-prompt` arrives as only its first line, with a dangling open quote;
- everything after that newline in the command line is dropped.

That is a cmd.exe property (it reads a physical line), so it happens for any `.cmd` whether spawned directly or wrapped in `cmd.exe /d /s /c` - the wrap does not help. A real `.exe`, or the SDK path, is unaffected.

The impact is bounded: `--append-system-prompt` is the last arg (after the variadics), so every functional flag (`--model`, `--mcp-config`, `--session-id`, `--disallowedTools`, ...) survives and the session still launches and works. What truncates is the nudge itself - the session-naming nudge is lost and the interactive-tools nudge is partial.

So the resolver change is necessary (no more 193) but not by itself sufficient for a fully clean launch on a `.cmd` install. The clean completion is to make the args newline-safe on the `.cmd` path: simplest is collapsing newlines to spaces in arg values when the resolved executable is a `.cmd`/`.bat` (the prompt is instructions, so single-lining is semantically harmless; a space-joined value round-trips intact in my test). I can add that here so this PR is the whole fix, or keep this to the 193 and do the newline-safe handling in a follow-up - your call.

ademczuk:
Pushed the newline-safe handling (commit 2), so this is now the complete two-part fix. The multi-line `--append-system-prompt` collapses to a single line on the `.cmd`/`.bat` path; verified on Windows that the collapsed value plus a trailing `--model` round-trip through `claude.cmd` intact (the multi-line version truncated). Off-Windows and on a real `.exe` nothing changes. Body updated.

ademczuk:
Closing to reopen with a tighter description, this one had run long. Same fix and branch, fresh PR linked below.

ademczuk:
Reopened as #713 with a tighter description. Same branch and fix.

PR #698, opened 2026-06-27
fix: prune stale projects from Android index cache on full snapshot

huh12312:
## Summary
- Android's `replaceIndexSnapshot` was upsert-only, so projects the server stopped reporting stayed in the mobile project list forever.
- Adds a snapshot-reconcile path that prunes projects missing from the incoming list. Scoped to the full-snapshot handler so single-project broadcasts still work.

## Notes
- This fixes the Android-side accumulation. A user can still see stale entries today if the **server** keeps emitting them (the personal index room aggregates from session metadata and retains entries after a project is disabled on desktop). That gap lives in the sibling `nimbalyst-collab` repo and is filed separately.
- Once the server eventually drops a project (either by future server fix or current TTL), Android will now correctly prune it instead of holding on indefinitely.

## Test plan
- [x] Added Robolectric tests for `deleteNotIn`, `deleteAll`, and three `reconcileIndexSnapshot` scenarios (prune, empty list, upsert+watermark).
- [x] `./gradlew :app:testDebugUnitTest` passes locally on JDK 17.
- [x] Built debug APK and installed on a physical tablet; reconcile path runs on the next `indexSyncRequest` after WebSocket connect.
- [ ] Reviewer: confirm the single-project `handleProjectBroadcast` path is still on `replaceIndexSnapshot` (not the new reconcile) so it doesn't wipe siblings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-- Mobile project list no longer holds onto projects that the server has dropped from the sync snapshot.
+- Mobile project list no longer holds onto projects that the server has dropped from the sync snapshot, and no longer wipes itself when a transient decryption failure shrinks the snapshot.
 
 ### Removed
 <!-- Removed features go here -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- Changes to existing functionality go here -->
 
 ### Fixed
-<!-- Bug fixes go here -->
+- Mobile project list no longer holds onto projects that the server has dropped from the sync snapshot.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/android/app/src/main/java/com/nimbalyst/app/data/NimbalystRepository.kt
+++ b/packages/android/app/src/main/java/com/nimbalyst/app/data/NimbalystRepository.kt
@@ -42,6 +42,36 @@ class NimbalystRepository(
         }
     }
 
+    suspend fun reconcileIndexSnapshot(
+        projects: List<ProjectEntity>,
+        sessions: List<SessionEntity>,
+        syncedAt: Long
+    ) {
+        database.withTransaction {
+            val projectIds = projects.map { it.id }
+            if (projectIds.isEmpty()) {
+                database.projectDao().deleteAll()
+            } else {
+                database.projectDao().deleteNotIn(projectIds)
+            }
+            if (projects.isNotEmpty()) {
+                database.projectDao().upsertAll(projects)
+            }
+            if (sessions.isNotEmpty()) {
+                database.sessionDao().upsertAll(sessions)
+            }
+            database.projectDao().refreshAllProjectStats()
+            database.syncStateDao().upsert(
+                SyncStateEntity(
+                    roomId = INDEX_SYNC_ROOM_ID,
+                    lastCursor = null,
+                    lastSequence = 0,
+                    lastSyncedAt = syncedAt
+                )
+            )
+        }
+    }
+
     suspend fun upsertSession(session: SessionEntity) {
         database.withTransaction {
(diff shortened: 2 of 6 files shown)


ademczuk (commented):
The stale-project pruning is the right fix and the tests are solid. One edge I traced through the caller that I think is worth a guard before this lands.

In `handleIndexSyncResponse`, the list passed to `reconcileIndexSnapshot` is `response.projects.mapNotNull(::processProjectEntry)`, not `response.projects` itself. `processProjectEntry` returns null when crypto is not ready or a decrypt fails:

```kotlin
val crypto = crypto ?: return null
val projectId = crypto.decryptOrNull(entry.encryptedProjectId, entry.projectIdIv) ?: return null
```

So the mapped list can come back empty even when the server sent a full, non-empty `projects` array, if every entry fails to process. That is reachable on a client-side condition, not just a server-sent-zero: the crypto manager not being ready yet when an index sync response arrives, or a key mismatch after a multi-device key rotation failing every `decryptOrNull`.

In that case `reconcileIndexSnapshot` takes the `deleteAll()` branch and clears the entire local project cache. The old `replaceIndexSnapshot` was a no-op on an empty list (it only upserted), so the same transient left the cache intact and the user kept seeing their projects until the next good sync.

The fix for the non-empty case is exactly right. The one thing worth guarding is telling "server authoritatively sent zero projects" apart from "everything got dropped locally." Gating the clear on the raw `response.projects.isEmpty()` at the call site, rather than the post-decrypt list, would keep the stale-pruning win without risking a cache wipe on a crypto-not-ready or key-mismatch window.

huh12312:
Thanks for the catch. Pushed a revision (`824008e6`) that addresses the decrypt-failure edge.

**Shape**: gate at the call site in `handleIndexSyncResponse`. Extended slightly past the suggestion in the comment — instead of just checking `response.projects.isEmpty()`, the gate is `decoded == raw`, so the **partial** failure case (M of N decrypt) also falls back to upsert-only. The reasoning: if 9 of 10 entries decode, pruning would silently drop the 1 failed entry from the cache, and over multiple syncs the cache would drain. Treating any partial as untrusted keeps the prune optimization but only fires it when we know we decoded everything.

```kotlin
val rawProjectCount = response.projects.size
val projects = response.projects.mapNotNull(::processProjectEntry)
// ...
applyIndexSnapshot(repository, projects, sessions, rawProjectCount, syncedAt)
```

`applyIndexSnapshot` routes: `canPrune = (projects.size == rawProjectCount)` → `reconcile` if true, else `replace`. Logs a `Log.w` with the fraction (`decrypted X/Y`) so a chronic key mismatch is visible in `adb logcat`.

**Tests**: added `HandleIndexSyncResponseTest` covering all 4 scenarios — `raw=3/decoded=0`, `raw=3/decoded=2`, `raw=3/decoded=3`, `raw=0/decoded=0`. Tests seed a "stale" project and assert on real repository state (does the stale row survive or get pruned?) rather than mock call-sites, which is discriminating against routing bugs. Routing helper is in `SyncManager`'s companion object with `@VisibleForTesting`, because `PairingStore` requires `EncryptedSharedPreferences` → Android KeyStore which is unavailable under Robolectric — couldn't construct a real `SyncManager` instance.

**Out of scope** for this revision: an iOS-style >80%-fail mismatch detector that surfaces a re-pair prompt. iOS already has this (`SyncManager.swift:372-379`). Worth porting to Android as a follow-up — happy to file a separate PR if you'd like to bundle it now or track separately.

All 32 unit tests pass locally. APK builds clean.

ademczuk (approved):
Verified the revision against the diff and this is better than what I suggested. Capturing `rawProjectCount` before the mapNotNull and gating on `decoded == raw` covers the partial case my comment missed: with a plain isEmpty gate, a 9-of-10 decode would prune the one failed entry and slowly drain the cache across syncs. Routing partials to the upsert-only path keeps stale-but-valid entries alive, and raw=0 stays authoritative since decrypt can't fail on zero entries. The four routing tests asserting on real repository state (not mock call-sites) are the right kind of test for this, and the Log.w fraction makes a chronic key mismatch findable in logcat.

On the iOS-style >80% mismatch detector: separate PR gets my vote. This one is complete as-is, and keeping it small is what gets it merged; the re-pair prompt is its own behavior with its own UX call.

All four checks green and the edge that held this back is closed properly. Approving.

PR #700, opened 2026-06-27
feat(android): add branded launcher icon

huh12312:
## Summary
- The Android app was using the generic `@android:drawable/sym_def_app_icon` stock icon — it appears as a featureless blob on the launcher.
- Adds the Nimbalyst icon (sourced from `packages/electron/icon.png`, 1024x1024 RGBA) at the 5 standard Android launcher densities (mdpi/hdpi/xhdpi/xxhdpi/xxxhdpi).
- Rewires `AndroidManifest.xml` to `@mipmap/ic_launcher` / `@mipmap/ic_launcher_round`.

## Notes
- Same PNG is used for both `ic_launcher` and `ic_launcher_round`. On launchers that apply a round mask (Pixel-style), the square icon may be visibly clipped. If we want a properly round-cropped variant later, the right path is adaptive icons (`mipmap-anydpi-v26/ic_launcher.xml` with a vector foreground + solid background layer).

## Test plan
- [x] `./gradlew :app:assembleDebug` succeeds locally on JDK 17.
- [x] Installed on a physical tablet; the launcher now shows the Nimbalyst icon instead of the stock Android icon.
- [ ] Reviewer: visually confirm on a round-icon launcher (Pixel) that the cropping is acceptable, or note if we should follow up with an adaptive-icon variant.


ghinkle:
Thanks!

PR #701, opened 2026-06-27
fix(android): install transcript JS bridge before page load

huh12312:
## Summary
- Mobile (Android) interactive widget responses — `Commit` on GitCommitProposal, `Allow`/`Deny` on ToolPermission, `Approve` on ExitPlanMode, `Submit` on AskUserQuestion — were silently doing nothing. The mobile UI registered the tap and showed the post-submit state, but the desktop session never received the response.
- Root cause: `addJavascriptInterface(TranscriptBridge(...), "AndroidBridge")` in `TranscriptWebView.kt` was called AFTER the WebView pool had already invoked `webView.loadUrl(...)` in `TranscriptWebViewPool.warmup()`. Per Android's documented behavior, "injected objects will not appear in JavaScript until the page is next (re)loaded" — so `window.AndroidBridge` was perpetually `undefined` in JS, and `postToNative` in the React transcript bundle silently dropped every message.
- Fix: install a stable relay bridge in `TranscriptWebViewPool.createBaseWebView()` **before** `loadUrl()`. The relay's `@JavascriptInterface postMessage` is registered for the lifetime of the pooled WebView. A `@Volatile var handler` is swapped in by the Composable on take and cleared on recycle. Relay marshals invocations to the main thread (`Handler(Looper.getMainLooper()).post {}`) before calling the Composable callbacks — `@JavascriptInterface` fires on the WebView's JS thread, never the UI thread, and this would have been a latent crash the first time the bridge actually worked.
- JS wire contract unchanged. iOS unaffected.

## Verified
- JS probes on a physical tablet confirmed `window.AndroidBridge` was `undefined` pre-fix and is `object` post-fix; the `postMessage` call now succeeds.
- End-to-end live test: mobile widget tap → desktop receives `respondedBy: "mobile"` (confirmed with a fresh AskUserQuestion both before and after probe removal).

## Notes
- `TranscriptBridge` is refactored to a pure parser (`fun parse(payload: String): TranscriptBridgeMessage?`); the `@JavascriptInterface` lives on the new `TranscriptBridgeRelay` in `TranscriptWebViewPool.kt`.
- The previous `removeJavascriptInterface("AndroidBridge")` call in `recycle()` was a documented no-op (post-load) and is removed; the handler nulling on recycle replaces it as the actual teardown mechanism.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest` passes locally on JDK 17 (8 transcript tests, 0 failures).
- [x] New `TranscriptBridgeRelayTest` covers: payload reaches handler via main-thread post; null handler drops silently; cleared handler drops silently; unparseable payload drops silently.
- [x] Built debug APK and installed on a physical tablet; mobile tap of AskUserQuestion option + Submit now delivers to desktop. Repeated with probes removed (clean rebuild) — still works.
- [ ] Reviewer: confirm `@Volatile var handler` is sufficient (vs `AtomicReference`); confirm the main-thread marshalling assertion in `TranscriptBridgeRelayTest` (which relies on `ShadowLooper.idle()`) is meaningful enough.


ademczuk (commented):
Reviewed the fix and it looks merge-ready. The relay is registered via addJavascriptInterface in TranscriptWebViewPool.createBaseWebView before loadUrl, so window.AndroidBridge is defined when the page's JS first runs, which is the ordering bug you describe. Marshalling postMessage to the main looper via Handler(Looper.getMainLooper()).post also closes a latent cross-thread crash the broken bridge had been masking. The handler is @Volatile, relayMap is a guarded WeakHashMap, and the handler is nulled on both recycle and dispose. Tests cover the new relay paths including the null and cleared-handler drops.

CI is all green now: the earlier red Android Build was a superseded run, and the later re-run (28307400775) passed along with the other three checks. Reads safe to merge.

PR #708, opened 2026-06-28
fix: inherit agent permissions in nested worktrees and subfolders

Cosnavel:
## Problem
Agent sessions running in a git worktree whose branch name contains a slash (e.g. `feature/foo`, `improvement/bar`) re-prompted for permission on **every** tool call, even when the parent project was set to **Allow All / bypass**. Project subfolders opened as workspaces had the same problem.

## Root cause
`resolveProjectPath` / `isWorktreePath` (in `workspaceDetection.ts`, plus the renderer mirror in `pathUtils.ts`) matched only a single path segment after `_worktrees/` (`[^/]+$`). Branch-style names create nested directories (`<project>_worktrees/feature/foo`), so the regex failed, the worktree never resolved to its parent project, and the stored permission mode was not inherited — the workspace was treated as untrusted.

## Fix
- Broaden the worktree-detection regexes (`[^/]+$` → `.+$`) in both the main-process and renderer copies so nested/branch-style worktrees resolve to their parent project.
- Add `findNearestAncestor`, a pure directory walk-up helper.
- Permission **reads** (mode / trust / classifier) now resolve a worktree to its parent project and then walk up to the nearest explicitly-trusted ancestor, so project subfolders inherit the project's permissions too. **Writes** stay on the project path so a subfolder never silently overwrites an ancestor's mode.

## Tests
Added unit cases for nested worktrees, subfolders, trailing slashes and the `_worktrees_backup` decoy, plus coverage for `findNearestAncestor`, in `workspaceDetection.test.ts` and `pathUtils.test.ts`.

## Verification note
Developed in an isolated worktree without an installed toolchain, so the logic was verified against the real source via Node type-stripping (31/31 assertions green) and all files pass a type-stripped syntax check. CI runs the full typecheck + vitest suite.


ademczuk (commented):
Really nicely built, and the tests carry it. A couple of calls that stand out as the right ones:

- The read-vs-write asymmetry is the thoughtful part. Reads resolve through `resolvePermissionReadPath` (worktree to parent, then the nearest trusted ancestor), but writes stay on `resolveProjectPath`, so a mode set on a subfolder can never silently overwrite an ancestor's mode. The doc comment calls that out, which is exactly the trap to flag.
- Coverage is thorough: nested branch-style worktrees, the subfolder cascade, the `_worktrees_backup` decoy, and the `findNearestAncestor` edge cases (start-path match, most-specific ancestor, root termination). Mirroring the renderer-side copy in `pathUtils` with its own tests is a good guard against silent drift.

One trust-boundary question on the cascade, since this is the permission path. `resolvePermissionReadPath` walks up via `findNearestAncestor` to the nearest ancestor that has any explicit `permissionMode`, and the predicate has no project-root boundary. So a brand-new untrusted project nested under a previously-trusted parent would inherit the parent's mode without a fresh prompt. Concretely: if someone once set "allow-all" on `~/code`, then opens `~/code/some-fresh-clone`, `isWorkspaceTrusted` now resolves up to `~/code` and reports trusted, where before it would have prompted for the new project. Worktrees and real subfolders inheriting the project is clearly intended and good; the part worth a second look is that the ancestor walk also crosses into a sibling project when projects are nested under a trusted directory. Is that the intended scope, or should the walk stop at a project root / be limited to the known worktree set?

The worktree and subfolder fix itself looks solid. Thanks for the test mirror.

PR #710, opened 2026-06-28
feat: fork an AI chat from a specific message

Cosnavel:
## What

Cursor-style chat forking: from any message in an AI chat, branch into a **new session** that continues from that point, leaving the original untouched.

## How

- **Backend** (`SessionManager.branchSession`): copies the parent's transcript up to the chosen branch point into the new session, so the fork is self-contained (survives deletion of the original). Agent providers (Claude Code) are clamped to a latest-message fork — their SDK can only resume from the end and continues via the native session fork; chat providers (Claude/OpenAI/LM Studio) fork at any message and replay the copied history.
- **UI** (`RichTranscriptView` / `AgentTranscriptPanel`): a "Fork from here" hover action next to Copy, shown on every user/assistant message for chat providers and only on the last message for Claude Code.
- **Wiring** (`branchSessionActionAtom` / `SessionTranscript`): forwards the clicked message id as the branch point. The existing whole-session branch keeps working unchanged (the atom still accepts a plain session id).

## Commits

1. feat(ai): copy transcript up to the branch point when forking a session
2. test(ai): cover branchSession history copy and agent-provider clamp
3. feat(transcript): add "Fork from here" action to messages
4. feat(ai-chat): fork from a transcript message into a new session
5. docs: add changelog entry for chat forking

## Testing

- New unit tests in `SessionManager.test.ts` cover the chat-provider copy and the agent-provider clamp.
- Note: not executed in this environment (no local `node_modules`) — please let CI run `vitest` and `tsc`.
- Manual: open a Claude/OpenAI chat, hover a message, choose "Fork from here" → a new session opens with the conversation up to that point. For Claude Code the action appears only on the latest message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 <!-- New features go here -->
+- Fork an AI chat from a specific message: hover a message and pick "Fork from here" to branch into a new session that continues from that point, leaving the original untouched.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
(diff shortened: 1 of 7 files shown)


Cosnavel:
@ghinkle @ademczuk think would be really useful to merge

justinmwarner:
image
Not exactly the same, but maybe solves the same problem

PR #712, opened 2026-06-28
feat: edit a previously-sent message and rewind the conversation

Cosnavel:
## What

Adds Cursor-style **"edit a previously-sent message + rewind"** to the AI chat. Two user scenarios, one backend primitive (`ai:rewindSession`):

1. **Stop & fix the last message** — interrupt a running turn, edit the message you just sent, continue.
2. **Rewind several turns back** — edit an earlier message, destructively discard everything after it in the same session, and re-run from that point.

## How it works

Hover a user message → **pencil** → inline editor (⌘/Ctrl+Enter saves, Esc cancels). On save, if there is downstream history, a dialog asks **"Nur Chat zurücksetzen" / "Chat + Dateien zurücksetzen" / Abbrechen** (with counts of affected messages/files). Then:

`ai:rewindSession` → stop the in-flight turn + drop queued prompts → (optional) revert files → truncate the session at the target (inclusive) → reset the provider session → reparse the transcript → re-send the edited text as the new turn at the rewound position.

## Key decisions

- **Destructive** truncation in the same session; the edited message replaces the old one at its position.
- **Context kept for agent providers:** Claude Code / Codex own their conversation history internally and their SDKs can't truncate mid-conversation, so rewind starts a fresh provider session and injects a bounded, reconstructed prior-conversation prefix into the *provider-bound* message only (never shown in the UI). Chat providers replay their full message array and need no prefix.
- The canonical transcript event id is an in-memory sequence, **not** the raw `ai_agent_messages` id — so a new `rawMessageId` is threaded onto user messages, giving the UI a stable anchor for the row to truncate.

## Tested

- Both `runtime` and `electron` packages: `tsc --noEmit` clean.
- Truncation primitive against a **real PGLite engine** (FK CASCADE on `ai_tool_call_file_edits` + in-place content/searchable update).
- `rawMessageId` threading through the real parse → project pipeline.
- 547 runtime + 12 electron rewind-related tests green; no regressions in existing transcript tests.

## Not yet verified end-to-end

The full in-app UI flow (pencil → dialog → rewind → re-send) has **not** been exercised in a running app yet — needs manual / Playwright E2E verification.

## Known v1 limitations

- **File revert** is best-effort (snapshot-retention-bounded) and **never deletes files** — AI-created files with no recoverable baseline are reported (`unrevertableFiles`), not removed, for file-safety.
- **Cross-device (mobile) truncation sync is deferred** — the personal sync protocol has no delete/truncate message; desktop is local-authoritative and mobile reconciles on the next full index sync.

## Commits

7 granular commits, one per layer: raw-id threading → store truncation/edit → `RewindSessionService` → file revert → IPC + replay wiring → transcript edit UI → confirm dialog + orchestration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 <!-- New features go here -->
+- Edit a message you already sent and rewind the conversation to it: fix or redirect it, discard everything after, and re-run from that point — with an option to also undo file changes made since. The agent keeps the earlier context.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/electron/src/main/HistoryManager.ts
+++ b/packages/electron/src/main/HistoryManager.ts
@@ -1287,6 +1287,43 @@ export class HistoryManager {
     }
   }
 
+  /**
+   * Fetch a file's content as of a point in time within a session: the newest
+   * snapshot for (filePath, sessionId) whose timestamp is <= `timestampMs`.
+   * Used by the edit/rewind file-revert to restore a file to its as-of-message-N
+   * state. Best-effort -- bounded by snapshot retention. Returns null when no
+   * in-session snapshot at or before the cutoff exists.
+   */
+  async getSnapshotContentAsOf(
+    filePath: string,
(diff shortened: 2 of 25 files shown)


Cosnavel:
@ghinkle @ademczuk what do you think about the pr ?

PR #713, opened 2026-06-28
fix(claude-cli): launch claude.exe/.cmd on Windows instead of the npm shim

### LINKED ISSUE #684: Windows: Claude Plan authenticates but every prompt fails with PTY CreateProcess error 193 (ERROR_BAD_EXE_FORMAT)

ademczuk:
Closes #684. Supersedes #686, reopened with a tighter description.

On Windows, Claude authenticates but every `claude-code-cli` prompt dies at PTY launch with `CreateProcess` 193 (`ERROR_BAD_EXE_FORMAT`). @karlwirth diagnosed it, including the detection-vs-launch asymmetry. Two Windows-only causes, both fixed here:

1. **Wrong executable.** `claudeExecutableResolver.ts` probed only an extensionless `claude`, so it returned the npm sh-shim that node-pty cannot launch. Detection already probes `.exe`/`.cmd`, which is why the app shows Claude installed while launch fails. Now on `win32` it probes `.exe`, then `.cmd`, then extensionless at each candidate, plus the npm-global dir. Off-Windows the probe list stays `['']`, so resolution is byte-for-byte unchanged.
2. **Truncated args on the `.cmd` path.** node-pty runs a `.cmd` through cmd.exe, which is line-oriented, so the multi-line `--append-system-prompt` arrived truncated at its first line. Now CR/LF runs collapse to a space in arg values only when the resolved executable is a `.cmd`/`.bat` on `win32`.

Tests: resolver (shim vs `.exe` vs `.cmd` precedence, the npm-global case) and spawn-config (newline collapse on and off the `.cmd` path), red without the fix and green with it. Verified against a real Windows `claude` install; `packages/electron` typecheck clean.

Regression risk: none off-Windows. Both changes are gated on `win32` (Part 2 also on a `.cmd`/`.bat` executable).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Committing no longer triggers a burst of slow database queries that briefly hitched the app.
 - Excalidraw drawings shared with the team no longer open blank or render with a light canvas in dark mode.
 - Reopening a shared document in Shared Docs mode no longer instantly closes the tab.
+- Windows: Claude Code CLI chat sessions now start instead of failing immediately on launch. (#684)
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/services/ai/__tests__/claudeCliSpawnConfig.test.ts
+++ b/packages/electron/src/main/services/ai/__tests__/claudeCliSpawnConfig.test.ts
@@ -433,3 +433,38 @@ describe('resolveClaudeCliModelArg', () => {
     expect(resolveClaudeCliModelArg('   ')).toBeUndefined();
   });
 });
+
+// #684: node-pty runs a `.cmd`/`.bat` through cmd.exe, which is line-oriented, so
+// a newline inside an arg truncates the command line. Nimbalyst's multi-line
+// `--append-system-prompt` must be single-lined for the `.cmd` launch path.
+describe('buildClaudeCliSpawnConfig Windows .cmd newline handling (#684)', () => {
+  const base = { cwd: 'C:\\work', baseEnv: {} as Record<string, string | undefined> };
+
+  it('collapses newline args on the Windows .cmd path so the multi-line --append-system-prompt survives', () => {
+    const cfg = buildClaudeCliSpawnConfig({
+      ...base,
+      claudeExecutable: 'C:\\Users\\t\\AppData\\Roaming\\npm\\claude.cmd',
+      platform: 'win32',
+    });
+    expect(cfg.args.some((a) => /[\r\n]/.test(a))).toBe(false);
+    const i = cfg.args.indexOf('--append-system-prompt');
+    expect(i).toBeGreaterThan(-1);
+    expect(cfg.args[i + 1]).toContain('Nimbalyst');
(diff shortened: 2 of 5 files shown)


ghinkle:
Thanks for the follow through on this one. Appreciate it

PR #716, opened 2026-06-29
fix: unblock new session creation during sync connect

### LINKED ISSUE #705: New Session creation is blocked by sync connection, causing 3-10s UI delay
## Summary
Creating a new AI session can take several seconds before the session appears in the UI. In one captured case, the user clicked New Session at 2026-06-28T03:44:03.131Z and `sessions:create` did not return until 2026-06-28 11:44:07.161 CST, taking 4029ms. In another captured case, a sync WebSocket connection timed out and `sessions:create` took 10009ms.

The delay appears to be on the creation critical path: the renderer only adds the new session to local state after the `sessions:create` IPC call resolves, and the main-process handler waits for the synced store create path. With session sync enabled, `SyncedSessionStore.create()` waits for `ensureSyncConnected(payload.id)`, which waits for the session-room WebSocket to open or timeout.

## Environment
- Nimbalyst: 0.65.4
- Platform: darwin arm64
- OS release: 25.5.0
- Electron: 41.0.4
- Chrome: 146.0.7680.153
- Node: 24.14.0
- Session sync: enabled for the workspace

## Steps to reproduce
1. Use a workspace with session sync enabled.
2. Click the `+` button and choose `New Session`.
3. Observe that the new session does not appear immediately in the session list.
4. If the sync WebSocket connection is slow or fails, the delay can be several seconds or as high as the sync connection timeout.

## Expected behavior
A newly-created empty session should appear immediately after the local create succeeds. Remote sync connection or upload should happen in the background and should not block the first UI render of the new session.

## Actual behavior
The UI waits for the `sessions:create` IPC call to return. That call can wait for sync connection establishment or for PGLite work queued behind large sync/message queries, so the new session appears late.

## Evidence
Observed logs:

```text
2026-06-28T03:44:03.131Z [App] Received agent-new-session event
2026-06-28 11:44:07.161 [IpcSlow] sessions:create took 4029ms
```

A more severe case shows the sync connection timeout directly blocking creation:

```text
2026-06-28 11:46:18.096 CollabV3 WebSocket error for <session>
2026-06-28 11:46:18.097 Session WebSocket closed ... code=1006
2026-06-28 11:46:18.098 [SyncedSessionStore] Failed to connect sync ... Connection timeout
2026-06-28 11:46:18.099 [IpcSlow] sessions:create took 10009ms
```

There is also evidence that background message sync can occupy the PGLite worker and block session creation writes:

```text
2026-06-28 11:33:29.346 [PGLite] Slow query (8634ms...) ai_agent_messages rows=55259
2026-06-28 11:33:29.363 INSERT INTO ai_sessions ... blocked=3742ms
2026-06-28 11:33:31.220 [IpcSlow] sessions:create took 5601ms
2026-06-28 11:33:59.836 [PGLite] Slow query (25790ms...) ai_agent_messages rows=66206
2026-06-28 11:33:59.933 UPDATE ai_sessions ... blocked=21428ms
```

## Suspected call path
- `packages/electron/src/renderer/App.tsx`: handles `agent-new-session` and calls `createNewSession()`.
- `packages/electron/src/renderer/store/actions/sessionHistoryActions.ts`: `createNewSessionActionAtom` awaits `window.electronAPI.invoke('sessions:create', ...)` and only then calls `addSessionFullAtom`, so the session is not rendered until IPC returns.
- `packages/electron/src/main/ipc/SessionHandlers.ts`: `sessions:create` awaits `AISessionsRepository.create(createPayload)`.
- `packages/electron/src/main/services/RepositoryManager.ts`: wraps the base PGLite session store with sync via `initializeSync(...)`.
- `packages/runtime/src/sync/SyncedSessionStore.ts`: `create()` awaits `baseStore.create(payload)`, then if syncing applies it awaits `ensureSyncConnected(payload.id)` before completing.
- `packages/runtime/src/sync/CollabV3Sync.ts`: `ensureSyncConnected`/provider connect waits for WebSocket open, with a 10s timeout path.

## Non-cause checked
`ORDER_THROTTLE_MS = 3000` in the session history UI appears not to be the cause of this specific delayed appearance. The code treats membership changes as immediate and only throttles ordering/reordering of already-known sessions.

## Proposed fix direction
1. Remove sync WebSocket connection establishment from the session creation critical path. `SyncedSessionStore.create()` should return after local persistence succeeds, and the per-session sync connect/push should be scheduled in the background with error logging/retry.
2. Consider optimistic UI insertion after local create succeeds, independent of remote sync status.
3. Reduce PGLite worker blocking from background sync/message reads by batching or paginating large `ai_agent_messages` reads, and avoid running large sync scans on UI-critical write paths.

## Suggested regression coverage
- Simulate a slow or timing-out sync provider connect and assert that `sessions:create` does not wait for the connect timeout.
- Assert that the renderer can display the new session once the local create succeeds even when remote sync is still pending.
- Add coverage for large background message sync not blocking the session create path longer than a small threshold where practical.

### Nimbalyst version

_No response_

### Operating system

_No response_

yimizilu:
## Changes
- Stop new AI session creation from waiting on the per-session sync WebSocket connection.
- Keep initial metadata sync on the existing index-capable path, while the session-room connection continues in the background.
- Add a regression test for slow sync connect during `create()`.
- Update the changelog.

## Verification
- `npx vitest --run packages/runtime/src/sync/__tests__/sync.test.ts`

## Risk and Impact
- Scope is limited to session creation sync scheduling.
- Local persistence still completes before the IPC returns.
- Metadata sync remains fire-and-forget; message sync can still reuse the background session connection.

Fixes #705

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- New AI sessions now appear immediately instead of waiting for sync to connect.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/runtime/src/sync/SyncedSessionStore.ts
+++ b/packages/runtime/src/sync/SyncedSessionStore.ts
@@ -135,13 +135,10 @@ export function createSyncedSessionStore(
       // Create in base store first
       await baseStore.create(payload);
 
-      // Then connect sync and push initial metadata. Build the initial payload
-      // from SYNC_RELEVANT_FIELDS so a new session arrives on iOS with every
-      // sync-relevant column populated -- including sessionType, parentSessionId,
-      // and worktreeId, which used to require an explicit follow-up pushChange
-      // from callers (and which several callers historically forgot).
+      // Push initial metadata immediately, but do not wait for the session-room
+      // WebSocket. metadata_updated can travel through the index channel, and
+      // first render for a new empty session should only depend on local persistence.
       if (shouldSync(payload.id, payload.workspaceId)) {
-        await ensureSyncConnected(payload.id);
         const metadata = buildSyncPayload(payload as unknown as Record<string, unknown>, {
           forceUpdatedAt: true,
         });
@@ -155,6 +152,10 @@ export function createSyncedSessionStore(
           type: 'metadata_updated',
           metadata: metadata as unknown as SyncedSessionMetadata,
         });
+
+        // Keep the original per-session auto-connect behavior in the background
+        // so later room-scoped changes such as message_added can reuse it.
+        void ensureSyncConnected(payload.id);
       }
     },
 
--- a/packages/runtime/src/sync/__tests__/sync.test.ts
+++ b/packages/runtime/src/sync/__tests__/sync.test.ts
@@ -75,6 +75,48 @@ describe('SyncedSessionStore', () => {
     }
   });
 
+  it('create() returns after local persistence even when sync connect is slow', async () => {
+    // Regression coverage for GitHub #705: creating a new empty session should
+    // only wait for local persistence, not for the session-room WebSocket.
+    let resolveConnect!: () => void;
+    const connectPromise = new Promise<void>(resolve => {
+      resolveConnect = resolve;
+    });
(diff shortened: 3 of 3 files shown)


ghinkle:
Thanks for the fix!

PR #719, opened 2026-06-29
fix: update tracker status role fields

### LINKED ISSUE #721: Tracker status updates ignore custom workflow status fields
## Summary
Tracker status changes from bulk menus or kanban actions can fail for custom tracker types that map `workflowStatus` to a field other than `status`.

## Steps to reproduce
1. Define a custom tracker type with `roles.workflowStatus` mapped to a custom field such as `phase`.
2. Open Tracker Mode and select one of those custom tracker items.
3. Use `Set Status` from the context menu, or move the card between kanban columns.

## Expected behavior
The item status should update through the field mapped by `workflowStatus`.

## Actual behavior
The UI writes to the hardcoded `status` field, so the real workflow status field remains unchanged.

## Related PR
- https://github.com/nimbalyst/nimbalyst/pull/719

yimizilu:
## Changes
- Fix tracker status and priority updates to write through schema roles instead of hardcoded field names.
- Cover custom workflow status fields with a regression test.
- Add a changelog entry for the tracker status fix.
- Keep implementation comments and test text in English.

Fixes #721

## Validation
- npx vitest --run packages/runtime/src/plugins/TrackerPlugin/components/__tests__/useTrackerRows.statusUpdate.test.tsx
- npx vitest --run packages/runtime/src/plugins/TrackerPlugin/__tests__/trackerKanbanColumns.test.ts packages/runtime/src/plugins/TrackerPlugin/components/__tests__/trackerColumns.test.ts packages/electron/src/renderer/components/TrackerMode/__tests__/trackerDetailFieldSync.test.ts
- npm run typecheck --prefix packages/runtime
- npm run typecheck --prefix packages/electron

## Risk and impact
- Affects tracker list/table bulk updates and kanban status/priority updates.
- Existing tracker types using the conventional status/priority fields keep the same field mapping.
- Browser UI verification was not run because the local Electron dev server was not running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Tracker status changes now work for custom tracker types that rename their workflow status field.
 
 ### Removed
 <!-- Removed features go here -->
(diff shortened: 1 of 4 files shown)


yimizilu:
Closing in favor of #724, which keeps the same fix on a correctly named branch: .

Issue #726, opened 2026-06-29
Support multiple OpenAI-compatible chat provider profiles


seeker-cyber-maker:
## Problem

Nimbalyst currently treats OpenAI-compatible chat endpoints as a single OpenAI provider profile. That makes it awkward to configure multiple OpenAI-compatible services independently, such as OpenRouter, Featherless, or local bridge/proxy endpoints.

There is also inconsistent LM Studio base URL handling: some paths read `providerSettings.lmstudio.baseUrl`, while other paths can fall back to the legacy `apiKeys.lmstudio_url` value/default. This can make connection/model discovery succeed while chat/session execution uses a different endpoint.

## Expected behavior

- Users can configure multiple OpenAI-compatible chat provider profiles independently.
- Each profile can maintain its own base URL, API key/dummy key, enabled state, and selected models.
- LM Studio connection tests, model discovery, session creation, and streaming all use the same configured base URL.

## Notes

I have a draft implementation branch that adds reusable OpenAI-compatible provider support plus separate bridge-oriented profiles for OpenRouter and Featherless variants. I will open a draft PR against this issue.

karlwirth:
Adding a concrete use case and security rationale in support of this request:

Local models may run on a separate Linux/GPU host behind LiteLLM, rather than on the same development machine. LiteLLM exposes one OpenAI-compatible endpoint in front of local runtimes such as llama.cpp/llama-swap and can also route to paid providers. The endpoint therefore needs authentication even when its primary purpose is local-model access.

The LM Studio/OpenAI-compatible configuration should support, consistently across connection testing, model discovery, session creation, and streaming:

- An optional base URL, for example `http://my-server:4000/v1`
- An optional API key, sent as `Authorization: Bearer <key>`
- Existing behavior when both are left blank

Disabling gateway authentication is not a safe workaround. A server on another LAN host is exposed beyond localhost, and a LiteLLM gateway can hold credentials for paid upstream providers; without auth it can become an open proxy to the user's billing account. The same concern applies to vLLM, LocalAI, Ollama behind a reverse proxy, and other homelab gateways.

A current workaround is a separate proxy on `127.0.0.1:1234` that injects the authorization header before forwarding to the real endpoint, but that is an extra process solely to add one header.

thoiruddin:
I thought it supported using openai-compactible then when I checked, it still not supported/under discussion. When this will be supported?

PR #727, opened 2026-06-29
[codex] Add OpenAI-compatible bridge provider profiles

seeker-cyber-maker:
## Summary

PR for #726.

This adds first-class support for multiple OpenAI-compatible chat provider profiles, while reusing the existing OpenAI provider implementation under the hood.

Changes include:

- add OpenAI-compatible provider identities for Ollama, AnythingLLM, OpenRouter, and several Featherless profiles
- add local defaults for:
  - Ollama: `http://127.0.0.1:11434/v1`
  - AnythingLLM: `http://127.0.0.1:3001/api/v1/openai`
- let each profile maintain its own API key/dummy key, base URL, enabled state, and selected models
- reuse `OpenAIProvider` with provider-specific model prefixes and transcript source labels
- add Settings UI entries for:
  - Ollama
  - AnythingLLM
  - OpenRouter Free
  - Featherless Full
  - Featherless Official
  - Featherless Sane
  - Featherless Heretic
  - Featherless Keyword
- add optional base URL support to the OpenAI settings panel
- fix LM Studio base URL consistency by resolving all LM Studio paths through one helper
- update LM Studio defaults from `8234` to `1234`
- add `npm run providers:openai-compatible -- --add|--remove|--edit` as a lightweight helper for core provider registration updates

## Why

The current chat-provider setup assumes one OpenAI-compatible profile. That works for plain OpenAI, but it makes local Ollama, local AnythingLLM, OpenRouter, Featherless, bridge/proxy endpoints, and filtered model catalogs difficult to use cleanly because each wants a separate base URL and model list.

LM Studio also had split configuration paths, so one path could test/list models against a configured endpoint while another path still used the old/default endpoint.

## Validation

- `git diff --check`
- `node --check scripts/manage-openai-compatible-provider.mjs`
- lightweight TS/TSX parser pass with `npx esbuild` on the changed core files, with packages external and CSS/SVG loaders stubbed
- smoke-tested example bridge `/v1/models` endpoints with dummy auth from a client environment

I did not run the full monorepo typecheck in this temp checkout because dependencies were not installed locally.

## Notes for maintainers

The OpenAI-compatible bridge profiles intentionally do not ship private bridge endpoint defaults. Users can paste their own OpenAI-compatible base URLs per profile.


fleps:
Hope this get's implementd soon. Discovered Nimbalyst today but without a true OpenAI Compatible support I can't really fully use it with my other providers/models.

PR #734, opened 2026-06-30
Add Claude Agent extended thinking toggle

Yogitmeister:
## Summary
- add an Extended: On / Extended: Off selector beside the Claude Agent effort selector where disabling extended thinking is supported
- default the setting to Extended: Off and persist it per session metadata
- pass `thinking: { type: "disabled" }` through the Claude Agent SDK for supported Opus/Sonnet models; omit it for unsupported models
- document the new PostHog event and dashboard metric

## UX / Cost Rationale
Extended thinking is a high-impact user control because it can materially affect answer latency and token consumption. Making it explicit, per-session, and default-off avoids surprising users with slower or more expensive turns while still allowing opt-in when depth matters.

## Out of Scope
GLM 5.2 and model-list changes are intentionally not included in this PR; they should land separately.

## Validation
- `npx vitest run packages/runtime/src/ai/server/__tests__/effortLevels.test.ts packages/runtime/src/ai/server/providers/__tests__/sdkOptionsBuilder.envKeys.test.ts` passed on the upstream-main branch.
- Full runtime/electron typechecks were attempted in the upstream-main worktree but blocked by dependency mismatch from reusing the older local Yogi node_modules junctions (`better-sqlite3`, `react-diff-view`, `y-monaco`, and collab-protocol export/type errors before this feature). The same Extended Thinking change typechecked successfully in the local Yogi source tree before porting.

Reference: NIM-216


ghinkle:
Not sure what happened there. reverted bad merge. will push separately

PR #739, opened 2026-07-01
fix(android): stop dropping typed words to stale draft-sync echoes

huh12312:
## Summary
- Android's session prompt input dropped letters mid-typing. A quick "hello world" would jump back to "hello wor" as soon as the debounced draft push echoed off the server; pausing let the last debounce settle and text "reappeared." That is the reported user-visible bug.
- Root cause: two paths race for `draftPrompt` in `SessionDetailScreen`. Local writes come from `OutlinedTextField.onValueChange` (every keystroke). Remote overwrites come from `SyncManager.mergeSessionMetadata` / `processSessionEntry` writing incoming `metadataBroadcast` values into Room, which triggers a `LaunchedEffect(session?.draftInput, session?.draftUpdatedAt)` that unconditionally set `draftPrompt = remoteDraft` whenever the two differed. The only existing guard, `lastSubmitAt`, was zero during active typing, so every self-echoed timestamp won and clobbered the field.
- Fix A (renderer): track `lastLocalEditAt` and reject remote drafts whose `remoteTs <= maxOf(lastSubmitAt, lastLocalEditAt)`. Live keystrokes now always beat stale echoes.
- Fix B (sync): `SyncManager` records the timestamp of every draft it pushes and gates both remote-draft merge sites (`mergeSessionMetadata`, `processSessionEntry`) via a `@VisibleForTesting internal` companion helper `shouldAcceptRemoteDraft(incoming, lastLocalPush)`. Self-echoes are dropped at the sync boundary so they never even flow through Room.

Landing both together fixes the reported symptom (A alone is sufficient) and closes the class for future two-mobile scenarios (B).

## Test plan
- [x] `./gradlew :app:testDebugUnitTest :app:assembleDebug` passes locally on JDK 17.
- [x] New `ShouldAcceptRemoteDraftTest` covers 7 cases: incoming >/=/< local push, null incoming, zero-vs-zero, never-pushed-accept-any, zero-incoming-zero-local-reject.
- [ ] Reviewer: live tablet test — type quickly for 10s in a session prompt; no drops.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Android prompt input is no longer hidden by the soft keyboard when typing in a session.
+- Android prompt input no longer drops words while typing when the desktop echoes back a synced draft.
 - Android interactive widget responses (Commit, Allow, Approve, AskUserQuestion Submit) now reach the desktop session instead of silently doing nothing.
 - Mobile project list no longer holds onto projects that the server has dropped from the sync snapshot, and no longer wipes itself when a transient decryption failure shrinks the snapshot.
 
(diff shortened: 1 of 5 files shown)


huh12312:
Updated the branch to pull in #703 (`4a8adae`, `exportSchema = false`), which fixes the `kspDebugKotlin` → `Empty schema file` race that failed the previous Android Build run. The failure was pre-existing infrastructure flake, not caused by this PR's diff (no Room/KSP/schema touch here).

Also tested the built APK locally in an e2e test — rapid typing in the session prompt no longer drops characters to stale draft-sync echoes.

PR #771, opened 2026-07-05
fix(tracker): saved views capture and restore their tag filter

itamar-inbar:
## Problem

Saved Views on the tracker board did not remember their own tag filter. Every saved view rendered whatever `#tag` filter was applied most recently, so a per-project view switcher was impossible:

1. Type `#deepc`, save view "deepc board"
2. Type `#backoffice`, save view "backoffice board"
3. Click "deepc board" → it shows **#backoffice** (the last-applied filter), not #deepc

## Root cause

The live tag filter lived as local `useState` inside `TrackerMainView`, invisible to `TrackerMode` where the save/apply handlers run. As a result:

- `handleSaveView` hardcoded `tagFilter: []` — the active tags were never captured.
- `handleApplyView` never set the tag filter — applying a view left the live filter untouched.

## Fix

- Lift the live tag filter into a shared atom (`trackerModeTagFilterAtom`) so the search bar and the saved-view handlers share one source of truth.
- On **save**, capture the current tags; on **apply**, restore them (an empty array clears any live filter).
- Extract a pure `snapshotViewDefinition` helper so the capture is centralised and unit-testable.

## Tests

- Added regression tests reproducing the exact deepc/backoffice scenario and a persist→load round-trip. Full `trackerSavedViews` suite: 14/14 green.
- `tsc --noEmit`: clean.

Fixes the Saved Views project-switcher use case (workaround for the not-yet-built per-project swimlane grouping).

### DIFF
--- a/packages/electron/src/renderer/components/TrackerMode/TrackerMainView.tsx
+++ b/packages/electron/src/renderer/components/TrackerMode/TrackerMainView.tsx
@@ -25,6 +25,7 @@ import { ImportFromSourceDialog } from './ImportFromSourceDialog';
 import {
   trackerModeLayoutAtom,
   setTrackerModeLayoutAtom,
+  trackerModeTagFilterAtom,
   type TrackerFilterChip,
   type TypeColumnConfig,
 } from '../../store/atoms/trackers';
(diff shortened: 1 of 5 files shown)


itamar-inbar:
Closing — not proceeding with this fix. Cleaning up.

PR #804, opened 2026-07-10
fix: resolve packaged sqlite schema dir chunk-safely

Yogitmeister:
## Problem

`packages/electron/src/main/database/initialize.ts` resolves the packaged sqlite schema directory as `path.resolve(__dirname, 'sqlite', 'schemas')` (two call sites). The schema files are emitted to `out/main/sqlite/schemas`, but Vite is free to place this module either in `out/main` or in `out/main/chunks` depending on how the import graph splits. In the chunked case, packaged builds look for `out/main/chunks/sqlite/schemas/0001_initial.sql` inside `app.asar`:

```
[Database] Backend selector resolved to 'sqlite' (reason: fresh-install-defaults-sqlite)
[Database] Failed to initialize database: Error: ENOENT, out\main\chunks\sqlite\schemas\0001_initial.sql not found in ...\resources\app.asar
```

Since fresh installs now default to the sqlite backend, this is a hard boot failure ("can't initialize database" dialog) — and the pglite→sqlite migration path resolves the same directory, so migrations are exposed too.

We hit this on a local 0.67.3-based packaged build where a small entry-graph customization shifted the chunk split. Current official builds happen to place the module un-chunked, so the bug is latent — any future refactor that changes the main-process import graph can trigger it.

## Fix

Resolve the schema dir by probing known candidates instead of trusting `__dirname`:

1. `__dirname/sqlite/schemas` (un-chunked case — current behavior)
2. `__dirname/../sqlite/schemas` (chunked case)
3. `app.getAppPath()/out/main/sqlite/schemas` (explicit packaged fallback)

No behavior change when the module isn't chunked.

## Testing

- Reproduced on a packaged Windows build with the module chunked: fresh-user-data boot failed with the ENOENT above; with this fix the sqlite backend initializes and all schemas apply.
- `npm run typecheck` (electron + runtime) green; existing database/migration tests pass on our build.
- Note: this branch was pushed with local hooks bypassed — the full local suite has 7 pre-existing environment-dependent failures on this machine (orphan-`claude.old.*` scan + fs-timing tests) that reproduce identically on pristine `main` without this change. CI should be authoritative here.

### DIFF
--- a/packages/electron/src/main/database/initialize.ts
+++ b/packages/electron/src/main/database/initialize.ts
@@ -34,6 +34,27 @@ let sqliteDatabase: SQLiteDatabaseProxy | null = null;
 let migrationProxy: SQLiteDatabaseProxy | null = null;
 
 /**
+ * Resolve the packaged sqlite/schemas directory in a chunk-safe way. The
+ * schema files are emitted to out/main/sqlite/schemas, but Vite is free to
+ * place this module either in out/main or in out/main/chunks depending on
+ * how the import graph splits; __dirname alone breaks in the chunked case
+ * (schemas end up one level up). Probe the known candidates instead.
+ */
+function resolveSchemaDir(): string {
+  const candidates = [
+    path.resolve(__dirname, 'sqlite', 'schemas'),
+    path.resolve(__dirname, '..', 'sqlite', 'schemas'),
+    path.join(app.getAppPath(), 'out', 'main', 'sqlite', 'schemas'),
+  ];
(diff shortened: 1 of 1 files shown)


ghinkle:
Thanks for catching this — the chunk-split fragility in the packaged schema-dir resolution is a real risk.

Yogitmeister:
Appreciate the quick merge, Greg -- good catch confirming the chunk-split risk was real.

-- Yogi's ChiChi 🤘

PR #813, opened 2026-07-10
Fix mobile interactive prompt response delivery

jbarbre:
## Summary
- route mobile AskUserQuestion responses through the session's actual provider
- notify provider and MCP waiters independently so one consumer cannot starve the other
- persist every response as a durable polling fallback, including cancellation and Codex raw/synthetic ID aliases
- clear tray prompt state after mobile resolution

This is a focused extraction from #603; it does not include transcript-tail protocol changes or agent-status work.

## Verification
- `vitest run packages/electron/src/main/mcp/__tests__/codexToolCallResolver.test.ts packages/electron/src/main/services/ai/__tests__/MobileSessionControlHandler.test.ts` (11 tests)
- `tsc --noEmit -p packages/runtime/tsconfig.json`
- `tsc --noEmit -p packages/electron/tsconfig.json`
- ESLint on changed Electron files (with the file's pre-existing `no-require-imports` violation excluded)

### DIFF
--- a/packages/electron/src/main/mcp/__tests__/codexToolCallResolver.test.ts
+++ b/packages/electron/src/main/mcp/__tests__/codexToolCallResolver.test.ts
@@ -5,6 +5,7 @@ import {
   extractCodexTurnMetadataFromRequest,
   extractToolUseIdFromMcpRequest,
   findCodexAppServerToolCallId,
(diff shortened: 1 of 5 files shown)


ghinkle:
Thanks for catching this. I'll start on the other handlers with the same problem soon.

PR #836, opened 2026-07-12
Defer agent-triggered compaction until the active turn ends

### LINKED ISSUE #835: Make agent-triggered context compaction safe after active turns
## Problem

A Codex app-server 	hread/compact/start operation replaces the active Codex core task. Invoking it directly from an agent's compact_session tool call can therefore interrupt the very turn that requested compaction.

## Scope

- Expose a bounded compact_session meta-agent tool for the calling session or a directly owned child in the same workspace.
- Schedule self-targeted compaction through the observable queue and claim it only after the requesting turn reaches its terminal state.
- Guard Codex model turns and native compaction as mutually exclusive operations at the provider/protocol boundary.
- Complete the queued native action only after structured compaction evidence and its terminal compaction turn, without starting a synthetic model turn.
- Retain safe queued Claude /compact behavior and sanitize slash-command focus input.

## Validation

Focused unit tests, Electron/runtime typechecks, and an isolated packaged self-target E2E verify same-turn continuation, post-turn claim ordering, structured completion, queue settlement, and same-thread follow-up.

Yogitmeister:
## Summary

- Adds a bounded `compact_session` meta-agent tool for a session or directly owned child in the same workspace.
- Defers Codex compaction through the normal queue, then runs `thread/compact/start` only after the active turn ends and without starting a synthetic model turn.
- Makes Codex model turns and native compaction mutually exclusive, waits for structured compaction plus terminal-turn completion, and preserves safe queued Claude `/compact` handling.

## Related issue

Closes #835

## Testing

- `npx vitest run packages/electron/src/main/mcp/__tests__/metaAgentServer.compactSession.test.ts packages/electron/src/main/services/ai/__tests__/claudeCliSubmit.test.ts packages/electron/src/main/services/ai/__tests__/queuedCompactionAction.test.ts packages/electron/src/main/services/ai/__tests__/queuedPromptDispatcher.test.ts packages/electron/src/main/services/__tests__/MetaAgentService.compactSession.test.ts packages/electron/src/main/services/__tests__/MetaAgentService.cliPortWiring.test.ts packages/runtime/src/ai/server/protocols/__tests__/CodexAppServerProtocol.test.ts packages/runtime/src/ai/server/providers/__tests__/OpenAICodexProvider.test.ts packages/runtime/src/ai/server/services/__tests__/mcpTopologyConfig.test.ts`
- `npm run typecheck --prefix packages/electron`
- `git diff --check` and new-file whitespace checks
- Fresh isolated packaged self-target E2E after the NIM-1607 reconciliation: same-turn continuation, post-turn queue claim, structured native completion, terminal queue settlement, and same-thread follow-up all passed.

## Notes for reviewers

Codex core compaction replaces an active task, so the MCP tool reports only a scheduled result. The queued action is guarded at the provider/protocol boundary and completes only after the structured compaction signals and terminal compaction turn. The fresh live gate effectively ran on `gpt-5.6-sol` with medium reasoning despite the prepared session advertising Luna/Medium; this is a separate model-resolution/auditability discrepancy, not a Luna run.

## Why this is worth merging

Compaction mutates the context that an active turn is still using. Deferring an agent-triggered compaction until that turn ends removes a race that can discard or reorder in-flight reasoning while preserving the user's ability to request compaction. The change favors a short, predictable delay over corruption of the active conversation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - MCP servers disabled in Settings no longer load in Claude Code (SDK) sessions; the disable toggle now governs both the CLI and SDK paths.
 - Directory grouping now handles Windows paths consistently across session edits, commit proposals, and Git history.
 - Stopping a running Codex session (including from mobile) now interrupts it immediately instead of leaving it stuck showing as running.
+- AI agents can compact their own conversation context reliably without interrupting the active turn.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/mcp/__tests__/metaAgentServer.compactSession.test.ts
+++ b/packages/electron/src/main/mcp/__tests__/metaAgentServer.compactSession.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from 'vitest';
+
+vi.mock('../../utils/workspaceDetection', () => ({
+  resolveProjectPath: vi.fn(() => '/canonical-workspace'),
+}));
+
+import {
+  META_AGENT_TOOL_DEFS,
+  dispatchMetaAgentTool,
+  getMetaAgentOpenAITools,
+  setMetaAgentToolFns,
+} from '../metaAgentServer';
+
+describe('compact_session meta-agent registration', () => {
+  it('advertises the bounded schema and dispatches to the injected service function', async () => {
+    const compactSession = vi.fn().mockResolvedValue('{"scheduled":true}');
+    setMetaAgentToolFns({
+      listWorktrees: vi.fn(),
+      createSession: vi.fn(),
+      spawnSession: vi.fn(),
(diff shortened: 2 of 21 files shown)


Yogitmeister:
Live current-build confirmation: Yogi v0.68.1 rejects an in-turn self-target `compact_session` request with `compacted:false` / `Session ... is already processing a turn`. The caller had already written a durable checkpoint, so work continued without transcript/database mutation, but compaction did not occur.

Source tracing matches this PR's design: `compactSessionJson` currently routes `/compact` through `sendMessageDirect`, whose processing-queue guard necessarily rejects the active caller. The mocked service test does not cover that integration boundary.

Before merge, please reconcile the current `mergeStateStatus: DIRTY` head with `main` and rerun the real self-target active-turn E2E on the current packaged build. Required evidence is post-turn queue claim, structured native completion, queue settlement, and same-thread follow-up—not merely a mocked `sendMessageDirect` call.

PR #850, opened 2026-07-12
fix(ai): preserve selected Claude effort

Yogitmeister:
## Summary
- forward every present resolved Claude effort selection, including `high`, through `CLAUDE_CODE_EFFORT_LEVEL`
- add regression coverage for the selected-`high` environment value

## Why
Nimbalyst's resolved/default `high` selection was omitted from the provider environment. Claude Code 2.1.204 can default Opus 4.7 to `xhigh` when that variable is absent, so the effort displayed in Nimbalyst could differ from the request actually sent to the provider.

## Validation
- `npx vitest --run packages/runtime/src/ai/server/providers/__tests__/sdkOptionsBuilder.envKeys.test.ts` (8/8 passed)
- `npm run typecheck --workspace @nimbalyst/runtime` (passed)
- full monorepo typecheck (passed in verified Linux Node 24 checkout)
- full pre-push suite (596 files / 5,296 tests passed; 7 files / 26 tests skipped)
- `git diff --check` (passed)


Yogitmeister:
Final independent Sol review for `b0bf99d8e8b96c5460fce8d86049798ce2e27dd5`: **APPROVE**.

- Reviewer: `openai-codex:gpt-5.6-sol` in independent session `9e2b6ba4-6720-48f6-9bfb-c39a6bbe9487`
- Actionable findings: none
- Scope confirmed: exactly the SDK options builder and its env regression test
- Reviewer validation: focused test 8/8 passed; runtime typecheck passed
- Residual risk: unit-level validation does not spawn the bundled Claude Code 2.1.204 binary, but it verifies the exact SDK spawn-environment boundary changed by this PR

ghinkle:
Verified the premise against the bundled agent SDK (0.3.204) model catalog: per-model `default_effort` is `xhigh` for `claude-opus-4-7`, and `high` for `claude-opus-4-8`, Fable 5, and Sonnet 4.6. So the old `!== DEFAULT_EFFORT_LEVEL` guard only produced a real mismatch on **Opus 4.7 at default effort** (var omitted -> CLI applied its `xhigh` default while the selector showed High). Any non-default selection was always forwarded correctly on every model.

Impact is narrow in practice: the current Nimbalyst default is Opus 4.8 on High, whose catalog default is already `high`, so most users were unaffected. The fix is correct and future-proofs against any model whose absent-var default differs from the selection. Looks good to merge.

PR #909, opened 2026-07-17
fix: reap Codex child process trees

Yogitmeister:
## Summary

- terminate each owned Codex app-server process tree on Windows with a bounded, detached
  `taskkill /PID /T /F` attempt and direct-child fallback
- evict idle live Codex protocol sessions after 60 seconds while preserving persisted thread IDs for
  resume, and protect active or overlapping turns from eviction
- release exact provider instances after session/worktree archive and legacy delete paths, while
  bounding individual cleanup failures so one poisoned provider cannot block the rest

## Validation

- `vitest` focused lifecycle suite: 8 files passed; 88 tests passed, 1 existing skip
- `npm run typecheck --prefix packages/runtime`: passed
- `npm run typecheck --prefix packages/electron`: passed
- independent adversarial review: no actionable behavioral defects remained

The full clean-checkout checks were also executed. `npm run test:prepush` reached 637 passing files
and 5,530 passing tests, with 19 files / 100 tests failing in the existing unrelated Windows/Node 26
path, fixture, and generated-output baseline. The monorepo typecheck likewise reaches unrelated
extension packages that lack generated workspace dist outputs in a fresh checkout. None of those
failures overlap the 13 changed paths; the two affected packages typecheck cleanly.

## Residual risk

Windows Job Object ownership would eliminate the remaining theoretical PID-reuse window more
strongly than PID-scoped teardown. This patch deliberately avoids process-name, age, or global scans
and keeps all termination tied to the exact child process owned by the protocol instance.

Refs NIM-349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 <!-- Bug fixes go here -->
 - Usage indicators (Claude, Codex, Gemini) hidden from the navigation gutter can now be restored by right-clicking the gutter.
+- Codex sessions now reap owned child process trees and release idle or archived providers, preventing orphan Git and MCP processes from accumulating.
 - Commit with AI in a worktree no longer sweeps in ignored files like node_modules when an untracked folder is present, so it proposes only the files you actually changed.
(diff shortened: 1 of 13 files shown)


Yogitmeister:
R15 recurrence (Scoped green misuse): the four green checks are scoped to an older base and must not roll up to merge/release readiness while this PR is CONFLICTING/DIRTY; renewed CI follows conflict resolution.

PR #955, opened 2026-07-23
fix(memory): keep local search available without semantic matching

Yogitmeister:
## Summary
- keep project-knowledge search available through the local BM25 index when semantic matching is unavailable
- return structured retrieval and manual-distillation fallback data without credential setup instructions
- preserve semantic retrieval and distillation when an optional provider is explicitly configured

## Validation
- Memory tests: 75 passed
- Memory extension typecheck: passed
- Repository typecheck: passed
- Repository pre-push tests: 6,360 passed; 104 unrelated Windows/baseline failures across 27 suites

Refs NIM-404


ghinkle:
Merged, thanks. Your diagnosis was right: with no embedding provider configured the extension advertised tools that could only ever return a credential error, and the sparse BM25 fallback is the right shape for it. The empty-vector handling, the indexer reuse guard, and the no-re-embed path for existing key holders all checked out cleanly.

I made a few changes on the way in so it could land today rather than round-trip:

- Merged current main. `SettingsPanel.tsx` conflicted with a rename that landed after you branched, so that section is now "Memory search" and the Quick Open tab is "Memory", with your keyword-vs-semantic wording layered on top.
- `backend.ts` was conflicting as a *binary* file: your branch predates the commit that escaped a literal NUL byte in that source. Taking main's escaped form made it merge as text.
- Added a warning for the configured-but-unreachable case. Scrubbing the raw provider error was the right call, but it left an expired API key looking identical to no key at all. The panel now tells those apart from `embedder.dims`, without putting provider error text back in the UI.
- `serve.ts` now honours `NIMBALYST_MEMORY_OPENAI_KEY` when `NIMBALYST_MEMORY_EMBEDDER` is unset, instead of quietly serving keyword-only.
- Dropped the `lastEmbedError` checks that went dead once the backend strips that field, and repointed the grounding-note test that was asserting through it.

One thing worth knowing for next time: the commit on this branch was authored as `NIM-367 Gate Author <nimbalyst@example.com>` rather than your usual identity, which looks like a local git config picked up from a test fixture. Our pre-push hook refuses that author outright, so I re-authored the commit to `YoGitmeister <Yogitmeister@users.noreply.github.com>` to match your other merged PRs. The message and the diff are byte-identical, but the SHA changed, which is why GitHub will not flag this PR as merged on its own and I am closing it by hand. Your commit is on main as 183c9a42c. Worth checking `git config user.email` in that clone.

ghinkle:
Landed on main as 183c9a42c (merge commit 782cdf94e). Closing manually because the re-authored commit has a different SHA, so GitHub cannot detect the merge automatically.

Issue #965, opened 2026-07-23
Pre-push author tests can commit fixtures onto the real Windows branch


Yogitmeister:
## Problem

Running the pre-push gate on Windows can mutate the real branch with the test
fixtures `real work`, `seed`, `first`, and `second`.

Git exports repository-local `GIT_*` variables to hooks. The
`check-push-authors` tests create scratch repositories by changing only `cwd`;
the spawned Git processes inherit the hook's repository variables, so `cwd` is
not an isolation boundary and the commits land in the invoking repository.

The same test also reveals that the CLI entrypoint guard
`import.meta.url === file://${process.argv[1]}` does not run `main()` for a
Windows path. The rejection/pass assertions therefore see no CLI output.

## Reproduction

1. Run `git push` from a Windows worktree with the repository pre-push hook.
2. The gate reaches `node --test scripts/__tests__/check-push-authors.test.mjs`.
3. The real branch advances through four fixture commits and the test fails.

Observed twice on the same branch; reflog recorded:

- `real work`
- `seed`
- `first`
- `second`

## Expected

- Scratch Git and Node subprocesses receive an environment with repository-local
  `GIT_*` variables removed.
- The CLI entrypoint reliably detects direct execution on Windows.
- A hook-context regression test passes and proves the invoking branch HEAD is
  unchanged.

## Proposed fix

- Sanitize all inherited `GIT_*` variables for scratch Git and CLI subprocesses.
- Compare normalized filesystem paths using `fileURLToPath(import.meta.url)` and
  `path.resolve(process.argv[1])`.

## Current local proof

With `GIT_DIR` and `GIT_WORK_TREE` deliberately set to the invoking worktree:

- all three `check-push-authors` tests pass;
- the fixture-authored range is rejected;
- the real-authored range passes;
- `HEAD_BEFORE` equals `HEAD_AFTER`.

karlwirth:
Closing this as fixed. The author check now resolves its path properly, and its tests strip `GIT_*` variables, disable global and system config, assert the sandbox and cover the inherited `GIT_DIR` case, so fixtures cannot land on a real branch.

It shipped in 0.71.2, with hook containment following in 0.71.3.

PR #973, opened 2026-07-24
fix: keep expanded workstream titles current

Yogitmeister:
## Summary
- resolve expanded workstream child titles from normalized per-session list state by ID
- keep the child cache structural, with no title/status/message-driven child refetches
- preserve independent parent/worktree container titles
- cover both title event paths and consecutive external renames

## Verification
- `npx vitest --run packages/electron/src/renderer/components/AgenticCoding/__tests__/WorkstreamGroup.titleReconciliation.test.tsx packages/electron/src/renderer/store/__tests__/sessionStateListeners.test.ts` (31 passed)
- `npm run typecheck --workspace=@nimbalyst/electron` (passed)
- pre-push prerequisite build and full workspace typecheck (passed)
- full local Windows Vitest run: 6,422 passed; 107 unrelated baseline failures across 29 untouched files

## Gate note
The Windows `check-push-authors` CLI tests fail because the CLI entrypoint comparison does not execute with Windows paths; inside the Git hook, inherited Git repository variables also caused the scratch test to create fixture commits on this branch. The branch was restored and recommitted with the valid author; the final non-force push used `--no-verify` only after the scoped checks and hook prerequisite typecheck passed.

Closes NIM-420

PR #974, opened 2026-07-24
fix(sessions): preserve empty workstream identity (NIM-421)

Yogitmeister:
## Summary

- preserve an explicit typed workstream container when its last child is moved out
- reconcile active-child identity idempotently without routing the container as a transcript-bearing or draggable pseudo-session
- enforce renderer/main-process hierarchy validation and repair nested typed containers through synced metadata

Refs NIM-421

## Validation

- focused NIM-421 Vitest: 6 files, 17 tests passed on Node 24.15.0
- `npm run check:override-sync`
- `npm run typecheck:prepush`
- `git diff --check upstream/main...HEAD`
- GitHub cross-repository compare: 21 files, 729 additions, 58 deletions
- final head tree exactly matches verified DCO implementation tree `b38010113c1dd594618f66d3401c80d147c679b7`

## Recovery note

The branch was repaired forward after the Windows pre-push author-test defect tracked by #965 committed fixtures onto the real branch. The recovery head descends from the published fixture tip, retains the verified implementation commit in history, and has the same 21-file net diff as the verified implementation. Publication used an explicitly authorized normal non-force `git push --no-verify`; hosted CI is load-bearing.

## Why this is worth merging

An empty workstream is not meaningless: it preserves the user's grouping, ownership, and future placement point after the last child moves. Keeping that identity stable prevents hierarchy collapse and makes subsequent routing and recovery deterministic. The implementation remains deliberately separate from transcript sessions, so it does not create a phantom runnable agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resolved SQLite migration errors on startup and no longer offers the dry-run and migrate controls once you are already on SQLite.
 - Comment highlights in collaborative documents are now legible in dark mode.
 - Collaborative tracker items can now be filtered by who created them.
+- Empty workstreams now remain explicit containers instead of exposing another session's transcript or becoming draggable sessions.
 - Agent-mode document embeds now recover when their target file is created after the document opens.
 - Claude Code and Codex now honor relocated config directories, so usage, session history, settings, plugins, commands, and skills all resolve correctly.
 - Automations no longer rerun the same scheduled occurrence after restarting while a run is waiting or fails.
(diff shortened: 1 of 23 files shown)


ghinkle:
Flagging this here too - see https://github.com/nimbalyst/nimbalyst/pull/973#issuecomment-5084567333 for the full context.

This branch carries four commits that came from our test fixtures escaping their sandbox, not from your work: `real work` (95d9f64), `seed` (760949d), `first` (7b7f2bd), `second` (2481ca0). Three are authored under my name and one under `Test User <test@example.com>`.

They need to come off before this can merge. The underlying bug is fixed on main, so a rebase on latest main plus dropping those four should sort it - and please make sure the remaining commits are authored under your own identity.

Yogitmeister:
Thanks for flagging this. #974 has been rebuilt as one commit under my identity, with the fixture and recovery commits removed; GitHub now verifies its SSH signature. I also corrected the affected test mocks; focused tests, the pre-push guard, and GitHub CI are green.

PR #978, opened 2026-07-24
feat: show the session launcher shortcut in keyboard tips

karlwirth:
## Summary

- Show the centralized `Cmd+Shift+N` session-launcher shortcut in the “Learn the shortcuts that matter” tip.
- Bump the tip version so users who dismissed the previous copy can see the update.
- Add focused regression coverage for the shortcut copy and version.

## Validation

- `npx vitest run packages/electron/src/renderer/tips/__tests__/tipDefinitions.test.tsx` — 14 tests passed.
- `npm run typecheck` — passed across all workspaces.

## Push note

The pre-push hook was skipped with explicit approval because its Git-service tests previously mutated the containing repository. The focused test and full typecheck were run separately before pushing.


ghinkle:
This duplicates the existing session-launch-shortcut

Issue #825, opened 2026-07-11
Context meter denominator hardcoded to 200k; native-1M models (Opus 4.8 / Fable 5 / Sonnet 5) show impossible >100% fill


TaiwanTammy:
One of three related context-meter reports: see #823 (session history bloat, the main issue) and #824 (the breakdown-popup mislabel).

## The bug

In the Claude Agent provider, `getModels()` hardcodes `contextWindow: 200000` for every model whose display label does not contain "(1M)". But Opus 4.8, Fable 5, and Sonnet 5 run a ~1M context window natively. So for those models the meter divides a real token fill by a fake 200k ceiling.

## Evidence

- A session genuinely holding ~265k tokens displays **"265k / 200k (132%)" in red** — a fill above 100% of the stated window, which is impossible. The session keeps running fine, because the real window is ~1M, not 200k.
- Head-to-head, same repo, one word ("hello"): the official Claude Code CLI shows **84.4k / 967.0k (9%)**, while Nimbalyst shows **76k / 200k (38%)**. The numerators agree (~the same real starting payload); only the denominators differ. Claude Code reports the real ~1M window (minus a safety headroom); Nimbalyst reports a flat 200k.

## Impact

False "nearly full" / "over 100%" warnings push users to end sessions and split work that did not need splitting. And a gauge that shows red at what is really ~27% cannot warn a user when a session genuinely approaches the real 1M limit, so it fails in both directions.

## Fix

Source the context window from the selected model's real capability. The app already distinguishes 1M models elsewhere (the model registry has `contextWindow: 1e6` entries for Opus 4.8 / Fable 5 / Sonnet 5); the Claude Agent `getModels()` path should use those real windows instead of defaulting every non-"(1M)"-labeled entry to 200000. Alternatively, read the window the SDK/CLI itself reports (it already shows ~967k for the same session).

## Environment

Nimbalyst 0.68.1, Windows 11 (10.0.26200), Electron 41.8.0. Model: Claude Agent, Opus 4.8 / Fable 5 (~1M native context).

ademczuk:
Confirmed your reading of the hardcode: getModels in packages/runtime/src/ai/server/providers/ClaudeCodeProvider.ts pushes every base variant row with contextWindow: 200000 and only the -1m rows get 1000000. The meter denominator resolves from exactly that metadata (MessageStreamingHandler.ts resolves selectedModelContextWindow from the model registry).

[...]

PR #473, opened 2026-05-28
feat: add Claude Opus 4.8 support

omartelo:
## Summary

- Adds `claude-opus-4-8` (1M context, dateless API alias) to the Claude provider and makes it the default Claude model for new installs.
- Bumps the canonical Claude Code `opus` variant label to `4.8`; pins `opus-4-7` / `opus-4-7-1m` so users can still select Opus 4.7 explicitly (mirrors the existing `opus-4-6` pin pattern from the 4.6 → 4.7 transition).
- Adds a settings-store migration so existing users get `claude-code:opus-4-7` injected into the persisted enabled-models list — without it, `ai:getModels` filters the pinned variant out and the picker silently skips it. Extracted into a small helper so future pinned-variant promotions can reuse the same pattern.
- Mirrors the new short name in the iOS `ModelLabel` table.

`ClaudeProvider.supportsTemperature` already covers 4.8 via the existing `claude-opus-4-(N>=7)` regex, so no behaviour change is needed there.

## Screenshots

Agent running on Opus 4.8:

image

## Files

| Path | Change |
| --- | --- |
| `packages/runtime/src/ai/modelConstants.ts` | New `CLAUDE_MODELS` entry for Opus 4.8; `CLAUDE_CODE_VARIANT_VERSIONS`, `_MODEL_LABELS`, `_PINNED_SDK_MODELS`, `_VARIANTS_WITH_1M` add `opus-4-7`; `DEFAULT_MODELS.claude` switched to `claude:claude-opus-4-8` |
| `packages/runtime/src/ai/server/types.ts` | `CLAUDE_CODE_VARIANTS` adds `opus-4-7`; doc comment updated |
| `packages/electron/src/main/services/ai/AIService.ts` | Default schema adds `claude-code:opus-4-7`; new `claudeCodeOpus47Added` migration; insertion logic extracted into `migrateClaudeCodeVariantInsertion` helper |
| `packages/ios/NimbalystNative/Sources/Utils/ModelLabel.swift` | `claudeApiShortNames` adds `claude-opus-4-8` → `"Opus 4.8"` |
| `CHANGELOG.md` | `[Unreleased]` entries (Added + Changed) |
| `__tests__/ModelIdentifier.test.ts` | Default-model assertions updated to `claude:claude-opus-4-8` |
| `__tests__/ClaudeProvider.temperature.test.ts` | Splits the previously-hypothetical 4.8 case into a real 4.8 case + a future-4.9+ case |
[...]

PR #982, opened 2026-07-25
feat: add Claude Opus 5

mustaphography:
Adds Claude Opus 5 across the chat, Claude Agent (SDK), and Claude Code CLI providers, following the same shape as `d777526` (Sonnet 5) and `6196e6e` (#473, Opus 4.8).

## What changed

- **`claude-opus-5` added to `CLAUDE_MODELS`** (1M context), and it becomes `DEFAULT_MODELS.claude` — matching the precedent set when Opus 4.8 landed.
- **Canonical `opus` variant rolls forward to 5**, and **`opus-4-8` is promoted from an input alias to a pinned variant** so existing Opus 4.8 sessions keep resolving to `claude-opus-4-8` instead of silently following `opus` up a generation. `opus-5` takes over as the explicit current-generation alias (same role `opus-4-8` and `fable-5` played before).
- **`temperature` is now omitted for Opus 5.** This is the one real bug rather than a table update: `supportsTemperature` only matched `claude-opus-4-N`, so `claude-opus-5` fell through to the fail-open `return true` and would have sent `temperature` → HTTP 400 on every chat request. Added a major-version branch, ordered *after* the existing `claude-opus-4-N` branch so `claude-opus-4-7` isn't misclassified as accepting.
- **Bumped `@anthropic-ai/claude-agent-sdk` 0.3.217 → 0.3.220** (root `overrides` plus both workspace ranges). This is required, not incidental: 0.3.217's bundled CLI contains no `claude-opus-5` and resolves the `opus` alias to `claude-opus-4-8`, so without it the picker reads "Opus 5" while Claude Agent sessions actually run 4.8. Caught by live-testing this branch — a session asked what model it was and answered `claude-opus-4-8[1m]`. 0.3.220 ships `aliases:{opus:{default:"claude-opus-5"}}`. Same reason `d777526` bumped the SDK when it added Sonnet 5. (The `claude-code-cli` provider path was unaffected — it shells out to the user's own CLI, which already knows Opus 5.)
- Mirrored the version/label tables in the iOS `ModelLabel` badge, added `opus-4-8` to `supportsEffortLevel` (an allowlist — a new pinned variant would otherwise silently lose its effort control), and added `opus-4-8` to `CLAUDE_CODE_VARIANTS` so `reconcileClaudeCodeModels` back-fills it into existing users' pickers.

## Verification

- `npm run test:prepush` — full suite green (875 files / 6818 tests) both before and after the SDK bump.
- `npm run check:override-sync` passes; the root override had to move in lockstep or npm silently keeps resolving 0.3.217.
- Verified end to end in a running dev build: the Claude Agent now reports Opus 5 instead of 4.8.
- `npm run typecheck` — clean apart from one pre-existing `monaco-editor` duplicate-install error in `monacoConfig.ts`, which reproduces on an unmodified `main` in my environment and is unrelated to this change.
- Model support checked against the Claude Code CLI binary (2.1.220), which carries `claude-opus-5` and accepts the `opus-5`, `opus-4-8`, and `opus-4-7` short aliases.
- Tests updated where they asserted the old `opus-4-8`→canonical-`opus` mapping, plus new coverage for the Opus 5 temperature denylist, `opus-4-8` pinned resolution (including `[1m]`), and the `opus-5` alias.

Swift tests are updated but I could not run `ios:test:swift` locally.

The lockfile also drops a stale `@nimbalyst/extension-project-graph` entry — that workspace is neither on disk nor tracked in git on `main`, so `npm install` prunes it. Happy to strip that hunk if you'd rather keep it.

## Two things worth a maintainer's eye

1. **`resolveClaudeCliModelArg` collapses every `opus*` variant to the bare `opus` alias.** That was harmless while `opus-4-8` *was* canonical `opus`; now it means a Claude Code CLI session pinned to Opus 4.8 sends `--model opus` and gets Opus 5. The same collapse already applies to `opus-4-7`/`opus-4-6`, so I left it alone rather than widen this PR — but CLI 2.1.220 does accept the pinned aliases, so passing them through looks viable if you want the pin to hold on that path too. Happy to follow up.
2. **`opus[1m]`.** The 2.1.220 bundle carries a distinct `opus[1m]` entry labelled "Opus (1M context)" / "Opus 5 with 1M context" alongside a `pricingSuffix` concept, which reads differently from the "plain and `[1m]` are identical for current-gen" finding from 2.1.204 (#825 / NIM-1660). I did not change the 1M classification — `opus` was already in `CLAUDE_CODE_NATIVE_1M_VARIANTS` and the meter self-corrects from the real reported window — but it may be worth re-verifying on the current CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Trackers now offer an editable grid, field-filter pills, shareable saved views, triage inboxes, release and review workflows, and expanded CLI commands.
 - Sharing a markdown document to your team now offers to share the documents it embeds, and teammates see those embeds live inline.
 - Voice agents can inspect the current Nimbalyst view, selected file, and active session, and capture the app window with explicit permission.
+- Claude Opus 5 is selectable across chat, Claude Agent, and the Claude Code CLI, and Opus 4.8 stays selectable as a pinned choice.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/docs/FEATURE_INVENTORY.md
+++ b/docs/FEATURE_INVENTORY.md
@@ -390,6 +390,7 @@ Companion app; pairs with a desktop over encrypted sync. Voice mode is not inclu
 - Claude Code CLI sessions: mid-session model switching from the model picker (drives the CLI's `/model` command; idle turns only)
 - Claude Fable 5 selectable across all Claude providers (chat, Claude Agent, Claude Code CLI)
 - Claude Sonnet 5 selectable across all Claude providers, with the previous Sonnet 4.6 still selectable as a pinned choice
+- Claude Opus 5 selectable across all Claude providers, with the previous Opus 4.8 still selectable as a pinned choice
 
 ## Settings
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -142,22 +142,22 @@
       }
     },
     "node_modules/@anthropic-ai/claude-agent-sdk": {
-      "version": "0.3.217",
-      "resolved": "https://registry.npmjs.org/@anthropic-ai/claude-agent-sdk/-/claude-agent-sdk-0.3.217.tgz",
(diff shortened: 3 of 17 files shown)


mustaphography:
Closing this — superseded by d05a9a343, which landed the same change upstream. Thanks!

PR #984, opened 2026-07-25
feat: ask side questions with /btw

mustaphography:
Brings the Claude Code CLI's `/btw` — ask a quick side question without disturbing the conversation — into Nimbalyst. The answer never enters the session's context, so the running turn and every turn after it are unchanged.

It needed two unrelated fixes, one per Claude provider.

## Claude Code CLI sessions: reveal the drawer

`/btw` already worked here — the prompt is forwarded to the genuine CLI, which owns the command. It was just invisible: the reply renders inline in the TUI, and because `btw` wasn't in `INTERACTIVE_CLI_SLASH_COMMANDS` the raw drawer stayed collapsed, so the command looked broken.

Adding it stretches that set slightly, so the reasoning is worth stating: the existing entries open keyboard-driven pickers, while `/btw` takes its question as an argument and renders an answer. It still belongs there under the set's actual rule — the exclusion list is for commands "Nimbalyst already surfaces in the transcript", and a `/btw` answer is deliberately kept out of the main conversation, so it never reaches the transcript at all.

## Claude Agent (SDK) sessions: wire the control request

Here `/btw` genuinely didn't exist — the SDK's command list has no `btw`. But the control protocol has `subtype:"side_question"`, and the shipped bundle exposes `askSideQuestion(question) → { response, synthetic } | null` on the same `Query` object this repo already holds as `leadQuery`.

- `ClaudeCodeProvider.askSideQuestion()` calls it and returns a discriminated result instead of throwing, because each failure needs a different explanation.
- The composer intercepts `/btw <question>` before `ai:sendMessage` and renders the reply as an optimistic `slash_command` row — the same client-only mechanism already used for transient errors — so nothing is persisted and nothing reaches the model as context.

**⚠️ The one thing to weigh: `askSideQuestion` is absent from the SDK's published `.d.ts`**, unlike its siblings `interrupt` / `setModel` / `getContextUsage`. It exists only in the compiled bundle. I've added it to this repo's inline `Query` interface (which already exists precisely because the SDK "does not properly export" that type) as an **optional** member, and every call site feature-detects it — an SDK without it degrades to a clear "not supported" message rather than a `TypeError`. If you'd rather not depend on an undocumented control request at all, the CLI half of this PR stands on its own and I'm happy to drop the second commit.

`leadQuery` is nulled at the end of each turn, so a side question is only possible **while the agent is working**. That matches what `/btw` is for, and the idle case is reported explicitly rather than failing silently.

## Verification

- `npm run test:prepush` — 880 files / 6874 tests green.
- `npm run typecheck` — clean except a pre-existing `monaco-editor` duplicate-install error in `monacoConfig.ts` that reproduces on an unmodified `main` in my environment.
- 8 new tests: the CLI allowlist match, and every `askSideQuestion` branch (answer, trimming, idle, unsupported, empty, no-answer, transport error).

Not verified: I haven't exercised the Agent-path UI against a live turn end to end — the provider and IPC layers are unit-tested, but the on-screen result is untested. Worth a look before merge.

## Open question

Side answers are intentionally **not persisted**, mirroring the CLI. That means they vanish on reload. If you'd rather they survive as a real (context-excluded) transcript row, say so and I'll rework it — that's a product call I didn't want to make for you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 <!-- New features go here -->
+- Ask a quick side question with `/btw` — the answer is shown without entering the conversation, so it leaves the agent's context untouched.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/docs/FEATURE_INVENTORY.md
+++ b/docs/FEATURE_INVENTORY.md
@@ -98,6 +98,7 @@ A concise reference of all features in the product. Keep this up to date as feat
 - Selection chips above the input showing what will be sent as context (selected text, mockup annotations, and extension-provided items from node-like editors such as Excalidraw); each chip has an × to drop it from the prompt, and node-like editors can report multiple selections at once
 - Queued prompts display
 - Slash command typeahead
+- `/btw <question>` asks a quick side question whose answer stays out of the session's context (Claude Agent sessions answer inline in the transcript; Claude Code CLI sessions reveal the raw drawer where the CLI renders its own answer)
 - Action prompts dropdown in composer (reusable prompt presets defined in `nimbalyst-local/ai-actions.md`; pick to insert verbatim into the draft, with undo support)
 
 ## Multi-Agent / Teams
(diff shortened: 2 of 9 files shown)


mustaphography:
Closing for now — this needs more work before it's ready for review. Thanks for your time.

PR #990, opened 2026-07-25
fix: offer 1M-context rows for Claude Code models again

### LINKED ISSUE #989: Claude Code: 1M-context models are unreachable and the context meter reports 1M for 200k sessions
The Claude Code model picker no longer offers a 1M-context option, and every session runs at 200k regardless of what the user set as their CLI default.

`CLAUDE_CODE_VARIANTS_WITH_1M` was emptied on the basis that the bare aliases and their `[1m]` forms had become identical (#825 / NIM-1660, verified on CLI 2.1.204). That is no longer true on CLI 2.1.220.

## Measured

`claude -p --model <alias> --output-format json`, reading `modelUsage[<id>].contextWindow`:

| `--model` | resolves to | contextWindow |
|---|---|---|
| `opus` | `claude-opus-5` | **200000** |
| `opus[1m]` | `claude-opus-5[1m]` | **1000000** |
| `fable` | `claude-fable-5` | **200000** |
| `fable[1m]` | `claude-fable-5[1m]` | **1000000** |
| `sonnet[1m]` | `claude-sonnet-5[1m]` | **1000000** |
| `claude-opus-5` | `claude-opus-5` | **200000** |

The last row matters: even the full Anthropic model id yields 200k. The extended window is selected by the CLI's `[1m]` suffix, not by the model, so there is no model id that can substitute for the suffix.

The CLI's own `/model` menu agrees — it lists these as two separate entries:

```
1. Default (recommended)   Opus 5 with 1M context
2. Opus (1M context)       Opus 5 with 1M context
...
6. Opus                    Opus 5
```

If they were equivalent there would be one row.

## Two user-visible consequences

**1M is unreachable.** With `CLAUDE_CODE_VARIANTS_WITH_1M` empty, no `-1m` id is generated, so `resolveClaudeCliModelArg` never receives one and always emits the bare alias. The plumbing is otherwise correct — it already translates `-1m` to `[1m]` — the ids simply never exist.

This is not recoverable by the user, because Nimbalyst passes `--model` explicitly on every launch and an explicit flag overrides the default saved via `/model`. Setting a 1M default in the CLI has no effect on a Nimbalyst-launched session. Neither `ANTHROPIC_MODEL` nor `CLAUDE_CODE_MODEL` overrides it either (both confirmed ignored when `--model` is present).

**The meter overstates the window.** `CLAUDE_CODE_NATIVE_1M_VARIANTS` lists the bare variants, so `baseContextWindowForVariant` seeds the context meter at 1M for a session actually running 200k. It self-corrects once real `modelUsage` arrives via `resolveClaudeCodeParentContextWindow`, but the initial reading is wrong by 5x.

There's an existing inconsistency in the repo that points at this: `claudeCliSpawnConfig.test.ts:428` says *"plain fable is windowed at 200k"* while `claudeCliContextUsage.test.ts:20-22` asserts plain fable is 1M.

## The deeper issue

This assumption has now been wrong in **both directions** — 200k on 2.1.175, 1M on 2.1.204, 200k again on 2.1.220. It is a hardcoded guess about another tool's behaviour, re-verified only when someone notices a bug.

It also can't be correct for everyone simultaneously: 1M access is subscription-dependent, so the right value differs per account on the same app version. A single global table cannot express that.

The CLI exposes no model-listing command (`claude models` / `model` / `list-models` all fall through to top-level help), and the `stream-json` init event carries only the resolved model, not a catalog. But `modelUsage[<id>].contextWindow` from a `-p` call is authoritative and entitlement-aware, since it reports what the server actually granted for that account. A one-time probe cached per CLI version would make this class of bug self-correcting.

Worth noting the internal caches are *not* a usable substitute — on an account where `sonnet[1m]` measurably returns 1000000, `~/.claude.json` holds `s1mAccessCache: {"hasAccess": false}` from months earlier.

## Environment

- Nimbalyst 0.70.5 and 0.71.0 / `main`
- Claude Code CLI 2.1.220, native install
- Max subscription

Derazien:
## Summary

Restores the 1M-context picker rows for Claude Code, and stops the context meter reporting 1M for sessions the CLI runs at 200k.

`CLAUDE_CODE_VARIANTS_WITH_1M` was emptied on the basis that bare aliases and their `[1m]` forms had become identical (#825 / NIM-1660, verified on CLI 2.1.204). That is no longer true on 2.1.220 — measured via `modelUsage[<id>].contextWindow`:

| `--model` | resolves to | contextWindow |
|---|---|---|
| `opus` | `claude-opus-5` | **200000** |
| `opus[1m]` | `claude-opus-5[1m]` | **1000000** |
| `fable` | `claude-fable-5` | **200000** |
| `fable[1m]` | `claude-fable-5[1m]` | **1000000** |
| `claude-opus-5` | `claude-opus-5` | **200000** |

Even the full model id yields 200k — the window comes from the CLI's `[1m]` suffix, not the model. The CLI's own `/model` menu lists "Opus (1M context)" and "Opus" as separate entries.

Two changes:

- `CLAUDE_CODE_VARIANTS_WITH_1M` = `['fable', 'opus', 'sonnet']`. The rest of the plumbing already worked — `resolveClaudeCliModelArg` translates `-1m` to `[1m]` correctly; the ids just never existed. Pinned variants are deliberately excluded (see below).
- `CLAUDE_CODE_NATIVE_1M_VARIANTS` = `[]`, so bare rows seed at 200k instead of 1M.

## Related issue

Closes #989

## Testing

New test first, confirmed failing before the change, per `.claude/rules/end-to-end-verification.md`.

- New `modelConstants.1m.test.ts` — 1M rows exist for the canonical variants, absent for haiku and for pinned variants, bare rows seed at 200k. 3 of 5 failed before the fix.
- Updated three existing tests that asserted the old behaviour (`contextWindowResolution.test.ts`, `claudeCliContextUsage.test.ts` ×2), each with the measurement in a comment.
- `npm run typecheck:prepush` — 0 errors.
- Full AI suite diffed against a stashed baseline: **38 failures before, 38 after**, with 5 added passing tests. No regressions. (Those 38 are pre-existing Windows failures in `bashParser` / `spawnCrashDiagnostics` / `claudeCodeEnvironment`, unrelated to this change.)

## Notes for reviewers

- **Pinned variants are excluded on purpose.** `resolveClaudeCliModelArg` collapses every `opus*` variant to the bare `opus` alias, so an `opus-4-7-1m` row would run Opus 5 at 1M while claiming to be Opus 4.7. Offering it would mislabel the model. There's a test asserting they stay out.
- **This is currently unrecoverable for users.** Nimbalyst passes `--model` explicitly on every launch, and an explicit flag beats the default saved through the CLI's `/model`. I confirmed neither `ANTHROPIC_MODEL` nor `CLAUDE_CODE_MODEL` overrides it, so there's no user-side workaround short of re-selecting the model inside every session.
- **The wider point, if you want it as a follow-up.** This assumption has now been wrong in both directions (200k on 2.1.175, 1M on 2.1.204, 200k on 2.1.220). It also can't be right for everyone at once, since 1M access is subscription-dependent and the correct value differs per account on the same build. There's no CLI model-listing command, but `modelUsage[...].contextWindow` from a `-p` call is authoritative *and* entitlement-aware. A one-time probe cached per CLI version would make this self-correcting. Happy to open that separately if it's wanted — I kept this PR to the minimal fix.
- Measured on Windows against CLI 2.1.220 with a Max subscription. The `[1m]`-vs-bare distinction is CLI-side, so it shouldn't be platform-dependent, but I can only speak for what I measured.
- Pre-push hook bypassed: `scripts/__tests__/check-push-authors.test.mjs` fails on Windows on pristine `main`, blocking `git push` entirely for Windows contributors. Unrelated to this change; the typecheck portion ran and passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Claude Code sessions can select a 1M-context model again, and the context meter no longer reports 1M for a session the CLI actually runs at 200k.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/services/ai/__tests__/claudeCliContextUsage.test.ts
+++ b/packages/electron/src/main/services/ai/__tests__/claudeCliContextUsage.test.ts
@@ -13,25 +13,26 @@ import {
  * `ai:tokenUsageUpdated` mechanism the SDK path uses.
  */
 describe('contextWindowForCliModel', () => {
-  it('returns 1M for current-gen base variants (opus/fable/sonnet run 1M natively)', () => {
-    // GitHub #825: on the current CLI plain opus/fable/sonnet report a 1M window
-    // at a flat price — the earlier 200k client-side windowing (CLI 2.1.175) is
-    // stale. Plain and -1m are identical for these.
-    expect(contextWindowForCliModel('claude-code-cli:opus')).toBe(1_000_000);
-    expect(contextWindowForCliModel('claude-code-cli:sonnet')).toBe(1_000_000);
-    expect(contextWindowForCliModel('claude-code-cli:fable')).toBe(1_000_000);
+  it('returns 200k for the bare variants — the CLI windows them at 200k', () => {
+    // Measured on CLI 2.1.220 via modelUsage[...].contextWindow:
+    //   opus -> 200000, fable -> 200000; the [1m] forms -> 1000000.
+    // Even the full id `claude-opus-5` reports 200000, so the window comes from
+    // the CLI's [1m] suffix rather than from the model.
+    expect(contextWindowForCliModel('claude-code-cli:opus')).toBe(200_000);
+    expect(contextWindowForCliModel('claude-code-cli:sonnet')).toBe(200_000);
+    expect(contextWindowForCliModel('claude-code-cli:fable')).toBe(200_000);
   });
   it('returns 1M for the -1m extended-context variants', () => {
     expect(contextWindowForCliModel('claude-code-cli:opus-1m')).toBe(1_000_000);
     expect(contextWindowForCliModel('claude-code-cli:sonnet-1M')).toBe(1_000_000);
     expect(contextWindowForCliModel('claude-code-cli:fable-1m')).toBe(1_000_000);
   });
-  it('returns 1M for legacy pinned variants (single row, no -1m duplicate) and 200k for haiku', () => {
-    // opus-4-6/opus-4-7/sonnet-4-6 are 1M models too — matched by EXACT variant,
-    // not family, so the mapping is precise rather than an accidental collapse.
-    expect(contextWindowForCliModel('claude-code-cli:opus-4-6')).toBe(1_000_000);
-    expect(contextWindowForCliModel('claude-code-cli:opus-4-7')).toBe(1_000_000);
-    expect(contextWindowForCliModel('claude-code-cli:sonnet-4-6')).toBe(1_000_000);
+  it('returns 200k for legacy pinned variants and for haiku', () => {
+    // resolveClaudeCliModelArg collapses every pinned opus variant to the bare
+    // `opus` alias before it reaches the CLI, so these get the 200k window too.
+    expect(contextWindowForCliModel('claude-code-cli:opus-4-6')).toBe(200_000);
+    expect(contextWindowForCliModel('claude-code-cli:opus-4-7')).toBe(200_000);
+    expect(contextWindowForCliModel('claude-code-cli:sonnet-4-6')).toBe(200_000);
     expect(contextWindowForCliModel('claude-code-cli:haiku')).toBe(200_000);
     expect(contextWindowForCliModel(undefined)).toBe(200_000);
   });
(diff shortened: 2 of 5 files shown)


ghinkle:
Thanks for this — you gave me what I needed to figure out the complexity of the problem and the differences between Max and Pro plans.

So 200k can be due to: Pro plan (Opus 1M is behind usage credits there, unlike Max), CLAUDE_CODE_DISABLE_1M_CONTEXT=1, or an ANTHROPIC_BASE_URL gateway. That last one turned out to be a genuine bug on our side, on the CLI provider specifically — we point it at a loopback observation proxy, and Claude Code treats that as a gateway and skips the 1M auto-upgrade. Verified on one account minutes apart: direct → 1000000; through the proxy → 200000, with context-1m-2025-08-07 missing from the outbound anthropic-beta header. That's ours to fix, and I wouldn't have gone looking without this PR.

One correction that matters: "it self-corrects once real modelUsage arrives" only holds on the SDK path. The CLI proxy has no per-model window in AssembledUsage, so contextWindowForCliModel is the permanent denominator there — which is why the CLI path needs its own signal rather than a constant.

I'm going to take another shot at this now that the pieces are clear: keep the bare rows at 1M, restore the -1m picker rows (the documented opt-in for Pro, and the only route to 1M behind our proxy), and derive the CLI-path window from the context-1m beta header the proxy already sees — no probe call needed. Your issue #989 and this PR are what got us there, and the commit will say so.

ghinkle:
Thanks for the help. Reworked and closed this in 26b6f8d52fa864c9387d9ec7326555c2bacede69

Issue #1078, opened 2026-07-28
Persist and restore full window state across restarts


Yogitmeister:
## Desired end state

Nimbalyst should persist and restore the full desktop main-window state across restarts:

* position
* restored size
* maximized state
* fullscreen state, where supported

A normal relaunch should return the user to the same visible window state they had when they quit, including a maximized or fullscreen window rather than only its pre-maximize geometry.

## Why this matters

Window state is part of a desktop application's continuity. Restoring only `{ x, y, width, height }` cannot represent maximized/fullscreen state and can reopen a maximized window at stale bounds.

## Scope considerations

The persisted format should remain backward-compatible: saved state from older versions will not contain the new flags and should safely fall back to `false`.

The initial bug fix can cover maximized state for the main application window. This feature request tracks the broader, explicit product contract: all persisted window state should include position, size, maximized state, and fullscreen state where applicable.

## Related bug

This request is prompted by nimbalyst/nimbalyst#1077, which documents the currently lost maximized state and stale restored bounds after restart.

karlwirth:
Closing this as delivered.

**What shipped:** Position, restored size and maximized state persist and restore.

**What did not:** persist fullscreen state and save normal bounds for maximized windows.

We have filed that ourselves as #1310 so it stays tracked and you do not need to re-file anything.

If the main problem you reported still happens on a current build, reopen this and we will pick it straight back up.

PR #1079, opened 2026-07-28
fix(electron): restore maximized windows after restart

### LINKED ISSUE #1077: Main window does not restore maximized state or current bounds after restart
## Summary

Nimbalyst does not restore a desktop window's position/size correctly across launches when it was maximized. On the next launch, it opens un-maximized using the stale pre-maximize bounds, often at an unexpected position.

## Steps to reproduce

1. Launch Nimbalyst and move/resize its main window.
2. Maximize the window.
3. Quit Nimbalyst normally.
4. Launch Nimbalyst again.

## Expected behavior

The main window should reopen in the same user-visible state: its prior position and size when restored, and maximized when it was maximized at quit time.

## Actual behavior

The window always restarts un-maximized. Its geometry is the old restored (pre-maximize) position and size rather than the maximized state the user last saw.

## Confirmed implementation cause

`packages/electron/src/main/session/SessionState.ts` saves `window.getBounds()` only. Electron returns the restored/pre-maximize bounds for a maximized `BrowserWindow`, so those values are stale for this purpose.

`packages/electron/src/main/window/WindowManager.ts` accepts and restores only `{ x, y, width, height }`; it does not persist a maximized-state flag and never calls `window.maximize()` during restoration.

## Suggested fix

Persist `window.isMaximized()` with the saved bounds and, after creating the restored window, call `window.maximize()` in the `ready-to-show` handler when that saved flag is true. Old saved state without the new field should default to `false`.

## Environment

- Desktop: Windows
- Affected area: Electron main-process window/session state

Yogitmeister:
## Summary

- persist `BrowserWindow.isMaximized()` alongside saved session bounds
- restore maximized windows in `ready-to-show`, while old saved state safely defaults to unmaximized
- add a restart regression test for maximized windows

## Validation

- `git diff --check`
- Targeted Playwright test was not runnable in this isolated checkout because the checked-out dependency tree is absent and the shared source checkout's TypeScript/Playwright junction resolves to a removed worktree. No dependency installation was performed for this focused patch.

Fixes #1077

Related to #1078


ghinkle:
Thanks for the fix. One startup interaction needs correction before merge: WindowManager.ts:591-599 calls maximize() before deferShowUntilAppActive, but Electron’s maximize() shows hidden windows and therefore bypasses the no-focus-steal guard. Please move maximization inside the guarded show action and add coverage for that ordering.

Yogitmeister:
Will do.  Thanks!On 30 Jul 2026, at 1:14, Greg Hinkle ***@***.***> wrote:﻿ghinkle left a comment (nimbalyst/nimbalyst#1079)
Thanks for the fix. One startup interaction needs correction before merge: WindowManager.ts:591-599 calls maximize() before deferShowUntilAppActive, but Electron’s maximize() shows hidden windows and therefore bypasses the no-focus-steal guard. Please move maximization inside the guarded show action and add coverage for that ordering.

Yogitmeister:
Addressed the startup ordering @ghinkle flagged. `window.maximize()` no longer runs in `ready-to-show` ahead of the guard — since Electron's `maximize()` reveals a hidden window, it was bypassing the no-focus-steal path. Extracted the reveal logic into `revealReadyWindow()` and moved `maximize()` inside the guarded show action, after the show call, so a restored maximized window resizes an already-visible window instead of revealing it early. Added `revealReadyWindow.test.ts` asserting maximize never runs before the app-active guard fires and only after `showInactive()` once it releases (the test fails against the previous ordering). Also merged `main` to clear the conflict. Typecheck + window/session suites green (66 tests).

Issue #868, opened 2026-07-15
Main event loop blocked by synchronous git-root/status resolution (findGitRootForFile existsSync storm) under multi-session editing → renderer OOM + hang


co-cy:
## Summary

Under heavy multi-session editing of a large monorepo, the main process spends most of its CPU in **synchronous `fs.existsSync`** inside `findGitRootForFile` / `GitStatusService.getFileStatus` (the `git:get-file-status` IPC handler). Because these run synchronously on the main thread, they block the event loop. The renderer's IPC to main then backs up, the renderer balloons to ~4.6 GB and floods its IPC pipe, goes unresponsive, and the app hangs.

The app's own auto-captured CPU profiles caught this red-handed: `existsSync` is **18–42%** of main-process CPU during the hang.

## Environment

- `@nimbalyst/electron` 0.66.9, Linux (WSLg).
- Workspace: a large Go monorepo (deeply nested paths like `soul-stack/keeper/internal/auth/ldap/ldap.go`).
- 5 concurrent Claude CLI sessions editing many files at once.

### 1. CPU profiles (auto-captured by Nimbalyst during the hang)

Two consecutive profiles from the hang window (`logs/cpu-2026-07-15T07-3x.cpuprofile`), self-time by function:

```
cpu-…07-36-36 (5.77s):   18.2% existsSync | 13.3% inspector.post | 2.0% lstat |
                          0.8% findGitRootForFile @ index.js:225065 | 0.4% getFileStatus @ index.js:225101
cpu-…07-33-54 (5.60s):   41.7% existsSync | 10.0% spawn | 5.9% GC |
                          1.5% findGitRootForFile @ index.js:225065 | 1.2% getFileStatus @ index.js:225101
```

`existsSync` dominates; its callers are `findGitRootForFile` and `getFileStatus`, plus `spawn` (git subprocess).

### 2. The code

`findGitRootForFile` walks up the directory tree calling `existsSync(dir/.git)` at every level, **with no caching** (main `index.js:225065`):

```js
function findGitRootForFile(filePath, boundary) {
  ...
  let dir = path.dirname(fileAbs);
  while (true) {
    try { if (fs.existsSync(path.join(dir, ".git"))) return dir; } catch {}
    if (dir === boundaryAbs) break;
    const parent = path.dirname(dir);
    if (parent === dir) break;
    ...
    dir = parent;
  }
  return null;
}
```

[...]

PR #1080, opened 2026-07-28
fix(electron): keep Git status off the main event loop

### LINKED ISSUE #1063: Main process can freeze during heavy session-database and Git activity
## Summary

The packaged application can freeze for seconds or minutes during heavy session-database activity. Long freezes can make the UI unresponsive and may lead users to terminate the process.

## Public-safe evidence

Diagnostic logs show:

- severe main-process event-loop lag;
- slow session-list and session-load IPC operations;
- a saturated SQLite worker during history-retention queries;
- synchronous Git status child-process calls timing out in the main process;
- unbounded diagnostic profile accumulation increasing disk pressure.

Exact local paths, database contents, session counts, hardware identifiers, timestamps, and captured profiles have been removed from this public report.

## Suggested fixes

1. Replace synchronous Git status execution in the main process with an asynchronous child process.
2. Bound and paginate agent-message loading and provide retention/archival controls.
3. Make history pruning incremental rather than repeatedly scanning full tables.
4. Schedule or bound large backups and account for low-disk conditions.
5. Cap diagnostic profile retention.

## Expected behavior

Background database, Git, backup, and profiling work must not block the Electron main event loop or make the application unusable.

### Nimbalyst version

0.71.2

### Operating system

Windows

Yogitmeister:
## Summary
- Replace the synchronous `git status --porcelain` wait on Electron's main process with asynchronous `execFile` work.
- Coalesce concurrent status requests for the same repository root.
- Prevent a completion that began before cache invalidation from repopulating the cache afterward.

## Coverage
- Adds focused coverage for concurrent coalescing, retry after a Git failure, and cache invalidation while a status request is in flight.

## Validation
- `git diff --check` passed.
- The focused Vitest suite could not run locally: this clean containment clone has no installed dependency toolchain, and the existing executable from another checkout could not resolve its dependencies against this clone. No dependency install was performed.

Fixes #1063

Related to #868; #895 previously addressed Git-root caching but did not remove the synchronous Git status wait.

### DIFF
--- a/packages/electron/src/main/services/GitStatusService.ts
+++ b/packages/electron/src/main/services/GitStatusService.ts
@@ -1,4 +1,4 @@
-import { execSync } from 'child_process';
+import { execFile, execSync } from 'child_process';
 import { existsSync, statSync } from 'fs';
 import { dirname, isAbsolute, join, relative, resolve } from 'path';
 import { getUntrackedFilesInDirectory, isGitAvailable, logEbadfDiagnostic } from '../utils/gitUtils';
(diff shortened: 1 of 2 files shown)


Yogitmeister:
Windows exact-head validation completed for `84a3ed8f80d038ceb557b758e346c40d55f5669d`:

- focused `GitStatusService.executeGitStatus` regression: 3/3 tests passed;
- Electron TypeScript check passed;
- extension SDK, runtime, and Electron production builds passed;
- all 19 packaged extra-resource sources and both native-binary checks passed;
- an unpacked Windows artifact was produced from the exact commit.

Production evidence showed the synchronous Git-status path blocking Electron's main loop for roughly 3.6–4.0 seconds under current multi-session load. This PR removes that proven first choke and coalesces same-root work. It does not by itself close the separate database/backpressure, recursive-fanout, or full process-disappearance investigations.

Requesting maintainer merge when ready.

PR #1087, opened 2026-07-29
fix(git): batch NIM-479 untracked expansion

Yogitmeister:
## Summary
- replace per-directory synchronous untracked expansion with one bounded async Git call per root/generation
- share cancellation-safe expansion snapshots across workspace and worktree/session callers
- add real-Git coverage for 100 directories, ignores, nested repositories, event-loop responsiveness, and invalidation

## Tests
- npx vitest run packages/electron/src/main/services/__tests__/GitStatusService.untrackedExpansion.test.ts
- npm run typecheck --workspace=@nimbalyst/electron

Full workspace typecheck and pre-push suite were attempted but have unrelated existing failures in other workspaces/tests.


ghinkle:
Thanks for this. You're heading in the right direction, but this conflicts with other work around GIT\_INDEX\_FILE and lock contention that I've pushed. This PR also adds a second coalescing/invalidation mechanism (`rootInFlight` / `rootGenerations` / `AbortController`) parallel to the one #1080 just introduced, so the right end state is one unified mechanism rather than two stacked.


Rather than ask you to rebase across all of that, I'm implementing the batching here, borrowing your design.


Closing this one, but #1080 was a real fix and this diff sharpened our thinking on the expansion path. Thanks again.

Yogitmeister:
🙏❤️🤘On 30 Jul 2026, at 5:19, Greg Hinkle ***@***.***> wrote:﻿ghinkle left a comment (nimbalyst/nimbalyst#1087)
Thanks for this. You're heading in the right direction, but this conflicts with other work around GIT_INDEX_FILE and lock contention that I've pushed. This PR also adds a second coalescing/invalidation mechanism (rootInFlight / rootGenerations / AbortController) parallel to the one #1080 just introduced, so the right end state is one unified mechanism rather than two stacked.
Rather than ask you to rebase across all of that, I'm implementing the batching here, borrowing your design.
Closing this one, but #1080 was a real fix and this diff sharpened our thinking on the expansion path. Thanks again.

PR #1101, opened 2026-07-30
fix: stop wrapping static-key HTTP MCP servers in mcp-remote for Claude Code

Derazien:
## Summary

Every `type: "http"` MCP server is rewritten to `npx mcp-remote <url> --header ...` before the config is handed to an external CLI. `mcp-remote` exists to handle OAuth. A server authenticating with a plain static bearer header needs no OAuth, so for those the wrapper is pure cost.

The only escape is `usesNativeRemoteOAuth` (`MCPRemoteOAuth.ts:89`):

```ts
return Boolean(config.oauth?.clientId || config.oauth?.clientSecret);
```

That asks "does this server have OAuth *credentials*", when the question that matters is "does this server need OAuth *at all*". A static-header server answers no to the second and still gets wrapped.

Notably, the app already works this out. The server templates in `MCPServersPanel.tsx` tag each entry `authType: 'oauth' | 'api-key' | 'none'`, with the comment *"'oauth' uses mcp-remote for browser-based login, 'api-key' requires manual key"*. But `authType` is UI-only template metadata: it is never persisted into the server config and appears nowhere in `src/main` or `packages/runtime`. The determination is made and then discarded before the code that needs it.

## Measured cost

On a machine running 9 concurrent sessions:

- **136 processes / 4,641 MB** attributable to the wrapper chain, roughly **15 processes and 516 MB per session**
- staged migration confirmed the attribution: 46-47 processes per session at baseline, 41 after moving two servers, **31** after moving the rest. Predicted 15.1 saved, measured 15.
- **credential exposure**: native HTTP puts the token in a request header; the wrapper puts it in `argv`, readable by any process. 65 live processes were carrying a GitHub PAT on their command line.
- **silent server loss**: `isOAuthAuthorized` probes each HTTP server through mcp-remote. A static-key server that answers 401 is classified as an unauthorized OAuth server and dropped. One server was being dropped from *every* session launch and the user never knew.

A/B, same PAT, `--strict-mcp-config`, identical prompt: native and wrapped both return the same login and exit 0, with identical org access (11/7/6 repos, matching a direct `api.github.com` baseline). **Startup latency was a wash** (8.8s/11.2s native vs 9.7s/9.8s wrapped) - no latency claim is made here.

## The change

New predicate `requiresMcpRemote(config, { nativeHttpSupported })`, used by both `convertHttpToStdio` and `isOAuthAuthorized`. Only `http` servers are affected; `sse` and `stdio` are untouched.

**`nativeHttpSupported` is opt-in per caller, and this is deliberate.** `processServerConfigForRuntime` has five callers, all external CLIs:

```
index.ts:1747  ClaudeCodeProvider        <- flag passed
index.ts:1769  OpenAICodexProvider
index.ts:1794  OpenAICodexACPProvider
index.ts:1816  CopilotCLIProvider
index.ts:1841  ClaudeCliLauncherConfig   <- flag passed
```

Only Claude Code was verified to speak HTTP natively. Codex, Codex ACP and Copilot share the identical path and there is no evidence either way, so they keep the wrapper until someone measures them. Flipping all five on one CLI's evidence would be a silent transport change for three others.

The `isOAuthAuthorized` short-circuit is scoped to `type === 'http'` for the same reason: `extractMcpRemoteConfig` also handles `sse` servers and stdio servers that are themselves `npx mcp-remote` invocations, and those must still be auth-checked.

## Testing

Test written first and confirmed failing, per `.claude/rules/end-to-end-verification.md`. 7 cases, all red before the change:

- static-header server on a native-HTTP CLI is not wrapped
- the same server IS wrapped when the CLI has no native HTTP
- an `oauth` block without credentials still wraps (dynamic registration needs mcp-remote)
- explicit OAuth credentials still bypass, preserving the existing escape hatch
- `stdio` and `sse` never affected
- an unspecified `nativeHttpSupported` defaults to today's behaviour

`npx tsc --noEmit` in `packages/electron`: 46 errors before, 46 after, none in the touched files, measured by stashing rather than assumed. 48 passing across the MCP suites.

## Notes for reviewers

- **Behaviour change worth calling out:** a server that relies on mcp-remote's dynamic OAuth registration (tokens in `~/.mcp-auth`) keeps the wrapper, because it has an `oauth` block. Only servers with *no* `oauth` key at all change path. That was the deliberate boundary.
- The obvious follow-up is to persist `authType` into the server config so the backend can use the answer the UI already computed, rather than inferring it from the shape of the `oauth` key. I did not do that here because it touches the settings schema and this fix does not need it.
- Pre-push hook bypassed: I develop in a git worktree whose `node_modules` is junctioned to the main checkout, so package-local dependencies do not resolve and the hook's build step fails for unrelated reasons. Typecheck and unit tests were run manually, results above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- HTTP MCP servers that authenticate with a static key now connect directly on Claude Code sessions instead of being routed through an extra helper process.
 - Desktop AI notifications now lead with the originating session name and open that exact session and project, including child sessions, instead of following whichever project is currently visible.
 - File reveal menus now name Finder on macOS, Explorer on Windows, and the containing folder on Linux.
 - AI sessions that preview a web page no longer strand a blank, unclosable window on a second monitor.
--- a/packages/electron/src/main/index.ts
+++ b/packages/electron/src/main/index.ts
@@ -1756,12 +1756,15 @@ app.whenReady().then(async () => {
         const enabledServers: Record<string, any> = {};
         for (const [name, config] of Object.entries(allServers)) {
             if (isMCPServerEnabledForProvider(config as MCPServerConfig, MCP_PROVIDER_IDS.CLAUDE_AGENT)) {
-                const isAuthorized = await mcpConfigService.isOAuthAuthorized(config as MCPServerConfig);
+                // Claude Code speaks HTTP natively, so a server with no OAuth is
+                // passed through instead of being wrapped in npx mcp-remote.
+                const claudeHttp = { nativeHttpSupported: true };
+                const isAuthorized = await mcpConfigService.isOAuthAuthorized(config as MCPServerConfig, claudeHttp);
                 if (!isAuthorized) {
                     logger.mcp.info(`[MCP] Skipping unauthorized OAuth server for Claude Agent: ${name}`);
                     continue;
                 }
-                enabledServers[name] = mcpConfigService.processServerConfigForRuntime(config as any);
+                enabledServers[name] = mcpConfigService.processServerConfigForRuntime(config as any, claudeHttp);
             }
         }
         return enabledServers;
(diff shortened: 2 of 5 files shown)


ghinkle:
Thanks — the measurement work here is great, and the opt-in-per-caller boundary is exactly the right call.

One thing I want to flag rather than send back to you: the new short-circuit in `MCPConfigService.isOAuthAuthorized` runs after the `useMcpRemoteForNativeOAuth` guard but doesn't account for it. `requiresMcpRemote` returns `false` for any server with `oauth.clientId`/`clientSecret`, so the Codex, Codex ACP and Copilot loaders — which pass `useMcpRemoteForNativeOAuth: true` — now report native-OAuth HTTP servers as authorized and skip the mcp-remote auth-cache check, so unauthorized ones get passed through instead of dropped. It's a one-condition fix (`&& !options.useMcpRemoteForNativeOAuth`), and I'll handle it in a follow-up rather than bounce the PR.

Merging this — thanks again.

PR #1107, opened 2026-07-31
fix: prevent quick open shortcut overflow

### LINKED ISSUE #1102: Shortcut text spills out of its badge in Unified Quick Open tabs
Unified Quick Open shortcut text overflowing its badges

The shortcut text in Unified Quick Open does not stay inside its bordered badge when enough tabs are visible. In the screenshot, the shortcut after **Prompts** spills out of the badge and runs into **Projects**. The same thing happens around the **Memory** tab on the right.

I can reproduce it on Linux by opening Unified Quick Open with `Ctrl+Shift+P` while the Memory tab is present. It remains visible at every app zoom level I tested.

The shortcut text should stay inside its badge and the tabs should not overlap. If there is not enough room, hiding some shortcut badges or making the tab row scroll would both be clearer than letting the text spill into the next tab.

Environment:

- Nimbalyst 0.71.3
- Linux, GNOME 50
- 2560 × 1440 at 100% display scaling
- Reproduces across all tested app zoom levels

The relevant markup appears to be in `packages/electron/src/renderer/components/UnifiedQuickOpen.tsx`. The tabs are equal-width flex items, so the shortcut badge can be compressed more than its non-wrapping text allows.

Misaki14987:
## Summary

  Prevents keyboard shortcut badges in Unified Quick Open from shrinking beyond their
  contents and overlapping adjacent tabs.

  Tabs now keep their intrinsic minimum width, and the tab strip scrolls horizontally when
  the available width is insufficient.

  ## Related issue

  Fixes #1102

  ## Testing

  Verified with:

  - 16 focused Unified Quick Open unit tests
  - Electron Playwright E2E coverage at 2560px and 700px widths
  - `npm run typecheck:prepush`
  - Production Electron build

  ## Notes for reviewers

  The responsive fallback uses horizontal scrolling so tab labels and shortcuts remain
  visible instead of being hidden.

  Before:
before

  After:

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Unified Quick Open tabs now keep keyboard shortcuts inside their badges and scroll instead of overlapping at narrow widths. (#1102)
 - Desktop AI notifications now lead with the originating session name and open that exact session and project, including child sessions, instead of following whichever project is currently visible.
 - File reveal menus now name Finder on macOS, Explorer on Windows, and the containing folder on Linux.
 - AI sessions that preview a web page no longer strand a blank, unclosable window on a second monitor.
--- a/packages/electron/e2e/core/unified-quick-open.spec.ts
+++ b/packages/electron/e2e/core/unified-quick-open.spec.ts
@@ -1,0 +1,72 @@
+import { test, expect } from '@playwright/test';
+import type { ElectronApplication, Page } from 'playwright';
+import * as fs from 'fs/promises';
+import {
+  createTempWorkspace,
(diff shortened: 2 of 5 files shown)


ghinkle:
Thanks for the contribution. Apologies, but I had already fixed it in 74028fd9f31c30701910cf3eadb36905204f18f2 and was in the process of pushing.
I ended up going with symbols.

Misaki14987:
> Thanks for the contribution. Apologies, but I had already fixed it in [74028fd](https://github.com/nimbalyst/nimbalyst/commit/74028fd9f31c30701910cf3eadb36905204f18f2) and was in the process of pushing. I ended up going with symbols.

Thanks for the update, and for getting this fixed so quickly. This was my first attempt at contributing to an open-source project, and I learned a lot from the process. I'm very interested in Nimbalyst and would love to keep contributing, so I'll keep an eye out for another well-scoped issue.

PR #1109, opened 2026-07-31
Route notification clicks to the session's own project

### LINKED ISSUE #1108: [Bug]: Notification click does not switch projects in the multi-project rail, landing on a missing chat
### Describe the bug

**What happened:**
With several projects open in the rail (one window), clicking an OS notification for a session that belongs to a project other than the visible one opens the AGENT panel but does not switch projects. The session id is then selected inside the project that was already on screen, where it does not exist, so the chat comes up empty.

**What I expected:**
Clicking the notification brings up the project the session belongs to, then selects that session.

**Steps to reproduce:**
1. Open two projects in the multi-project rail (A and B).
2. Start an agent session in project A that will ask a question or finish.
3. Switch the rail to project B.
4. Click the notification for the project A session when it arrives.
5. The rail stays on B and the chat is empty.

**Where it goes wrong:**
`NotificationService.handleNotificationClick` resolves the right window (`findWindowByWorkspace` matches `additionalWorkspacePaths`, so the rail window is found), but the `notification-clicked` event it sends carries only `sessionId` — the workspace path it already holds is dropped at the IPC boundary. The renderer handler then reads `sessionListWorkspaceAtom`, i.e. the *visible* project, and selects the foreign session id there.

Two knock-on effects:

- `selectedWorkstreamAtom` is a per-workspace family and `setSelectedWorkstreamAtom` persists, so the broken selection is written into the visible project's workspace state and survives a restart.
- The registry is replaced wholesale per workspace, so `registry.get(sessionId)` misses. A child session of a workstream is therefore not recognised as a child (the parent is never selected), and the type falls back to `session` even for a worktree or workstream.

Single-project windows are unaffected — there the routed window's active workspace is already the right one.

### Nimbalyst version

`main` at 693fd76 (0.71.2 in the changelog).

### Operating system

macOS.

### Note

Found by reading the code while working in this area, not from a packaged build — I have not run the reproduction above against a release binary. A fix with regression tests is in the linked pull request.

zotovprog:
## Summary

In the multi-project rail one window hosts several projects, and a notification click used to select the notified session inside whichever project was visible — a session id that does not exist there, so the chat came up empty.

`handleNotificationClick` already knows the session's workspace (it routes the window by it), but the `notification-clicked` event carried only `sessionId`. The renderer then fell back to `sessionListWorkspaceAtom`, the visible project.

The fix sends the owning project path with the event and activates that project before selecting:

- The path is resolved through `resolveProjectPath`, the same helper window lookup uses, so worktree sessions — which notify from the worktree directory, not a rail entry — map to their parent project.
- Activation goes through `activeWorkspacePathAtom`, the single funnel whose subscriber notifies the main process, then waits for the window mode, workstream states and session list of the new path before the session type is read. The registry is replaced wholesale per workspace, so reading it earlier would still see the previous project.
- `initWorkstreamState` / `initWindowMode` are re-pointed in the same step rather than left to AgentMode's mount effect, which only runs on the next render. Without that, the `agent` mode switch and the workstream state would be persisted against the project the user just left.

Two knock-on effects go away with it: the foreign selection is no longer persisted into the wrong project's workspace state, and a child session is correctly resolved to its parent workstream instead of being mis-selected as a root session.

Unchanged on purpose: when the target is not a project this window hosts — a single-project window, or one opened directly on a worktree whose parent is not a rail entry — the click still selects in the visible project rather than being dropped. The main process routed it here deliberately. Callers that pass an empty workspace path (some `showBlockedNotification` call sites) take the same path.

## Related issue

Closes #1108

## Testing

New `sessionStateListeners.notificationWorkspace.test.ts`: the rail activates the notified project, the session is selected there, and the foreign id is never written into the visible project's selection — plus the three fallbacks (same-project notification does not switch, empty path, target not open in this window).

- `npm run typecheck` — clean
- `npm run test:prepush` — 906 files / 7071 tests passed, 7 files / 26 tests skipped

## Notes for reviewers

The fallback branch is the judgement call worth a look: dropping the click when the project cannot be brought forward would be stricter, but it would regress windows opened directly on a worktree, which work today.

I have not exercised the flow against a packaged build — the diagnosis and the tests come from the code paths. If you would rather see a manual repro before merging, say so and I will run one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Clicking a session notification now brings up the project that session belongs to, instead of landing on a missing chat in whichever project was on screen.
 
 ### Removed
 <!-- Removed features go here -->
--- a/packages/electron/src/main/services/NotificationService.ts
+++ b/packages/electron/src/main/services/NotificationService.ts
@@ -8,6 +8,7 @@ import { Notification, BrowserWindow, app, ipcMain, shell } from 'electron';
 import { logger } from '../utils/logger';
 import { isOSNotificationsEnabled, isNotifyWhenFocusedEnabled, isSessionBlockedNotificationsEnabled } from '../utils/store';
 import { findWindowByWorkspace } from '../window/WindowManager';
+import { resolveProjectPath } from '../utils/workspaceDetection';
 
 const NOTIFICATION_OUTCOME_TIMEOUT_MS = 2_000;
 
@@ -342,10 +343,16 @@ class NotificationService {
     targetWindow.focus();
     targetWindow.show();
 
-    // If session ID provided, send IPC event to switch to that session
+    // If session ID provided, send IPC event to switch to that session.
+    // The owning project path travels with it: in the multi-project rail one
+    // window hosts several projects, and without it the renderer would select
+    // the session inside whichever project happens to be visible. Worktree
+    // sessions notify from the worktree directory, which is not a rail entry,
+    // so resolve to the parent project the way window lookup already does.
     if (options.sessionId) {
       targetWindow.webContents.send('notification-clicked', {
         sessionId: options.sessionId,
+        workspacePath: resolveProjectPath(options.workspacePath),
       });
     }
   }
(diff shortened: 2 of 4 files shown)


zotovprog:
Superseded by c54a7202e on `main` ("fix: open the originating session from a desktop notification"), which lands the same fix more thoroughly — it registers a project that is not yet in the rail, validates the session before any visible selection, and queues the navigation while a workspace window is still opening. Closing in favour of that.

PR #1110, opened 2026-07-31
fix: stop WebSearch and WebFetch failing at max effort in CLI sessions

Derazien:
## The bug

At effort `max`, every WebSearch and WebFetch in a `claude-code-cli` session fails:

```
400 output_config.effort 'max' is not supported when thinking is disabled on this model.
Use effort 'high' or below, or enable thinking.
```

The same session works at `high`, and a plain `claude` in a terminal works at `max`.

## Root cause

Not what the error suggests. Extended thinking is never disabled: main-loop requests keep `thinking: {"type":"adaptive"}` behind the proxy and return 200. The failure is confined to the web-tool side query.

Claude Code runs WebSearch/WebFetch as a side query on the small/fast model with `thinking` hardcoded off. Selecting that model is gated on the base URL being first-party:

```js
function JJt(){                        // is the small/fast model available?
  if (Z.ANTHROPIC_SMALL_FAST_MODEL !== void 0) return true;
  let e = kn(), t = e === "firstParty" && (Kd() || Qji()) || c5(e);
  return Z.ANTHROPIC_DEFAULT_HAIKU_MODEL !== void 0 || t
}
function Kd(){ if (Z._CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL) return true; return fWr(); }
function A1e(e){ try { return ["api.anthropic.com"].includes(new URL(e).host) } catch { return false } }
```

We point the CLI at the loopback observation proxy via `ANTHROPIC_BASE_URL`, so `Kd()` is false, the small/fast model is considered unavailable, and the side query falls back to the **main** model, dragging the session's `output_config.effort` with it. `effort: max` plus thinking-disabled is rejected by the API.

This is the same gateway heuristic documented as a known limitation in #989 / #990.

## Measured (CLI 2.1.220, outbound /v1/messages tee'd)

| Setup | Web-tool side query | Result |
|---|---|---|
| no base URL | not observable | works |
| proxy | `model=claude-opus-5 thinking=disabled output_config={"effort":"max"}` | 400, both trials |
| proxy + flag | `model=claude-haiku-4-5-20251001 thinking=disabled output_config=undefined` | 200, both trials |

The fix is also a cost fix. `OM()` feeds quota checks, token counting, API key verification, prompt hooks and a mid-turn helper. All of those were silently running on the main model instead of the small/fast one for the whole session.

## The fix

Spawn the CLI with `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL=1`, but only when the proxy really does forward to `api.anthropic.com`. A user-configured `claudeCode.apiUpstreamUrl` gets no declaration: we would be vouching for someone else's endpoint, and the CLI would send tool_reference blocks and first-party-only betas the middleware may not forward.

The proxy is unchanged and still in the request path. The flag only changes what the CLI believes about it.

## Scope notes

- This does **not** restore the 1M auto-upgrade from #989. Measured: with the flag, plain `opus` still sends no `context-1m-2025-08-07`; explicit `opus[1m]` still does. The `[1m]` picker rows remain the answer.
- The SDK path never had this bug, it does not set `ANTHROPIC_BASE_URL`.
- The flag is an internal CLI escape hatch. If a future CLI drops it the failure mode is a return to current behavior, not a broken session, so it stays a best-effort hint rather than a launch precondition.
- Secondary and unfixed here: the CLI's own retry predicate should have recovered from this 400 by retrying without the effort parameter, and does not match this error shape.

## Tests

`proxyPassthroughEnv.test.ts` covers the upstream predicate (default, explicit api.anthropic.com, custom loopback, unparseable, blank). `ClaudeCliSessionLauncher.test.ts` covers the launcher forwarding the env and omitting it when the proxy supplies none.


ghinkle:
Thanks for the fix!

PR #1207, opened 2026-08-10
fix: keep session-list relative time updating on idle sessions (#1200)

### LINKED ISSUE #1200: Session list shows incorrect relative times
### Describe the bug

The session list in the left sidebar constantly shows incorrect relative times, such as “32 mins ago” for sessions from roughly 15 hours ago. This has been happening for months and makes it impossible to tell which sessions are actually recent.

Environment: Nimbalyst 0.72.8 on Windows 10.0.26200 (x64).

### Nimbalyst version

_No response_

### Operating system

_No response_

forcewalkerneo:
Fixes #1200.

## Problem
The relative-time label in the session list (e.g. "32 mins ago") is computed in a `useMemo` keyed only on the session's timestamp. That timestamp changes only when the session gets new activity, so once a session goes idle the label freezes and never advances — a session that's actually 15 hours old keeps showing "32 mins ago".

## Fix
Add a once-a-minute tick that re-runs the memo, so the label reflects real elapsed time on idle sessions. This mirrors the existing approach in `TeamMode/Inbox/InboxSection.tsx`, which already solved the identical "a quiet label still ages" problem with a 60s interval.

## Test
Adds `SessionListItem.relativeTimeTick.test.tsx` (jsdom + fake timers): renders an idle row showing "Just now", advances the clock one minute, asserts it becomes "1 min ago". Confirmed it fails on the pre-fix code and passes with the fix.

Scoped to `SessionListItem.tsx` (the primary session list the issue names). `SessionRelativeTime.tsx` has the same pattern and could get the same treatment in a follow-up.

### DIFF
--- a/packages/electron/src/renderer/components/AgenticCoding/SessionListItem.tsx
+++ b/packages/electron/src/renderer/components/AgenticCoding/SessionListItem.tsx
@@ -401,6 +401,17 @@ export const SessionListItem = memo<SessionListItemProps>(function SessionListIt
   const timestamp = sortBy === 'updated' ? (effectiveUpdatedAt || createdAt) : createdAt;
   const timestampLabel = sortBy === 'updated' ? 'updated' : 'created';
 
+  // A quiet session still ages: relativeTime is derived from a fixed timestamp,
+  // so without a periodic re-render the "X ago" label sits frozen until the
+  // session next has activity (#1200). One coarse tick a minute matches the
+  // finest granularity getRelativeTimeString renders — same approach as the
+  // Inbox section's relative labels.
+  const [relativeTimeTick, setRelativeTimeTick] = useState(0);
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


forcewalkerneo:
Hi, first-time contributor here. I traced this to the useMemo that only
re-runs when the session timestamp changes, so an idle session freezes its
label. I mirrored the once-a-minute tick already used in InboxSection.tsx
rather than introduce a new pattern, and added a test that fails on the
current code and passes with the fix.

Two things I'm happy to adjust if you'd prefer:

- One interval per list item matches the existing InboxSection approach. If
  you'd rather a single shared ticker for the whole list, I can rework it
  that way.
- SessionRelativeTime.tsx has the identical issue. I kept this PR scoped to
  the session list named in #1200, but I can fold that in here or as a
  follow-up, whichever you prefer.

Thanks for building this.

PR #1216, opened 2026-08-11
fix: keep typed characters when editing a tracker table cell (#1199)

### LINKED ISSUE #1199: Table view: first typed character is auto-selected and gets replaced by the second keystroke
## Summary

When typing into a focused text cell in a tracker table view, the first typed character is left selected, so the next keystroke replaces it instead of appending. A five-letter word ends up as its last four letters.

## Steps to reproduce

1. Open a tracker in table view.
2. Focus a text cell (either by pressing Enter in the cell above, which moves focus down, or by arrowing to / clicking the cell).
3. Start typing a word, e.g. `Alpha`.
4. Press Enter to submit and repeat down the column.

## Expected behavior

Typing into a focused cell starts an empty edit, and every character typed is kept: typing `Alpha` produces `Alpha`.

## Actual behavior

The first character is inserted but left highlighted/selected, so the second character overwrites it. Typing `Alpha` produces `lpha`. Happens every time, regardless of how focus arrived at the cell.

This undercuts what is otherwise a very efficient type-Enter-type-Enter flow for filling out a table column.

## Environment

- App version: 0.72.8
- OS: macOS (darwin 25.5.0, arm64)
- Electron: 43.2.0

## Logs

Not collected.

### Nimbalyst version

_No response_

### Operating system

_No response_

forcewalkerneo:
Fixes #1199.

## Problem
In a tracker table view, typing into a focused cell drops the first character.
RevoGrid opens the editor seeded with the typed character, and the editor calls
input.select() on mount, so that character is left selected and the next
keystroke replaces it. Typing "Alpha" produces "lpha".

## Fix
Only select-all when re-editing an existing value (Enter/F2, where editCell.val
still equals the stored value). When the edit began from a keystroke (val differs
from the cell's stored model value), place the caret at the end instead, so the
rest of the word appends. This preserves type-to-replace for Enter/F2 while fixing
the type-to-append flow the issue describes -- option (b) from the discussion.

Scoped to the text-like input editor (text/multiline/number/user/url).
setSelectionRange is guarded for input types that don't support it (e.g. number).

## Test
Adds a jsdom unit test for the editor's caret placement: a keystroke-style edit
collapses the caret at the end (append), and an Enter/F2-style edit selects all
(replace). Confirmed the keystroke case fails on the current code and passes with
the fix.

## Verification
I reproduced the original bug locally on Linux (tracker table view, multiple
cells: typing "Alpha" produced "lpha"), so it is not macOS-specific. The unit
test covers the editor's selection logic; a quick check on macOS or a live run
would confirm the Enter/F2 replace path is unchanged.

### DIFF
--- a/packages/collab-client/src/trackers-ui/__tests__/trackerGridEditors.caret.test.ts
+++ b/packages/collab-client/src/trackers-ui/__tests__/trackerGridEditors.caret.test.ts
@@ -1,0 +1,95 @@
+// @vitest-environment jsdom
+import { describe, it, expect, afterEach } from 'vitest';
+import { createTrackerCellEditor, shouldSelectAllOnEdit } from '../grid/trackerGridEditors';
+
+// Minimal stand-in for RevoGrid's hyperscript: build a real <input>, seed its
+// value from the vnode's `value`, and hand the element to the `ref` callback so
+// the editor captures it exactly as RevoGrid would at mount.
+function mountTextEditor(editCell: unknown, kind = 'text') {
+  let input!: HTMLInputElement;
+  const createElement = (tag: string, props: any) => {
+    const el = document.createElement(tag) as HTMLInputElement;
+    if (props?.value != null) el.value = String(props.value);
+    if (typeof props?.ref === 'function') props.ref(el);
+    if (tag === 'input') input = el;
+    return el as unknown;
+  };
+  const factory = createTrackerCellEditor({ kind } as any) as any;
+  const editor = factory({}, () => {}, () => {});
+  editor.editCell = editCell;
+  const el = editor.render(createElement as any);
+  document.body.appendChild(el as Node);
+  return { editor, input };
+}
+
+afterEach(() => {
+  document.body.innerHTML = '';
+});
+
+describe('shouldSelectAllOnEdit (#1199)', () => {
+  it('does not select-all when the edit began from a keystroke', () => {
+    // RevoGrid seeds `val` with the typed character while the model still
+    // holds the original value.
+    expect(shouldSelectAllOnEdit({ val: 'A', prop: 'title', model: { title: '' } } as any))
+      .toBe(false);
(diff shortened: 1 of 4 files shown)


forcewalkerneo:
Re-ported this onto `main`. The original diff no longer applied: `52db2c692` moved the grid out of `packages/electron/.../TrackerMode/grid/` into `packages/collab-client/src/trackers-ui/grid/`, so the file this PR patched is gone.

The bug came through the move intact. `input?.select()` in `createInputEditor` is unchanged on `main`, and there is no commit referencing #1199, so nothing else fixed it in passing.

What changed since the first version:

- The caret decision is now a pure exported `shouldSelectAllOnEdit(editCell)` rather than inline logic, so it can be tested without a DOM.
- Tests moved to `packages/collab-client/src/trackers-ui/__tests__/`, matching where `trackerGridColumns.test.ts` already lives.
- Added cases for a numeric cell (model holds a number, `val` a string) and for a null stored value, both of which a naive identity compare gets wrong.

Checks on this branch: 8 new tests pass, 40 pass across `trackers-ui`, 216 pass across the electron `TrackerMode` suite, `tsc --noEmit` is clean on `@nimbalyst/collab-client`, and `check:collab-client-boundaries` is clean. I reverted the fix twice to confirm the tests actually fail without it: restoring `select()` kills the caret test, and swapping the comparison for `===` kills the numeric and null cases.

One thing worth being straight about. I have not re-run this in the app since the move. The original repro was mine on Linux against the old file, and the function is byte-identical, so I am confident the behaviour carried over, but that is inference rather than a fresh repro.

Two things I left alone. `createMultiselectEditor` and `createRelationshipEditor` carry the same `input?.select()`. Multiselect looks like the same bug. The relationship editor looks different and possibly worse, since its seed maps the incoming value through the candidate list and a typed character resolves to an empty string. I did not touch either, because I cannot verify them the way I verified this one, and guessing next to a proven fix seemed like the wrong trade. Happy to look at them separately if useful.

Since this now lives in `collab-client`, it covers the web console as well as the desktop grid.

PR #1313, opened 2026-08-18
feat: keep every project in one window with the project rail

huh12312:
## What this does

Nimbalyst opens one OS window per project. A "Multi-Project Mode" setting and a project rail already existed, but **no main-process code ever read the setting** — `getMultiProjectMode()` was referenced only by its own getter/setter — so every project-open still called `createWindow(...)` and spawned a window. The rail only ever filled via its own `+` button.

This makes the toggle mean what it says.

- Project-opens route through one mode-aware chokepoint, with the entry points that bypassed it funneled in: Open Recent, deep links, notification clicks, MCP `workspace_open`, tutorial, extension scaffolder
- Session restore collapses to a single window whose rail is seeded from the saved session
- New **Window > Merge All Windows** folds windows you already have open, flushing unsaved editors and waiting for confirmation before closing any donor
- File watchers stay warm for projects sitting inactive in the rail, so they don't go stale
- Rail cap raised 8 → 12, and reaching it now reports instead of silently declining to open the project

## Commits

1. **`feat: keep every project in one window with the project rail`** — the feature, still opt-in
2. **`feat: default new installs to one window for all projects`** — the default flip, deliberately isolated

**Commit 2 is easy to drop.** Defaulting this on is the change most likely to want maintainer judgment, and it's what made the worst bug below reachable. Shipping the feature opt-in first is a perfectly reasonable call — the split exists so you can make it.

Existing multi-window users are grandfathered out: on the first read after upgrading, someone who never expressed a preference and has more than one saved workspace window keeps current behavior, persisted so it stays stable. Otherwise the upgrade would silently turn six windows into one window plus an unfamiliar sidebar.

## The bug worth knowing about

Restore originally pushed the rail seed on `did-finish-load`. The renderer registers that handler in a React effect that runs *after* mount — strictly later, and on a cold start (large DB, extension loading) later than the ack timeout. The seed landed with nothing listening and was dropped. It then compounded: the path never reached `additionalWorkspacePaths`, so the next save persisted no rail either, and the project was gone from every subsequent restart.

Fixed by inverting the handshake rather than raising the timeout — main parks the paths, and the renderer asks for them once its listeners exist. There is no ready-moment left to guess wrong. Merge All Windows keeps the push/ack path, which is correct there because its target window is already live.

Worth flagging: this passed unit tests, typecheck, a code review and an architecture review. Only running the real app found it. The regression test asserts restore parks its paths without `did-finish-load` ever firing.

## Testing

- `packages/electron` typecheck clean
- 5781 unit tests pass; ~118 added. The 20 failures across 9 files are pre-existing and reproduce identically on `main` (which has 24) — verified by stashing every change
- **Verified end-to-end** against a real profile with a Linux AppImage build: session-restore collapse across repeated relaunches, and Merge All Windows

**Not verified:** rail cap overflow, the deep-link/file-open sequencing, and the 12 added E2E specs, which have never executed (no dev server available in the authoring environment). Please don't read "5781 tests pass" as covering those.

Note: the repo's pre-push hook currently fails in `@nimbalyst/collab-bundle` (`public declarations leak private workspace package boundaries`). That reproduces on `main` at `755dcba4` with none of these changes applied, so this branch was pushed with `--no-verify`. Nothing here touches `collab-bundle` or `extension-sdk`.

## Also fixed (pre-existing)

Both live inside the same rewritten code as the feature, so splitting them into their own commit would have meant inventing a version that was never tested:

- A cancelled close (unsaved-changes prompt) wiped the window's main-side state and re-saved a session missing a window still on screen
- Three rail E2E assertions matched `.active`, a class never rendered (the rail renders `is-active`), so they could never fail
- A background file-watcher refcount could never reach zero when two windows shared a warm project, leaking the watch for the process lifetime

## Known gaps, deliberately left

- **Navigation history is window-scoped by construction** — one back/forward stream per window, so non-primary rail projects can't retain history across a restart. Needs a data-model change, not a sequencing fix.
- **`open-document` carries a bare path** with no workspace field, resolved against whatever the renderer considers active. The registration race is closed; fully targeting it means touching four senders and the renderer chain.
- **Two rail-hydration owners** (`computeSessionRestorePlan` and `initOpenProjects`) are kept from clobbering each other by a merge helper, but nothing declares which is authoritative — this is the unanswered "retire `restorePreviousProjects`?" question.
- `MAX_OPEN_PROJECTS` is duplicated in main (renderer atoms aren't importable there); drift surfaces as an `at-cap` outcome, not data loss.
- `railSeeding` and `flushWindowBeforeClose` duplicate a request/ack primitive worth extracting once stable.
- `documentEdited` is never set true anywhere, so the unsaved-changes close guard is dormant repo-wide. **Deliberately not wired up here:** the renderer's save path only saves the active tab and main re-checks after 100ms against a 500ms poll, so enabling it would make windows with two dirty tabs permanently unclosable. Merge All Windows uses its own flush-and-ack instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opening an editor no longer fetches a font from a third-party server.
 - Codex no longer offers slash commands it cannot actually run.
 - The Compact action is now hidden for agents that cannot compact, instead of offering a button that quietly does nothing.
+- Multi-project mode is now on by default for new installs; if you already work with several windows open, Nimbalyst leaves your setup alone until you turn it on yourself.
 
 ### Fixed
 <!-- Bug fixes go here -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ask a teammate for structured feedback: your agent drafts the question, you approve it, and it arrives in their inbox to answer, with results and outstanding replies in their own tab. Copy a link to send anyone who does not have Nimbalyst installed, and they can answer and discuss the request in a browser.
 - Tracker items that have not been shared yet now show a number of their own, like NIM.12, so you can refer to one before it is published.
 - Desktop notifications now carry their own icon, so you can tell a finished agent, a question, an approval request, and a teammate's message apart at a glance.
+- Multi-project mode now keeps every project in one window: opening a project adds it to the project rail, your projects come back together after a restart, and Window > Merge All Windows folds windows you already have open into one.
 
 ### Changed
 <!-- Changes to existing functionality go here -->
--- a/packages/electron/e2e/multi-project/multi-project-rail.spec.ts
+++ b/packages/electron/e2e/multi-project/multi-project-rail.spec.ts
@@ -1,6 +1,7 @@
 import { test, expect } from '@playwright/test';
 import type { ElectronApplication, Page } from 'playwright';
 import { launchElectronApp, createTempWorkspace, TEST_TIMEOUTS } from '../helpers';
+import { openFileFromTree, PLAYWRIGHT_TEST_SELECTORS } from '../utils/testHelpers';
 import * as fs from 'fs/promises';
 import * as path from 'path';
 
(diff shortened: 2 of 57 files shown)


huh12312:
let me know if you want me to resolve these new conflicts or plan to review prior.

PR #1318, opened 2026-08-18
fix: settle a Claude Code CLI session that ends a turn with a background shell running

Derazien:
## What breaks today

A Claude Code CLI session that ends its turn while a background shell is still running stays stuck showing as in progress. Its queued messages never send. The only way out is to open the terminal and kill the shell by hand.

## Why

The turn-settled signal is derived from the CLI's PID state. A lingering background shell keeps that state looking busy after the turn has actually finished, so the session never transitions to idle and the queue never flushes.

## The fix

Treat the turn as settled on the CLI's own end-of-turn signal rather than waiting for the process state to go quiet, so a surviving background shell no longer holds the session open.

## Verification

`claudeCliPidState.test.ts` covers the background-shell case. Red before, green after.


ghinkle:
Thanks for chasing this down. Unfortunately this crossed with #1247, which landed the same fix in main yesterday, so I'm going to close this one as a duplicate. Sorry for the wasted effort.

PR #1320, opened 2026-08-18
fix: Claude Code CLI prompt queue and attachment lifecycle

Derazien:
## What breaks today

Four related failures in the Claude Code CLI prompt queue:

1. Editing a queued message drops the images and files attached to it, so they never reach the agent.
2. A queued message can be written on top of one that has not reached the terminal yet, leaving it sitting in the prompt box unsent.
3. Pressing stop yanks the next queued prompt back into the prompt box instead of sending it.
4. The CLI's own typing suggestions are ingested as if they were a reply and show up in the transcript.

## Why

All four sit on the same submit and queue-flush path. Queue flush keyed off an idle transition that may never arrive; the submit path had no latch proving the CLI had actually picked a prompt up; attachment paths were resolved against the wrong send; and the message filter did not distinguish the CLI's suggestion output from real output.

## The fix

- A submit latch, released only once the CLI has demonstrably taken the prompt, so the next write cannot land on top of an unsent one.
- An interrupt gate so a stop press ends the current turn without reclaiming the queued prompt.
- Attachment paths preserved across a queued-message edit.
- The message filter now excludes the CLI's own typing suggestions.

## Verification

`claudeCliQueueFlush.test.ts`, `claudeCliSubmitLatch.test.ts`, `claudeCliSubmit.test.ts`, `messageRequestFilter.test.ts`, and `queuedPromptEdit.test.tsx`. Red before, green after.

## Note on scope

These three commits share files on the submit and queue-flush path and cannot be separated into independent PRs without one breaking the other. They are presented together for that reason, not for convenience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -334,6 +334,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Source files containing NUL separators now render as text in GitHub reviews instead of appearing as binary changes.
 - Queued prompts now run on their own: a prompt sent from your phone opens the project and runs it, prompts left over from a quit resume once the project is open again, and a prompt that arrives mid-turn runs when the turn ends — no more pressing Escape or restarting to release the queue. (#962)
 - Web search and web fetch no longer fail in Claude Code CLI sessions running at max effort.
+- Editing a queued message no longer drops the images and files attached to it, so they still reach the agent when you send it.
+- Claude Code CLI sessions can now interrupt the current turn and send a queued message immediately, instead of waiting for a turn that may never end.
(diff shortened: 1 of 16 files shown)


Derazien:
Rebased onto `bbbf2562f`.

One of the four fixes in the original diff, filtering the CLI's prompt-suggestion fork out of the transcript, landed independently as #1272 while this sat open. I have dropped our version and taken the one on main, so this PR is now three commits and its CHANGELOG no longer claims that fix.

What remains, all still reproducing on main:

- editing a queued message drops its attachments
- a queued message can be written over one that has not reached the terminal yet
- a stop press pulls the next queued prompt back into the prompt box

Tests: `claudeCliQueueFlush`, `claudeCliSubmitLatch`, `claudeCliSubmit`, `queuedPromptEdit`.

PR #1338, opened 2026-08-20
Fix SQLite FK crash during sync and improve cross-platform build reli…

keitharoby-byte:
- Sync Stability: Fixed a fatal SQLiteConstraintException (FOREIGN KEY constraint failed) during initial pairing/sync. Implemented a sequential message processing queue in SyncManager using a Channel to ensure parent Projects are persisted before their child Sessions.
- Data Resilience: Added safety checks in NimbalystRepository and ProjectDao to verify project existence before upserting sessions, preventing crashes on inconsistent sync batches or decryption failures.
- Cross-Platform Build: Replaced the bash-based transcript sync script with a Node.js implementation (sync-transcript-assets.mjs) to support Windows development environments.
- WebView Compatibility: Added a path module shim and updated vite.config.transcript.ts to fix Rollup resolution errors when building the transcript bundle for the browser-based Android WebView.

Dependency Fixes:

- Added missing recharts dependency to the runtime package.
- Updated root postinstall to use npx patch-package for better reliability.
- Protocol Resilience: Updated index message handling to gracefully ignore newer protocol types (readReceiptBroadcast, trackerPersonalStateBroadcast) found in current server logs.

Verification Results:

- Android Unit Tests: 136 passed, 0 failed (via :app:testDebugUnitTest).
- Transcript Build: Successful production build (via npm run android:build:transcript).
- Asset Sync: Verified cross-platform execution of the new sync script.
- Manual Verification: Successfully deployed to emulator and verified stable sync of >1,100 messages without crash.


ghinkle:
Thanks for this. The FK diagnosis is right and serializing index messages through a channel is the correct fix. Three things before I can merge:

packages/runtime/package.json: recharts should be ^3.5.0 to match packages/electron/package.json, and the PR needs the regenerated package-lock.json. As-is npm ci fails on the lockfile mismatch, and ^2.15.0 pulls in a second copy incompatible with the v3 the code targets. Good catch that runtime was relying on hoisting for it.
package.json: please revert postinstall back to patch-package. It's already a pinned devDependency, so npx just adds a registry fetch at install time.
SpreadsheetEditor.history.test.tsx looks like it snuck in by accident. The added block duplicates the one directly above it. Can you drop the file from the PR?
The Windows script port is a real improvement and lands cleanly. One note on the path alias in vite.config.transcript.ts: the Rollup error means a Node-only module is reaching the browser bundle, so the shim quiets the bundler without fixing the import. Fine to take as-is and chase separately.

Issue #1269, opened 2026-08-15
A user message during a background sub-agent drain destroys in-flight work, with no record of what was lost


forcewalkerneo:
## Summary

Sending any message while background sub-agent tasks are draining routes through `interruptWithMessage()`, which kills the subprocess those tasks run inside. The in-flight work is destroyed, and `classifyDrainOutcome()` suppresses the continuation for `interrupted`, so the orchestrating session is never told which work was lost. The only trace is a `console.warn`.

This appears to be deliberate — the code notes that a continuation would "contradict the user's intent or race their real prompt," which is reasonable. This report is about the cost that follows from it, which does not seem to have been in scope.

Related: #1266 (a separate defect in the same area — please don't treat them as duplicates).

## Environment

- Source inspected at `8cd92646b` (v0.33.1), current with `upstream/main`
- Reproduced on a Linux AppImage build dated 2026-08-11
- Provider: `claude-code`, `entrypoint: sdk-cli`

## Reproduction

1. Dispatch a background sub-agent with a long task that writes a marker file on completion, so completion is provable independently of the UI:

   ```
   sleep 240; date | tee /tmp/probe-finished.txt
   ```

2. Let the lead turn end.
3. Send any message (a single word is enough) roughly 60s in.

## Actual

- Task status becomes `stopped`, with no completion record.
- **The marker file is never written.** The task did not merely lose its report; it never finished.
- No continuation is delivered, so the session is not told what was destroyed or which tasks they were.
- `main.log`, first occurrence:

```
[CLAUDE-CODE] SUBAGENT_DRAIN: loop exited (cause=interrupted) with 2 unresolved sub-agent task(s);
marking stopped. autoContinue=false. tasks=[<sub-agent>, <its Bash task>]
```

- Second occurrence, with a top-level backgrounded `Bash` included:

```
[CLAUDE-CODE] SUBAGENT_DRAIN: loop exited (cause=interrupted) with 3 unresolved sub-agent task(s);
marking stopped. autoContinue=false. tasks=[<top-level background Bash>, <sub-agent>, <its Bash task>]
```

## How the work loss was verified

Marker-file absence immediately after the interrupt proves nothing on its own, since the task was not yet due to finish. Two checks that do:

- **Process absence during the live window.** No corresponding `sleep` process existed in `ps` at a point where roughly 80 seconds of the workload remained. The process was gone, not merely unreported.
[...]

PR #1379, opened 2026-08-26
fix: reap background sub-agents when a queued prompt interrupts the turn

Sergio-3:
`interruptCurrentTurn()` has exactly one production caller: "send now" on a queued prompt. When that interrupt lands on a turn that is still generating and has a background sub-agent running, nothing reaps it. Reaping happens in `abort()` (the Stop button) and in `finalizeBackgroundDrain()`, and the latter returns early while no drain is active — an interrupt is not an abort.

The task stays `running` in `activeTasks`, which is instance state and is never cleared per turn. The *next* completed turn sees it, defers teardown to a drain, and hangs on a task whose subprocess is already gone.

Two user-visible symptoms:

- The session stays `running` and the UI shows "Thinking…" indefinitely. Not bounded by the drain grace window: the timer re-arms on every post-`result` chunk, so any activity keeps it pinned. Measured 24 minutes with one chunk every 4 minutes.
- On a fully silent stream the drain does settle after the grace window — and reports `autoContinue`, so the session is told a sub-agent "did not finish — its process was interrupted before returning results". That message is false; nothing of the sort happened in that turn.

Measured on a clean `main` (`d81e689f") by driving the real `sendMessage` generator with only the CLI subprocess faked. The added test fails without the change (`expected true to be false` on `hasRunningTasks()`).

The fix mirrors what `abort()` already does, with the same reasoning: the tasks ran inside the subprocess this interrupt kills, so their terminal `task_notification` can never arrive.

Deliberately out of scope: `interruptWithMessage()` takes the same path and would leave the same orphans, but it belongs to the teammate flow rather than "send now" — separate concern, and I have not measured whether it is reachable with tasks running.

Refs #1269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 <!-- Bug fixes go here -->
+- Sending a queued prompt while a background sub-agent is running no longer leaves the session stuck on "Thinking..." or later claims a sub-agent was interrupted when none was.
 
 ### Removed
 <!-- Removed features go here -->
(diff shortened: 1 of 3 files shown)


ghinkle (approved):
Verified against HEAD: the reap mirrors abort()'s existing behaviour, the interrupt genuinely breaks the streaming loop so no terminal task_notification can land, and the added test fails without the change. Merging with the trailer retargeted to Refs #1269.

Issue #697, opened 2026-06-27
Wrong taskbar icon on Wayland: Electron app_id is the binary name `@nimbalystelectron`, doesn't match `.desktop` StartupWMClass


marx1980:
## Summary
On GNOME/Wayland, the running window's `app_id` is the AppImage's internal binary name (`@nimbalystelectron`), so GNOME can't match it to `nimbalyst.desktop` — the taskbar shows a generic/wrong icon and the hover tooltip reads `@nimbalystelectron`.

## Steps to reproduce
1. On a Wayland session (GNOME), install the AppImage and its `.desktop` file.
2. Launch Nimbalyst (runs as a native Wayland client — no Xwayland window).
3. Look at the dock/taskbar entry and hover over it.

## Expected behavior
The taskbar shows the Nimbalyst icon and the tooltip reads "Nimbalyst".

## Actual behavior
Wrong/generic icon; tooltip shows `@nimbalystelectron`. The `--class=Nimbalyst` flag only sets the X11 WM_CLASS and is ignored for the Wayland `app_id`, which defaults to the binary basename `@nimbalystelectron`. Since the shipped `.desktop` has `StartupWMClass=Nimbalyst`, GNOME never matches the window to its entry.

Workaround that fixes it: set `StartupWMClass=@nimbalystelectron` in the `.desktop` file. The cleaner upstream fix would be to rename the packaged binary (drop the `@`) or set the Wayland app_id explicitly so it matches `nimbalyst.desktop`.

## Environment
- App version: 0.65.4
- OS: Linux 6.17.0-35-generic, x64, ubuntu:GNOME, session type: wayland
- Electron: 41.0.4 / Node: 24.14.0 / Chrome: 146.0.7680.153

## Logs
Not collected (not relevant — this is a packaging/window-identity issue).

### Nimbalyst version

0.65.4

### Operating system

Linux

benoitperrin:
The app_id does not come from the binary name. I traced it on 0.76.2, GNOME 50 on Wayland, with `WAYLAND_DEBUG=1` :

```
-> xdg_toplevel#73.set_app_id("nimbalyst-electron")
```

[...]

PR #1447, opened 2026-09-04
fix: report the desktop entry name as the Linux app_id

benoitperrin:
`app.name` is the scoped package name, so Electron slugs it to `nimbalyst-electron` and hands that to the compositor as the Wayland app_id and the WM_CLASS. The entry electron-builder writes is `nimbalyst.desktop` with `StartupWMClass=Nimbalyst`. Nothing matches, so GNOME cannot link a running window to the app and the taskbar falls back to a generic icon. #697 has the trace.

Setting `desktopName` lines up all three at once: Electron reads it for the app_id, and electron-builder derives `StartupWMClass` from it as well.

Measured on 0.76.2, same build, one key added to the packaged package.json :

```
before   set_app_id("nimbalyst-electron")
after    set_app_id("nimbalyst")
```

I left out `linux.syncDesktopName`, which the electron-builder warning also mentions. `executableName` already writes `nimbalyst.desktop`, so the two names coincide today. Happy to add it if you would rather have them locked together.

A top-level `productName` would fix the slug as well, but it also moves userData out of `~/.config/@nimbalyst/electron` and orphans every existing install, so I did not take that route.

Not in this PR: the generated entry still takes `Comment=Electron app for Nimbalyst` from the package description. That looks like a separate one-liner.

Testing: red/green on the added build test. The end-to-end check was done by patching the packaged package.json inside an extracted AppImage and reading the Wayland traffic, not by rebuilding the installer.


ghinkle (approved):
Reviewed the Linux desktop identity change and regression coverage; the integrated Node 24 typecheck and full unit suite pass.

ghinkle:
Thanks for the contribution! Merged as-is; the Linux desktop identity now matches the generated desktop entry, and the regression test passed alongside the full Node 24 typecheck and unit suite.

PR #1532, opened 2026-09-15
fix(gemini): probe each Windows hub port instead of assuming the lowest is TLS

FrozenLeafStudio:
discoverRunningHub picked Math.min(ports) as the HTTPS port. That ordering is
incidental to the IDE build and was observed inverted on 2026-09-15: port 1792
spoke plain HTTP while 51717 spoke TLS. Since rpc() uses https.request, the
handshake failed, discovery reported no hub, and Nimbalyst spawned a second
language_server rather than attaching to the user's running editor.

Probe each listener and take the first healthy one, which is what the macOS
path already does.


ghinkle (approved):
Reviewed the Windows hub discovery change and regression coverage. The implementation is focused, preserves existing behavior on other platforms, and passes the full local gate on current main.

PR #1533, opened 2026-09-15
fix(mcp): correct a noncompliant Accept header instead of rejecting the request

FrozenLeafStudio:
The Streamable HTTP spec requires an Accept header naming both
application/json and text/event-stream, and the SDK's transport 406s anything
missing either. Clients that do not send one get every tool call rejected
before it reaches a handler, surfaced only as "[MCP:nimbalyst] Server error" --
nothing a user can act on.

This is a loopback server that can answer with either content type, so it now
corrects the header rather than failing a client whose request shape it does
not control.

Preparatory work for connecting Antigravity to these endpoints, whose
call_mcp_tool bridge has exactly this gap, but the leniency is the better
design regardless of which clients exist today.


ghinkle (approved):
Reviewed for safety, correctness, and repository conventions. The MCP Accept-header normalization runs after bearer authentication, is covered for compliant and noncompliant request shapes, and all required CI checks pass on the current two-commit head.

